PR #5, opened 2018-08-21
Support both CI and product builds

ironcladlou:


ironcladlou:
Depends on https://github.com/openshift/release/pull/1238 becoming effective for the image build to pass

PR #9, opened 2018-09-07
Makefile should generate bindata before the build

pravisankar:


pravisankar:
@openshift/sig-network-edge  PTAL

PR #15, opened 2018-09-25
Do not use 'operator' name for dns cluster role and role bindings

pravisankar:
Otherwise it will create confusion with roles used in cluster-dns-operator.


pravisankar:

ironcladlou:
Same as https://github.com/openshift/cluster-ingress-operator/pull/34#issuecomment-424417582; are we sure it's better to have unprefixed names for cluster scoped resources? I kind of liked the idea of "namespacing" our cluster roles via prefixes so it's very obvious what they're for.

pravisankar:
@openshift/sig-network-edge updated, PTAL

PR #17, opened 2018-09-26
And manifest asset content export for use with the installer.

ramr:
Need this for the installer to be able to directly make a call into this operator and get assets required to start it up.
@pravisankar  PTAL Thx


ironcladlou:
Background on this? I don't get it.

ramr:
Update: had a conversation w/ Dan and Rajat this morning re: the need for this PR and resulted in #18 
@ironcladlou  @pravisankar  rebased [and updated] this PR for the new methods. PTAL Thx

Aside: For some reason my local build changed bindata (2 bytes off in the size). Can drop that commit if not's the case for you folks. Thx

pkg/manifests/manifests.go
```diff
@@ -236,3 +236,14 @@ func NewCustomResourceDefinition(manifest io.Reader) (*apiextensionsv1beta1.Cust
...
+
+func (f *Factory) OperatorAssetContent() (map[string][]byte, error) {
+	return map[string][]byte {
+		OperatorCustomResourceDefinition: MustAsset(OperatorCustomResourceDefinition),
+		OperatorNamespace:                MustAsset(OperatorNamespace),
+		OperatorServiceAccount:           MustAsset(OperatorServiceAccount),
+		OperatorClusterRole:              MustAsset(OperatorClusterRole),
+		OperatorClusterRoleBinding:       MustAsset(OperatorClusterRoleBinding),
+		OperatorDeployment:               MustAsset(OperatorDeployment),
+	}, nil
```
ironcladlou: `MustAsset` panics on error, so no reason to return `error` from the func

test/integration/integration_test.go
```diff
@@ -162,6 +163,18 @@ func testDNSRecords(t *testing.T) {
 	}
 }
 
+func testOperatorAssetContent(t *testing.T) {
```
ironcladlou: Please move this to the unit test

ironcladlou:
The 2 bytes are my fault; I added a line break in my last commit without regenerating biodata. Please commit an updated biodata with this PR. We need a verification script to catch that sort of thing

ironcladlou:
Awesome, thanks.

/lgtm

PR #18, opened 2018-09-26
Expose operator manifests through factory

ironcladlou:
Incorporate the operator manifests into code generation and expose them through
the factory to support direct integration with the new OpenShift installer.

Supports https://github.com/openshift/installer/pull/331.


pkg/manifests/manifests.go
```diff
@@ -21,6 +22,13 @@ const (
 	DNSConfigMap          = "assets/dns/configmap.yaml"
 	DNSDaemonSet          = "assets/dns/daemonset.yaml"
 	DNSService            = "assets/dns/service.yaml"
+
+	OperatorCustomResourrceDefinition = "manifests/00-custom-resource-definition.yaml"
```
ramr: Typo in name - it works but ... [OperatorCustomResour]`r`[ceDefinition]

pravisankar: *Resource (fix typo)

pkg/manifests/manifests.go
```diff
@@ -37,6 +45,54 @@ func NewFactory() *Factory {
 	return &Factory{}
 }
 
+func (f *Factory) OperatorCustomResourceDefinition() (*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(OperatorCustomResourrceDefinition))
```
pravisankar: Same here: CustomResource

ironcladlou: Fixed

pkg/manifests/manifests.go
ironcladlou: Fixed

ironcladlou: Fixed

PR #22, opened 2018-09-27
Ensure operator can work in absence of cluster DNS

ironcladlou:
Eliminate the dependency on cluster DNS for the operator itself by using host
networking and talking to the apiserver over loopback. Ensure the operator is
only scheduled to masters.

### DIFF
--- a/manifests/02-deployment.yaml
+++ b/manifests/02-deployment.yaml
@@ -13,17 +13,31 @@ spec:
       labels:
         name: cluster-dns-operator
     spec:
-      containers:
-        - name: cluster-dns-operator
-          image: openshift/origin-cluster-dns-operator:latest
-          command:
-          - cluster-dns-operator
-          imagePullPolicy: IfNotPresent
-          env:
-            - name: WATCH_NAMESPACE
-              valueFrom:
-                fieldRef:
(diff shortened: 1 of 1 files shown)


imcsk8:
LGTM

ironcladlou:
As @rajatchopra pointed out, we'll need to address the apiserver from CoreDNS itself using the apiserver service static clusterIP; since SDN doesn't depend on DNS, could *both* the operator and CoreDNS should be able to rely on the service network, and both address the apiserver via clusterIP without host networking?

/cc @deads2k @ramr @pravisankar @knobunc @squeed

rajatchopra:
> As @rajatchopra pointed out, we'll need to address the apiserver from CoreDNS itself using the apiserver service static clusterIP; since SDN doesn't depend on DNS, could _both_ the operator and CoreDNS rely on the service network, and both address the apiserver via clusterIP without host networking?

Using the apiserver static IP is one way, the other is to ask for the apiserver dns itself from install config. I think we should do the latter. Put that as an input to the config?

ironcladlou:
Still discussing how we could inject the apiserver cluster IP or external dns name into the pods which is going to be necessary to propagate down into the CoreDNS DaemonSets anyway.

PR #23, opened 2018-09-27
Make dns daemonset, configmap and service creation specific to given custom resource

pravisankar:
- This will help us in handling deletion events. We will know exact resources to delete.
- This also helps to support multiple dns deployments with different configs if needed.


pravisankar:

pkg/manifests/manifests.go
```diff
@@ -130,18 +131,36 @@ func (f *Factory) DNSConfigMap(dns *dnsv1alpha1.ClusterDNS) (*corev1.ConfigMap,
...
+
+	if ds.Spec.Selector.MatchLabels == nil {
+		ds.Spec.Selector.MatchLabels = map[string]string{}
+	}
+	ds.Spec.Selector.MatchLabels["dns"] = ds.Name
+
+	if len(ds.Spec.Template.Spec.Volumes) == 0 {
+		return nil, fmt.Errorf("expected volume for core file not found")
+	}
+	ds.Spec.Template.Spec.Volumes[0].ConfigMap.Name = ds.Name
```
ironcladlou: Look it up by name (config-volume) instead of index?

Miciah:
Might as well update `assets/dns/service.yaml` to add a `# Name is set at runtime` comment as you have done for `assets/dns/configmap.yaml` and `assets/dns/daemonset.yaml`.

pravisankar:
@ironcladlou @Miciah  updated, PTAL

pkg/manifests/manifests.go
```diff
@@ -130,18 +131,45 @@ func (f *Factory) DNSConfigMap(dns *dnsv1alpha1.ClusterDNS) (*corev1.ConfigMap,
...
+	ds.Spec.Template.Labels["dns"] = ds.Name
+
+	if ds.Spec.Selector.MatchLabels == nil {
+		ds.Spec.Selector.MatchLabels = map[string]string{}
+	}
+	ds.Spec.Selector.MatchLabels["dns"] = ds.Name
+
+	if len(ds.Spec.Template.Spec.Volumes) == 0 {
+		return nil, fmt.Errorf("no volumes found, expected volume for core file")
+	}
```
Miciah: This check is superfluous now.

ironcladlou:
Thanks!

/lgtm

PR #24, opened 2018-09-27
 Handle delete events in cluster dns operator

pravisankar:
Depends on https://github.com/openshift/cluster-dns-operator/pull/23


pravisankar:

pkg/stub/handler.go
```diff
@@ -23,16 +24,42 @@ type Handler struct {
...
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	if event.Deleted {
-		return nil
-	}
 	switch o := event.Object.(type) {
 	case *dnsv1alpha1.ClusterDNS:
-		return h.syncDNSUpdate(o)
+		if event.Deleted {
+			return h.deleteDNS(o)
```
ironcladlou: Something else to note about the event-driven delete pattern is that if any step fails, I'm not sure we'll actually get a new delete event dequeued. Just another random aside regarding level driven deletes.

pkg/stub/handler.go
```diff
@@ -23,16 +24,42 @@ type Handler struct {
...
+		} else {
+			return h.syncDNSUpdate(o)
+		}
 	}
 	return nil
 }
 
+func (h *Handler) deleteDNS(dns *dnsv1alpha1.ClusterDNS) error {
```
ironcladlou: Maybe add a TODO to use the GC

pravisankar: Unless router, dns has got configmap and configmap need to created before daemonset. Do we want to make dns configmap as owner ref parent for daemonset and service  or daemonset as owner ref for configmap and service?
Other thing to consider is owner ref seems suitable for driving the dependent resources. In this case, even though some of these resources are not tightly coupled, we want to add owner ref to leverage GC during deletion. Is that the intended purpose of owner ref or we misusing it?

ironcladlou: Hm. Let's think more about it later. Since we're using deterministic naming it's not hard to clean up after ourselves anyway in the meantime.

ironcladlou: Or if naming isn't good enough, assign labels to things so we can reference them that way. One potential pitfall with naming or label selector approach is the possibility of erroneous adoption of same-named things (hence UIDs for owner refs).

Anyway, food for thought for post-alpha.

PR #25, opened 2018-09-28
Use cluster dns ip address from install config if "available"

ramr:
Configure the operator to use the same DNS IP used by the kubelet

jira ticket: https://jira.coreos.com/browse/NE-86

PTAL  Thx

*Edited* Removed old test info

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -6,8 +6,12 @@ import (
 
 	dnsv1alpha1 "github.com/openshift/cluster-dns-operator/pkg/apis/dns/v1alpha1"
 	"github.com/openshift/cluster-dns-operator/pkg/manifests"
+	"github.com/openshift/cluster-dns-operator/pkg/util"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sclient"
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
+
+	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -61,6 +65,16 @@ func (h *Handler) deleteDNS(dns *dnsv1alpha1.ClusterDNS) error {
 }
 
 func (h *Handler) syncDNSUpdate(dns *dnsv1alpha1.ClusterDNS) error {
+	if dns.Spec.ClusterIP == nil {
+		// Check for default cluster ip.
+		ipaddr, err := util.ClusterDNSIP(k8sclient.GetKubeClient())
+		if err != nil {
+			logrus.Errorf("Getting cluster dns ip: %v", err)
+		} else {
+			logrus.Infof("Using default cluster dns ip address %s", ipaddr)
+			dns.Spec.ClusterIP = &ipaddr
+		}
+	}
 	ns, err := h.manifestFactory.DNSNamespace()
 	if err != nil {
 		return fmt.Errorf("couldn't build dns namespace: %v", err)
--- a/pkg/util/config.go
+++ b/pkg/util/config.go
@@ -1,0 +1,82 @@
+package util
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/apparentlymart/go-cidr/cidr"
+	"github.com/ghodss/yaml"
+
+	kapicore "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// INSTALLER_CONFIG_NAMESPACE is the namespace containing the installer config.
+	INSTALLER_CONFIG_NAMESPACE = "kube-system"
+
+	// CLUSTER_CONFIG_RESOURCE is the resource containing the installer config.
+	CLUSTER_CONFIG_RESOURCE = "cluster-config-v1"
+)
+
+// installerClusterDNSIP
+func installerClusterDNSIP(cm *kapicore.ConfigMap) (string, error) {
+	if cm == nil {
+		return "", fmt.Errorf("invalid installer config")
+	}
+
+	ic, ok := cm.Data["install-config"]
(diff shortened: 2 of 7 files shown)


ramr:
@ironcladlou  am off most of the day tomorrow - but I can take a look at comments in the afternoon/eve.
 There is one pending change  remaining here for this PR to be fully tested - which is to allow the cluster-dns-operator service account access to the kube-system namespace (get configmaps). Wasn't sure if we should do that by default?  Thx

@rajatchopra  so I saw from the sample file @ironcladlou  sent me that this is stored in a configmap and not in a secret as discussed earlier. Is that is correct?  Thx

ramr:
Also we do want to turn off the logs here. I see this every 5 secs - really hot reconciler loop looks like: 
```
time="2018-09-28T05:14:12Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:17Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:22Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:27Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:32Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:37Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:42Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:47Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:52Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
time="2018-09-28T05:14:57Z" level=info msg="Using default cluster dns ip address 172.30.1.10"
```

ironcladlou:
This is an absolutely bonkers amount of dependencies to drag in for an API type- I'd rather just consume the raw YAML of the installer if they're not going to be publishing this type to the API repos at this point. Since you're out today I might pick up your branch and play around with it.

pravisankar:
@ramr 5 sec resync loop is very aggressive, I will update to some sane value, may be 10 mins?

ramr:
@ironcladlou  yeah, I think I glazed over the number of lines changed last night. Didn't realize it was 51k jeez! - thought it was like 5k which is normal.  I'll fix this tonight/over the weekend if you haven't worked on it. 

@pravisankar yeah 5 secs might be too aggressive - increasing the value aka decreasing how often we do a resync is good. We could probably do a wee bit more to check if the `ClusterDNS` object has really changed. Something like keep a map of `ClusterDNS` objects and on a create/modify do like a deep equal check on the spec (and possibily metadata) to see if anything's changed. Thx

ramr:
@ironcladlou  PTAL  thx

ironcladlou:
@ramr while you're at it, want to go ahead and default `ClusterDomain` to `cluster.local`?

pkg/util/config.go
```diff
@@ -0,0 +1,82 @@
+package util
```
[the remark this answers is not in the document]
ramr: d'oh i normally do and didn't run it after the few rebases I just did. Fixed.  
i'll create a makefile change to do a lint check + target for a local build.

ramr:
@ironcladlou  the corefile (the configmap really) has `cluster.local` by default, so we don't replace it if `ClusterDomain` is not set.  So that's the default.
Do you still want to explicitly set `ClusterDomain` to `cluster.local`?

ironcladlou:
@ramr 
>@ironcladlou the corefile (the configmap really) has cluster.local by default, so we don't replace it if ClusterDomain is not set. So that's the default.
Do you still want to explicitly set ClusterDomain to cluster.local?

Ah yeah, guess it doesn't matter

pkg/stub/handler.go
```diff
@@ -61,6 +65,14 @@ func (h *Handler) deleteDNS(dns *dnsv1alpha1.ClusterDNS) error {
 }
 
 func (h *Handler) syncDNSUpdate(dns *dnsv1alpha1.ClusterDNS) error {
+	if dns.Spec.ClusterIP == nil {
+		// Check for default cluster ip.
+		ipaddr, err := util.ClusterDNSIP(k8sclient.GetKubeClient())
+		if err == nil {
+			logrus.Infof("Using default cluster dns ip address %s", ipaddr)
+			dns.Spec.ClusterIP = &ipaddr
+		}
```
ironcladlou: Forgot to log when `err != nil`

ironcladlou: Should probably just throw the error out and let the handler re-queue it

ironcladlou:
Please make sure the integration test passes with this, provide output in a comment here so we can verify the defaulting worked

pkg/stub/handler.go
ramr: We were generating too many messages before the resync period changed but I can add that in now.

ironcladlou:
@pravisankar disabled the test cache for integration tests at one point and I think the change got lost somehow- we should definitely disable the Go test cache for the `test-integration` target.

Thanks for this work!

/lgtm

pkg/stub/handler.go
pravisankar: Oops, I should have reviewed this PR earlier. 
This will stamp ClusterIP from CVO (if not set) for all dns custom resources. This seems like a hack and we are not leveraging any of the CRD/CR stuff for cluster dns and will make it hard to support multiple ClusterDNS configurations in the future.
Why not do this instead:
- CVO has the needed info, populate the ClusterDNS custom resource with the necessary info from CVO and then create the custom resource (probably we need to do this in CVO repo as it owns the responsibility of running the operator config). Code [1] and [2] ensures everything works as expected. 
[1]  https://github.com/openshift/cluster-dns-operator/blob/master/pkg/manifests/manifests.go#L136
[2]  https://github.com/openshift/cluster-dns-operator/blob/master/pkg/manifests/manifests.go#L81

@ironcladlou @ramr ^^

PR #26, opened 2018-09-28
Updated operator-sdk to v0.0.6

pravisankar:
Changed informer resync period to 10 mins


pravisankar:

ironcladlou:
I think you need to update the integration test code as well with the `Watch` signature change

PR #28, opened 2018-10-02
Fix RBAC roles for cluster DNS

pravisankar:
As Miciah suggested in https://github.com/openshift/cluster-ingress-operator/pull/35#issuecomment-424836114
This PR will allow operator to do anything in openshift-cluster-dns namespace and assigns only required cluster roles in openshift-cluster-dns-operator namespace.


pravisankar:

pravisankar:
@Miciah yes, we need to expose role and role manifests for CVO. I'm adding these changes
/hold

ramr:
@Miciah  we aren't really using the `OperatorAssetContent` currently, so ...  plus it is not certain if we will need it eventually or not.

ironcladlou:
OperatorAssetContent is now defunct, CVO landed and we can abandon our efforts to integrate directly with the installer

PR #29, opened 2018-10-02
 Changed delete handling to use garbage collection

pravisankar:
- ClusterDNS specific resources are tied to its daemonset as owner ref
so that deletion of this daemonset will trigger deletion of corresponding
configmap and service resources.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -14,7 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pravisankar:

Miciah:
The `sdk.Create` has the following doctext:

    // Create creates the provided object on the server and updates the arg
    // "object" with the result from the server(UID, resourceVersion, etc).

So `getDaemonSetOwnerRef` should be able to use `ds.UID` directly, with no need for the `d` variable or `sdk.Get` call.

pravisankar:
@openshift/sig-network-edge @Miciah  updated, PTAL

ironcladlou:
I don't think deletes will work when managed by CVO (we don't grant `delete` for DaemonSet yet), but I think it's okay for now since we only care about create at the moment for getting our stuff integrated. I mention this as another example to consider for our e2e tests.

/lgtm

PR #32, opened 2018-10-03
Makefile changes to add targets for building local image and gofmt verification

ramr:
Add makefile targets to build local images and verifying go code with gofmt.
Plus fix the gofmt issue that this found. 
@pravisankar  PTAL Thx

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ BIN=$(lastword $(subst /, ,$(MAIN_PACKAGE)))
 BINDATA=pkg/manifests/bindata.go
 TEST_BINDATA=test/manifests/bindata.go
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,14 +30,37 @@ generate: $(GOBINDATA_BIN)
...
 	go get -u github.com/jteeuwen/go-bindata/...
 
-test:
+test:	verify
 	go test ./...
 
 test-integration:
 	go test -v -tags integration ./test/integration
 
+verify:	verify-gofmt
```
ironcladlou: is this formatting error a subtle joke?

ramr: hmm, something weird there - it shows up as tab in vi.  Something expanding it on github?

ironcladlou: Are you suggesting a tab there was *intentional*? I suspect we'd have a lively argument over a `make fmt` command... 😁

ironcladlou:
Thanks!

/lgtm

Makefile
ramr: lol

PR #33, opened 2018-10-03
Move common dns custom resource operations to manifests

pravisankar:
- Perform CR agnostic operations beforehand and only perform
CR specific operations during watch handling.
- Only create assets for those resources that need to be patched before creation.
- Static resources are moved to manifests dir and are created with 'oc create or apply'
  If we need to fetch an object in memory, we will use sdk.Get() on the resource.
  This also reduces the boilerplate code that we generate for the assets.


pravisankar:

ironcladlou:
/hold

Let's go over this later tomorrow or Friday- would also like to get some e2e tests going in the PRs before we touch manifests or RBAC again

ironcladlou:
I think this one's outdated. 

/close

PR #34, opened 2018-10-05
update resources to avoid openshift cycles

deads2k:
Also add tolerations to ensure the daemonset always schedules on nodes.

/assign @ironcladlou


ironcladlou:
/lgtm

Thanks

PR #35, opened 2018-10-09
Use multistage Dockerfile

ironcladlou:
Convert the Dockerfile to multistage for pipeline build style conformance.

### DIFF
--- a/images/cluster-dns-operator/Dockerfile
+++ b/images/cluster-dns-operator/Dockerfile
@@ -1,25 +1,17 @@
+FROM openshift/origin-release:golang-1.10 as builder
+COPY . /go/src/github.com/openshift/cluster-dns-operator/
+RUN cd /go/src/github.com/openshift/cluster-dns-operator && make build
+
 FROM centos:7
-
-RUN yum install -y golang make
-
-ENV GOPATH /go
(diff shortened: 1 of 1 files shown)


images/cluster-dns-operator/Dockerfile
```diff
@@ -1,25 +1,17 @@
...
+COPY . /go/src/github.com/openshift/cluster-dns-operator/
+RUN cd /go/src/github.com/openshift/cluster-dns-operator && make build
 
+FROM centos:7
+LABEL io.openshift.release.operator true
 LABEL io.k8s.display-name="OpenShift cluster-dns-operator" \
       io.k8s.description="This is a component of OpenShift Container Platform and manages the lifecycle of cluster DNS services." \
       maintainer="Dan Mace <dmace@redhat.com>"
 
+COPY --from=0 /go/src/github.com/openshift/cluster-dns-operator/cluster-dns-operator /usr/bin/
```
knobunc: I would feel happier if the stage were named:
  https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages

ironcladlou: Fixed and confirmed with Luke that named stages are supported, good tip!

PR #40, opened 2018-10-23
Improve source verification tooling

ironcladlou:
* Verify generated go-bindata output
* Vendor go-bindata
* Apply gofmt more universally
* Clean up Makefile


Gopkg.toml
```diff
@@ -9,8 +9,34 @@ required = [
...
   "k8s.io/gengo/args",
+  "github.com/kevinburke/go-bindata",
 ]
 
+[prune]
+  go-tests = true
+  non-go = true
+  unused-packages = true
+
+[[prune.project]]
```
pravisankar: Don't this need to be indent under `[prune]` section? (https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#prune)

Gopkg.toml
```diff
@@ -9,8 +9,34 @@ required = [
...
+  name = "k8s.io/code-generator"
+  non-go = false
+  unused-packages = false
+
+[[prune.project]]
+  name = "github.com/kevinburke/go-bindata"
```
pravisankar: For cluster-ingress-operator repo, we could change "github.com/kevinburke/go-bindata" dep from `[[constraint]]` to `[[prune.project]]`

imcsk8:
LGTM

Gopkg.toml
ironcladlou: Indentation isn't significant in TOML.

Gopkg.toml
ironcladlou: Yeah, ingress could use some cleaning up with pruning rules.

ironcladlou:
>Any reason not to make "all" a phony target? Otherwise looks fine.

I went ahead and fixed it

pravisankar:
We should also make corresponding `.PHONY: all` change in cluster-ingress-operator Makefile.

PR #41, opened 2018-10-23
Rename operator and dns namespaces

pravisankar:
* openshift-cluster-dns-operator => openshift-dns-operator
* openshift-cluster-dns => openshift-dns


imcsk8:
LGTM

manifests/0000_08_cluster-dns-operator_02-deployment.yaml
```diff
@@ -28,5 +28,5 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
-              value: "cluster-dns-operator"
-      serviceAccountName: cluster-dns-operator
+              value: "openshift-dns-operator"
```
Miciah: The "openshift-" prefix stands out to me as different from what we did with the ingress operator.  Is the "openshift-" prefix there deliberately to distinguish the operator from kube-dns?

ironcladlou:
Are we past the point of no return with this one? If we can knock it out in a day I say let's get it over with, otherwise let's close and move on.

pravisankar:
/hold
openshift-cluster-dns-operator is referenced outside of this repo in these 2 places:
(1) https://github.com/openshift/origin/blob/c5769124a7bcbefb921fc1a2ba4585adff0dd481/test/extended/operators/operators.go#L151
(2) https://github.com/openshift/installer/blob/b4faf68e3767b5d0f78c77ec437f726d7f11ff78/docs/user/troubleshooting.md

We could do these steps to avoid breaking other repos/CI:
- Merge https://github.com/openshift/origin/pull/21623
- Merge this pr
- Follow up prs to patch (1) and (2)

pravisankar:

manifests/0000_08_cluster-dns-operator_02-deployment.yaml
ironcladlou: I think it could just be `dns-operator` but I also don't think this affects anything but MAYBE log output?

Miciah: Actually, because we don't use `k8sutil.GetOperatorName()`, it doesn't affect anything.  We can go ahead and delete `OPERATOR_NAME` entirely.

Miciah:
Looks like a flake

    level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
    2018/12/18 10:46:46 Container setup in pod e2e-aws failed, exit code 1, reason Error

/retest

Miciah:
CI problems:

    level=error msg="\t* aws_vpc.new_vpc: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached."

PR #44, opened 2018-11-02
manifests: allow dns operator to schedule on any node

abhinavdahiya:
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts has docs
that define a toleration that tolerates evrything
```
An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything.
```

/cc @ramr

PR #45, opened 2018-11-05
Do no generate bindata for manifests used by CVO

pravisankar:
Needed for this merged pr https://github.com/openshift/cluster-dns-operator/pull/44


pravisankar:
@ironcladlou PTAL

ironcladlou:
How did this get missed by the PR testing which verifies biodata?

ironcladlou:
We don't generate bindata for the root manifests, do we?

pravisankar:
@ironcladlou hack/verify-generated-bindata.sh does show the differences but may be this verify-generated-bindata.sh is not part of the CI merge gate (need to look CI test job)?

ironcladlou:
1. The job is indeed [incorrect](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-dns-operator/openshift-cluster-dns-operator-master.yaml#L33) and should be invoking `make`
2. The [update-generated-bindata.sh](https://github.com/openshift/cluster-dns-operator/blob/master/hack/update-generated-bindata.sh#L27) script is incorrectly processing `/manifests` which are used only by the CVO and are NOT used by our operator and so don't need packaged into bindata.

Let's fix all that.

pravisankar:
@ironcladlou looks like CI is only testing: go test and dns specific e2-aws before merging. We need to add hack/verify-generated-bindata.sh check here: https://github.com/openshift/release/blob/c1d2c5c1405ad1bdf3cf18283ec87a459f405bef/ci-operator/config/openshift/cluster-dns-operator/openshift-cluster-dns-operator-master.yaml#L31

pravisankar:
@ironcladlou  Ok, I will make the change

PR #48, opened 2018-11-09
Fix dns service reconciliation

pravisankar:
- Creating dns service and ignoring if exists is not sufficient as api server
can throw 'provided IP is already allocated' error.


ironcladlou:
Thanks!
/lgtm

PR #50, opened 2018-11-13
Add baseline end-to-end tests

Miciah:
Add baseline end-to-end tests for cluster-dns-operator.

* `Makefile` (`test-e2e`): Add target.
* `test/e2e/operator_test.go`: Add tests.

This commit is related to NE-97.

https://jira.coreos.com/browse/NE-97

PR #51, opened 2018-11-13
e2e tests follow up - wait for operator deployment

ramr:
wait for operator deployment and minor cleanup to use a constant - follow up from #50 
@Miciah  PTAL thx

I gave up waiting on a cluster to come up to test this and figured let's use the e2e tests to test it.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -1,0 +1,55 @@
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func TestMain(m *testing.M) {
+	// When the e2e tests are run, we have no guarantees that the cluster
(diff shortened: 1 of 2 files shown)


test/e2e/main_test.go
```diff
@@ -0,0 +1,55 @@
...
+			APIVersion: appsv1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-cluster-dns-operator",
+			Name:      "cluster-dns-operator",
+		},
+	}
+
+	deploymentExists := func() (bool, error) {
+		if err := sdk.Get(d); err != nil {
```
pravisankar: Exit early if the error is other than object not found?
```
if !errors.IsNotFound(err) {
   return false, err
}
```

ironcladlou:
1. Doesn't polling for the operator's status publishing implicitly test that the deployment occured?
2. In the case of cluster DNS specifically, there's no way our e2e tests are started before cluster DNS is up and running

Just wondering what problem this solves

ramr:
@ironcladlou  so that's the pieces I had in the prior work that I pulled into here. As re: why I did that was since the status and deployment are decoupled, it made sense (to me!!) to check on those separately.

ironcladlou:
Can we close this one out? I'm still not sure it's necessary.

ramr:
Closing it out.

PR #52, opened 2018-11-26
Make CoreDNS image configurable

ironcladlou:
Consume the coredns image from configuration and add it to image-references.


pkg/manifests/manifests_test.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	dnsv1alpha1 "github.com/openshift/cluster-dns-operator/pkg/apis/dns/v1alpha1"
+	"github.com/openshift/cluster-dns-operator/pkg/operator"
 	"github.com/openshift/cluster-dns-operator/pkg/util"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestManifests(t *testing.T) {
-	f := NewFactory()
+	f := NewFactory(operator.Config{CoreDNSImage: "quay.io/openshift/coredns:test"})
```
Miciah: We could use a named constant for the image name used in `TestManifests` and `TestDefaultCluserDNS` and check in `TestManifests` that the daemonset from the factory has the correct image.

PR #55, opened 2018-11-27
Add registry service cluster IP to node resolver

pravisankar:


manifests/0000_08_cluster-dns-operator_03-daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
+
+          # TODO: Is this the correct service name for internal registry?
+          # Is there one global registry or multiple registries for dev, test and external users?
+          REGISTRY_SERVICE="docker-registry.default.svc.cluster.local"
```
ironcladlou: Omit the `cluster.local` since that's configurable and should be in the container resolver search path.

manifests/0000_08_cluster-dns-operator_03-daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  selector:
+    matchLabels:
+      openshift-app: dns-node-resolver
+  template:
+    metadata:
+      labels:
+        openshift-app: dns-node-resolver
+    spec:
+      dnsPolicy: Default
```
ironcladlou: Doesn't this mean the pod won't actually be using the cluster DNS service to resolve the registry name?

manifests/0000_08_cluster-dns-operator_03-daemonset.yaml
```diff
@@ -0,0 +1,56 @@
+# Node resolver for internal registry
```
ironcladlou: If we deploy this as a manifest via CVO, what happens when we need to roll out a bug fix or other change?

pravisankar:
Closing in favor of https://github.com/openshift/cluster-dns-operator/pull/56

PR #56, opened 2018-11-29
Added sidecar container to resolve registry service cluster IP by the node resolver

pravisankar:


pravisankar:

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,44 @@ spec:
...
+          REGISTRY_SERVICE="docker-registry.default.svc.cluster.local"
+          OPENSHIFT_MARKER="openshift-generated-node-resolver"
+          HOSTS_FILE="/etc/hosts"
+          TEMP_FILE="/etc/hosts.tmp"
+
+          while true; do
+            ( grep -v "${OPENSHIFT_MARKER}" "${HOSTS_FILE}"; dig @"${NAMESERVER}" +short "${REGISTRY_SERVICE}" | sed 's/$/ '"${REGISTRY_SERVICE}"' # '"${OPENSHIFT_MARKER}"'/' ) > "${TEMP_FILE}" && \
+            cp "${TEMP_FILE}" "${HOSTS_FILE}"
+            sleep 60
+          done
```
ironcladlou: Another approach to consider which doesn't require `sed` or a subshell:

```bash
# Build a new hosts file from /etc/hosts with our custom entries filtered out
grep -v "${OPENSHIFT_MARKER}" "${HOSTS_FILE}" > "${TEMP_FILE}"

# Append refreshed custom entries to the new hosts file
dig @"${NAMESERVER}" +short "${REGISTRY_SERVICE}"  >> "${TEMP_FILE}"

# Replace /etc/hosts with our modified version
cp -f "${TEMP_FILE}" "${HOSTS_FILE}"
```

pkg/manifests/manifests.go
```diff
@@ -149,8 +149,22 @@ func (f *Factory) DNSDaemonSet(dns *dnsv1alpha1.ClusterDNS) (*appsv1.DaemonSet,
 		return nil, fmt.Errorf("volume 'config-volume' not found")
 	}
 
-	ds.Spec.Template.Spec.Containers[0].Image = f.config.CoreDNSImage
-
+	for i := range ds.Spec.Template.Spec.Containers {
```
ironcladlou: Consider minor refactor:

```go
for c, i := range ds.Spec.Template.Spec.Containers {
  switch c.Name {
    case "dns":
    // do stuff
    case "dns-node-resolver":
    // do stuff
  }
}

File: assets/dns/daemonset.yaml [hunk 3]
```diff
@@ -44,14 +44,44 @@ spec:
           timeoutSeconds: 5
           successThreshold: 1
           failureThreshold: 5
+      - name: dns-node-resolver
+        image: "openshift/origin-cli:v4.0"
```
ironcladlou: This image reference must come from configuration and be represented in `manifest/image-references`.

assets/dns/daemonset.yaml
Miciah: Alternatively, replace parentheses with braces to group commands without forking a subshell.

Miciah: You do still need the host name in the appended hosts entry, but you can avoid using sed:
```bash
printf '%s %s\n' "$(dig "@$NAMESERVER" +short "$REGISTRY_SERVICE")" "$REGISTRY_SERVICE"
```
(`echo` would work too.)

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,56 @@ spec:
...
+
+          REGISTRY_SERVICE="docker-registry.default.svc"
+          OPENSHIFT_MARKER="openshift-generated-node-resolver"
+          HOSTS_FILE="/etc/hosts"
+          TEMP_FILE="/etc/hosts.tmp"
+
+          while true; do
+            # Fetch registry service IP from cluster dns if present
+            ips=($(dig @"${NAMESERVER}" +short "${REGISTRY_SERVICE}.${CLUSTER_DOMAIN}"))
+            if [[ ("$?" -eq 0) && ("${#ips[@]}" -ne 0) ]]; then
```
Miciah: The parentheses are unnecessary.

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,56 @@ spec:
...
+              # Build a new hosts file from /etc/hosts with our custom entries filtered out
+              grep -v "${OPENSHIFT_MARKER}" "${HOSTS_FILE}" > "${TEMP_FILE}"
+
+              # Append resolver entries for registry
+              for ip in "${ips[@]}"; do
+                echo "${ip} ${REGISTRY_SERVICE} # ${OPENSHIFT_MARKER}" >> "${TEMP_FILE}"
+              done
+
+              # Replace /etc/hosts with our modified version
+              cp -f "${TEMP_FILE}" "${HOSTS_FILE}"
```
Miciah: It's traditional to use `mv` in order to perform the replacement atomically.  For `/etc/hosts`, we also want to take care to preserve attributes including SELinux context, by doing something along the lines of the following:
```bash
            if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
              # Make a temporary file with the old hosts file's attributes.
              cp -f --attributes-only "${HOSTS_FILE}" "${TEMP_FILE}"
              
              # Write to $TEMP_FILE using > and >> as usual.
              # ...
              
              # Atomically replace /etc/hosts with the temporary file.
              mv -f "${TEMP_FILE}" "${HOSTS_FILE}"
            fi
```
Based on item 9 at http://www.pixelbeat.org/docs/unix_file_replacement.html.

pkg/manifests/manifests_test.go
```diff
@@ -46,8 +47,34 @@ func TestManifests(t *testing.T) {
...
+				} else if *dns.Spec.ClusterIP != nameserver {
+					t.Errorf("expected NAMESERVER env for dns node resolver image %q, got %q", *dns.Spec.ClusterIP, nameserver)
+				}
+				clusterDomain, ok := envs["CLUSTER_DOMAIN"]
+				if !ok {
+					t.Errorf("CLUSTER_DOMAIN env for dns node resolver image not found")
+				} else if *dns.Spec.ClusterDomain != clusterDomain {
+					t.Errorf("expected CLUSTER_DOMAIN env for dns node resolver image %q, got %q", *dns.Spec.ClusterDomain, clusterDomain)
+				}
+			}
```
Miciah: Maybe also check that both containers, and no additional containers, are present in the daemonset.

assets/dns/daemonset.yaml
Miciah: Design question: If the name does not resolve to anything (i.e., `[[ "${#ips[@]" -eq 0 ]]`), should we still regenerate `/etc/hosts` in order to clear out any stale entries? Or do we want to keep stale entries in order to avoid turning an API or CoreDNS outage into a registry outage? Whatever decision we make on the design, I suggest we add a comment here to document it.

assets/dns/daemonset.yaml
pravisankar: Tried using `mv` from temp file to /etc/hosts before but looks like that is not supported. Got this error: 
`mv: cannot move '/etc/hosts.tmp' to '/etc/hosts': Device or resource busy`
Searching online, it was recommended to use `cp -f` for this use case.

I will copy initial /etc/hosts file attributes when creating the new temporary file to preserve selinux context.

Miciah: Oh, probably because it's a bind mount—atomicity might not be possible here.

Miciah: Related issue: https://github.com/moby/moby/issues/17190

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,64 @@ spec:
...
+          while true; do
+            # Fetch registry service IP from cluster dns if present
+            ips=($(dig @"${NAMESERVER}" +short "${REGISTRY_SERVICE}.${CLUSTER_DOMAIN}"))
+
+            # Update /etc/hosts only if we get valid registry service IPs
+            # We will not update /etc/hosts when there is coredns service outage or api unavailability
+            # Stale entries could exist in /etc/hosts if the registry service is deleted
+            if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
+              # Build a new hosts file from /etc/hosts with our custom entries filtered out
+              grep -v "${OPENSHIFT_MARKER}" "${HOSTS_FILE}" > "${TEMP_FILE}"
```
knobunc: Can we grep for "# ${OPENSHIFT_MARKER}" since it's always a comment?

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,64 @@ spec:
...
+
+              # Append resolver entries for registry
+              for ip in "${ips[@]}"; do
```
knobunc: How can the service name resolve to a list of service IPs (unless it's headless or something weird)?

But I don't mind looping to be safe, but I wouldn't really expect to have more than one.

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,64 @@ spec:
...
+          mountPath: /etc/hosts
+        # env NAMESERVER and CLUSTER_DOMAIN are set at runtime
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
+
+          REGISTRY_SERVICE="docker-registry.default.svc"
```
knobunc: I wonder if we should make this a space separated list in case we find we need to add some services here late in the game?

Then just loop over them to build the entries.  And set it in the env section rather than inline.  We can default this here though if the env is not set.

assets/dns/daemonset.yaml
```diff
@@ -44,14 +44,65 @@ spec:
...
+        volumeMounts:
+        - name: hosts-file
+          mountPath: /etc/hosts
+        # env NAMESERVER and CLUSTER_DOMAIN are set at runtime
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
```
Miciah: If we `set -e`, we will exit immediately if the `dig` command below fails; we should probably just not `set -e`.

assets/dns/daemonset.yaml
pravisankar: done

assets/dns/daemonset.yaml
pravisankar: done

assets/dns/daemonset.yaml
pravisankar: Fixed

PR #62, opened 2018-12-05
Assign system-cluster-critical priority to operator and component

ironcladlou:
Apply the `system-cluster-critical` pod priority class to the operator and
CoreDNS deployments in accordance with the designation of cluster DNS as
a core feature.

Fixes https://github.com/openshift/cluster-dns-operator/issues/57


ironcladlou:

assets/dns/daemonset.yaml
```diff
@@ -15,6 +15,7 @@ spec:
         openshift-app: dns
     spec:
       serviceAccountName: dns
+      priorityClassName: system-cluster-critical
```
Miciah: Should this be `system-node-critical`? Without the operator, the cluster is broken; without the local CoreDNS instance, the node is broken.

ironcladlou: Hmm, could be, especially now that we're responsible for image registry connectivity via /etc/hosts on the node. @aveshagarwal, thoughts?

pravisankar: Yes, I think this should be `system-node-critical` (should never be evicted from the node).

aveshagarwal: +1 for system-node-critical for dns pods.

ironcladlou: Updated to use system-node-critical

Miciah:
/lgtm
Besides flakes, it looks like e2e-aws is blocked on https://github.com/openshift/cluster-dns-operator/pull/58.

PR #64, opened 2018-12-13
Make DNS CRD cluster scoped

pravisankar:
Depends on https://github.com/openshift/cluster-dns-operator/pull/58

Last commit https://github.com/openshift/cluster-dns-operator/pull/64/commits/181310cf3db5304f12ac64831040763d0ac216a5 is relevant to this PR.


pravisankar:

test/e2e/operator_test.go
```diff
@@ -51,8 +51,7 @@ func TestDefaultClusterDNSExists(t *testing.T) {
 			APIVersion: "dns.openshift.io/v1alpha1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "default",
-			Namespace: "openshift-cluster-dns-operator",
+			Name: "openshift-dns",
```
Miciah: This change is in the "Rename operator and dns namespaces" commit but should be in the "Make DNS CRD cluster scoped" commit.

pravisankar:
@openshift/sig-network-edge  Updated, PTAL

assets/cluster-dns-cr.yaml
```diff
@@ -2,8 +2,7 @@
 apiVersion: dns.openshift.io/v1alpha1
 kind: ClusterDNS
 metadata:
-  name: default
-  namespace: openshift-dns-operator
+  name: openshift-dns
```
Miciah: Is there a reason to change the name from "default"? The resource kind and API group are sufficient to describe what it is, and I see other operators tend to use "default" or "instance".  For example, [cluster-network-operator ignores, and logs that it ignores, any networkconfig that is not named "default"](https://github.com/openshift/cluster-network-operator/blob/master@%7B2018-12-19%7D/pkg/controller/networkconfig/networkconfig_controller.go#L85-L89), and [cluster-openshift-apiserver-operator only looks for openshiftapiserveroperatorconfigs named "instance"](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master@%7B2018-12-19%7D/pkg/operator/workload_controller.go#L103).  We also use the name "default" for [the default clusteringress](https://github.com/openshift/cluster-ingress-operator/blob/master@%7B2018-12-19%7D/assets/defaults/cluster-ingress.yaml#L6).

ironcladlou: Changing this has ripple effects, let's not change it in this PR

ironcladlou:
Needs rebased

lihongan:
Why need to make DNS CRD cluster scoped? For cluster scoped CRD, I think it should also be updated to use "config.openshift.io/v1" apiVersion, right?

Miciah:
@lihongan,
> Why need to make DNS CRD cluster scoped?

DNS is configured by the cluster administrator, for the entire cluster, so it does not make sense for it to be namespace-scoped.  Moreover, even if the operator supported namespace-scoped clusterdnses by deploying a namespace-specific CoreDNS instance, the kubelet does not support namespace-specific DNS configuration (concretely, the kubelet will write `/etc/resolv.conf` with the same nameserver for all containers in all namespaces).  So having clusterdnses be namespace-scoped does not fit the common use-case and does not fit well with the technical reality.

Making clusterdnses cluster-scoped will also enable us to use owner references.  Currently we cannot use owner references to associate the clusterdns in the openshift-dns-operator namespace with the associated daemonset etc. in the openshift-dns namespace because cross-namespace owner references do not consistently work in Kubernetes.

> For cluster scoped CRD, I think it should also be updated to use "config.openshift.io/v1" apiVersion, right?

The dnses.config.openshift.io/cluster object describes cluster configuration.  The clusterdnses.dns.openshift.io/default object describes application configuration, which the operator uses to deploy the CoreDNS application.  All operators, including "singleton" operators (i.e., operators that manage a single CR; see https://github.com/openshift/cluster-dns-operator/pull/64#discussion_r243028255 for some examples), work off of a CR that describes some application and the desired configuration for that application.  An operator may read config objects, but it acts on its CR.

Miciah:
/retest
Flakes:
* openshift-tests [Conformance][Area:Networking][Feature:Router] The HAProxy router converges when multiple routers are writing conflicting status [Suite:openshift/conformance/parallel/minimal] [Suite:openshift/smoke-4] 2m17s
```
fail [github.com/openshift/origin/test/extended/router/stress.go:253]: Expected
    <int>: 42
to be <=
    <int>: 40
```
* openshift-tests [Feature:Platform][Suite:openshift/smoke-4] Managed cluster should have no crashlooping pods in core namespaces over two minutes [Suite:openshift/conformance/parallel] 2m3s
```
fail [github.com/openshift/origin/test/extended/operators/cluster.go:109]: Expected
    <[]string | len:1, cap:1>: [
        "Pod openshift-cluster-machine-approver/machine-approver-6bcbc5dbf5-kjwpb is not healthy: container machine-approver-controller has restarted more than 5 times",
    ]
to be empty
```

Issue #65, opened 2018-12-18
Node DNS sync container takes 30-60s to shut down because it doesn't handle SIGTERM properly


smarterclayton:
Bash processes run as pid 1 need to follow a certain set of patterns in order to avoid being unresponsive when the container runtime invokes them.

Right now, the sidecar sync container is ignoring SIGTERM and so pod deletion takes however long a single wait lasts, instead of happening near instantly.  This is frustrating and delays upgrade and can lead to extended downtime on the node for DNS.

Specifically, you need to

1. Listen for SIGTERM explicitly: https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_node_group/files/sync.yaml#L59
2. Always `sleep X & wait` instead of just `sleep X`: https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_node_group/files/sync.yaml#L83
3. If you do run sub processes, be careful to wait on the right object (some examples exist, I don't have them at hand).

pravisankar:
Fixed by https://github.com/openshift/cluster-dns-operator/pull/68
/close

PR #68, opened 2019-01-08
Allow node resolver sidecar container to exit in case of SIGTERM signal

pravisankar:
Fixes issue: https://github.com/openshift/cluster-dns-operator/issues/65


pravisankar:

ironcladlou:
Looks like this does the job; a once-over from some resident shell wizards would be greatly appreciated (/cc @Miciah @wking)

Miciah:
Is there a race condition if a SIGTERM comes between the `sleep` and the `wait`? In any case, I don't know a better approach, and this is the prescribed solution, so
/lgtm

PR #74, opened 2019-02-06
manifests: Add ClusterOperator resource

ironcladlou:
This makes the operator conform to:

https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

### DIFF
--- a/manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
+++ b/manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
@@ -1,0 +1,5 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: dns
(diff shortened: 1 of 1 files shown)


manifests/0000_08_cluster_operator.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: ingress
```
Miciah: Bad copypasta!

Miciah: bootstrap-complete timed out, presumably because it was looking for the "ingress" clusteroperator instead of "dns".

pravisankar: yes, should be dns

PR #75, opened 2019-02-07
config: Replace install config with cluster config

ironcladlou:
Replace usage of the deprecated install configmap with the config.openshift.io
`Network` API type.


ironcladlou:
@openshift/sig-network-edge was missing an RBAC rule, PTAL

PR #76, opened 2019-02-12
Set semantic versions in clusteroperator status

Miciah:
## Add `coreos/go-semver` dependency

* `Gopkg.lock`:
* `Gopkg.toml`:
* `vendor/github.com/coreos/go-semver`: Import package in order to parse and compare semantic versions.

## Set semantic versions in clusteroperator status

This commit resolves [NE-147](https://jira.coreos.com/browse/NE-147).

* `cmd/cluster-dns-operator/main.go` (`main`): Read `OPERATOR_IMAGE_VERSION` and add use its value to set `OperatorImageVersion` in the operator config.
* `manifests/0000_08_cluster-dns-operator_01-version-mapping.yaml`: New file with the `version-mapping` configmap.
* `manifests/0000_08_cluster-dns-operator_02-deployment.yaml`: Add `OPERATOR_IMAGE_VERSION` environment variable.
* `pkg/operator/config.go` (`Config`): Add `OperatorImageVersion` field.
* `pkg/stub/status.go` (`syncOperatorStatus`): Use new functions `getVersionMap` and `computeStatusVersions` to set `status.versions`.
(`getVersionMap`): New function to get the `version-mapping` configmap.
(`computeStatusVersions`): New function to compute `status.versions` for the operator and its operands given the operator image, daemonsets, and version map, using `getSemVerForPullspec`.
(`getSemVerForPullspec`): New function to get the semantic version from the pullspec, using the version map.
* `pkg/stub/status_test.go` (`TestComputeStatusVersions`): New test.
* `version/version.go`: Deleted.

## Get new clusteroperator before updating status

The operator must get a newly created clusteroperator from the API before attempting to update its status, or else the attempted update will fail with the following error:

    metadata.resourceVersion: Invalid value: 0x0: must be specified for an update

* `pkg/stub/status.go` (`syncOperatorStatus`): Get the clusteroperator after creating it.

## Improve logging and formatting in status code

Change some log messages and code formatting to be more readable and stylistically consistent.

* `pkg/stub/status.go` (`syncOperatorStatus`, `getOperatorState`): Unwrap long lines, reformat some statements, delete blank lines before returns, improve some log messages, and use `%q` when printing the name of a resource.
(`computeStatusConditions`): Unwrap long lines.
* `pkg/stub/status_test.go` (`TestComputeStatusConditions`): Unwrap or reformat some long lines.

## Use `SchemeGroupVersion.String()` to get API versions

[...]

PR #77, opened 2019-02-12
Move clusterIP and clusterDomain spec fields to status

pravisankar:
https://jira.coreos.com/browse/NE-160


pravisankar:

Miciah:
Initializing status in the factory method seems wrong: the responsibility of the factory should be to provide a resource with metadata and spec, and it should be up to the reconciliation logic (perhaps `ensureCoreDNSForClusterDNS`, or some new method that the `reconcile` loop would invoke) to update the status; does that seem reasonable?

In addition, what do you think about enabling the `/status` subresource on the CRD in order to get correct semantics for updates to the resource metadata and spec versus updates to its its status?

pkg/apis/dns/v1alpha1/types.go
```diff
@@ -22,10 +25,13 @@ type ClusterDNS struct {
 }
 
 type ClusterDNSSpec struct {
+}
+
+// ClusterDNSStatus defines the observed state of ClusterDNS
+type ClusterDNSStatus struct {
+	// ClusterIP is the service IP reserved for cluster DNS service
 	ClusterIP *string `json:"clusterIP"`
```
ironcladlou: Should these still be pointers?

pkg/manifests/manifests.go
```diff
@@ -196,8 +174,8 @@ func (f *Factory) DNSService(dns *dnsv1alpha1.ClusterDNS) (*corev1.Service, erro
 	}
 	s.Spec.Selector["dns"] = s.Name
 
-	if dns.Spec.ClusterIP != nil {
-		s.Spec.ClusterIP = *dns.Spec.ClusterIP
+	if len(clusterIP) > 0 {
+		s.Spec.ClusterIP = clusterIP
```
ironcladlou: I must be missing something, didn't `Spec` get deleted? surprised this compiles

pkg/stub/handler.go
```diff
@@ -175,7 +166,19 @@ func (h *Handler) ensureDNSNamespace() error {
...
-	ds, err := h.ManifestFactory.DNSDaemonSet(dns)
+	clusterIP := ""
+	// TODO: fetch this from higher level openshift resource when it is exposed
+	clusterDomain := "cluster.local"
+
+	if dns.Name == "default" {
+		var err error
+		clusterIP, err = getClusterIPFromNetworkConfig()
+		if err != nil {
+			return fmt.Errorf("failed to getch cluster IP from network config for clusterdns %s, %v", dns.Name, err)
```
Miciah: Typo: "getch".

pkg/stub/handler.go
```diff
@@ -175,7 +166,19 @@ func (h *Handler) ensureDNSNamespace() error {
 // ensureCoreDNSForClusterDNS ensures all necessary CoreDNS resources exist for
 // a given clusterdns.
 func (h *Handler) ensureCoreDNSForClusterDNS(dns *dnsv1alpha1.ClusterDNS) error {
-	ds, err := h.ManifestFactory.DNSDaemonSet(dns)
+	clusterIP := ""
+	// TODO: fetch this from higher level openshift resource when it is exposed
+	clusterDomain := "cluster.local"
+
+	if dns.Name == "default" {
```
Miciah: I'm ambivalent about special-casing on the clusterdns's name.  For now, clusterdns is a singleton, right? Does it make sense to special-case the name now? Will it make sense to special-case the name in the future, if we expand the API to support multiple clusterdnses?

pkg/stub/handler.go
```diff
@@ -240,3 +244,58 @@ func (h *Handler) ensureDNSDeleted(dns *dnsv1alpha1.ClusterDNS) error {
...
+// syncClusterDNSStatus updates the status for a given clusterdns.
+func syncClusterDNSStatus(dns *dnsv1alpha1.ClusterDNS, clusterIP, clusterDomain string) error {
+	dns.Status = dnsv1alpha1.ClusterDNSStatus{
+		ClusterIP:     clusterIP,
+		ClusterDomain: clusterDomain,
+	}
+
+	unstructObj, err := k8sutil.UnstructuredFromRuntimeObject(dns)
+	if err != nil {
+		return fmt.Errorf("failed to get unstructed obj from runtime: %v", err)
```
Miciah: Typo: "unstructed" should be "unstructured".  (In https://github.com/openshift/cluster-dns-operator/pull/76/commits/2f85300f92b76f3723b5fb7b01d3ba136aff0e34#diff-f33ebabd902db5bc0032360e4c24c3c8R97, I used the phrasing "failed to convert"—not sure what phrasing most useful in the logs.)

pkg/stub/handler.go
```diff
@@ -240,3 +244,58 @@ func (h *Handler) ensureDNSDeleted(dns *dnsv1alpha1.ClusterDNS) error {
...
+			Name: "cluster",
+		},
+	}
+	err := sdk.Get(networkConfig)
+	if err != nil {
+		return "", fmt.Errorf("failed to get network 'cluster': %v", err)
+	}
+
+	if len(networkConfig.Status.ServiceNetwork) == 0 {
+		return "", fmt.Errorf("no service networks found in cluster network config")
```
Miciah: :+1:, adding this check is good.

pkg/manifests/manifests.go
Miciah: That's setting `.spec.clusterIP` on the service, not the clusterdns.

pravisankar:
@openshift/sig-network-edge ready for review/merge, PTAL

PR #78, opened 2019-02-28
status: report cluster operator version from payload

derekwaynecarr:
Every operator shall report a version with name=operator whose value is equal to the RELEASE_VERSION provided by the cluster version operator.

will be enforced via [openshift/origin#22156](https://github.com/openshift/origin/pull/22156)

### DIFF
--- a/manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
+++ b/manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
@@ -3,3 +3,7 @@ kind: ClusterOperator
 metadata:
   name: dns
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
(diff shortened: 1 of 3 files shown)


manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
```diff
@@ -3,3 +3,7 @@ kind: ClusterOperator
 metadata:
   name: dns
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
```
Miciah: Is it valid to have an entry for the operator in `.status.versions` when the available condition is not true? According to [the godoc](https://github.com/openshift/api/blob/f1796006a30cb15d8ffd7f0584bdf87bedaf9e4f/config/v1/types_cluster_operator.go#L59-L60), the version "must always match the Available condition."

derekwaynecarr: this manifest is not directly applied, it is only used to signal to the cluster version operator.

smarterclayton:
Without updates being support of the daemonset this is still broken.

As discussed with Derek yesterday this can only happen after you verify the change has been rolled out, and the current code structure can’t guarantee that.

Please ensure you fix daemonset updates and then we need to come back and fix it so that you correctly observe rollout before marking level.

PR #79, opened 2019-03-01
Check RELEASE_VERSION env existence before starting the reconciliation.

pravisankar:
- Removed version/version.go, no longer used.

Follow up pr for https://github.com/openshift/cluster-dns-operator/pull/78

### DIFF
--- a/cmd/cluster-dns-operator/main.go
+++ b/cmd/cluster-dns-operator/main.go
@@ -49,8 +49,9 @@ func main() {
 	}
 
 	operatorConfig := operator.Config{
-		CoreDNSImage:      coreDNSImage,
-		OpenshiftCLIImage: cliImage,
+		OperatorReleaseVersion: os.Getenv("RELEASE_VERSION"),
+		CoreDNSImage:           coreDNSImage,
+		OpenshiftCLIImage:      cliImage,
 	}
 
 	handler := &stub.Handler{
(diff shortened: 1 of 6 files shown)


cmd/cluster-dns-operator/main.go
```diff
@@ -39,6 +39,10 @@ func main() {
 	// namespace from config.
 	sdk.Watch("apps/v1", "DaemonSet", "openshift-dns", resyncPeriod)
 
+	releaseVersion := os.Getenv("RELEASE_VERSION")
+	if len(releaseVersion) == 0 {
+		logrus.Fatalf("RELEASE_VERSION environment variable is required")
```
Miciah: I don't think this should be a fatal error.  The description for https://github.com/openshift/cluster-image-registry-operator/pull/219 has the following:

> If no release version is set we reset versions to empty.
> 
> This is the model we will be implementing across all operators for 4.0.

So we can read it and store it in `operator.Config`, but we shouldn't treat its absence as fatal, and we should continue to conditionalize on non-zero length in `syncOperatorStatus`.

pravisankar:

manifests/0000_08_cluster-dns-operator_03-cluster-operator.yaml
```diff
@@ -4,6 +4,4 @@ metadata:
   name: dns
 spec: {}
 status:
-  versions:
-    - name: operator
-      version: "0.0.1-snapshot"
\ No newline at end of file
+ # status is set at runtime
```
Miciah: `status: {}` would be more consistent, though I'm not sure whether it's more correct.  However, I'm not sure we want to remove the `versions:` stanz (I asked Derek about it here: https://github.com/openshift/cluster-dns-operator/pull/78#discussion_r261258848).

PR #81, opened 2019-03-07
Allow dns operator to update coreDNS image

pravisankar:
- Use get resource and if not found, create the resource pattern for all DNS resources
- Added log info messages for create and delete resources related to DNS
- Update daemonset when the operator publishes new core dns image
- Added test cases for validating coreDNS image update and operator version update

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1673989

### DIFF
--- a/pkg/stub/status.go
+++ b/pkg/stub/status.go
@@ -37,14 +37,12 @@ func (h *Handler) syncOperatorStatus() {
 	if err != nil && !isNotFound {
 		logrus.Errorf("syncOperatorStatus: error getting ClusterOperator %s: %v",
 			co.Name, err)
-
 		return
 	}
 
 	ns, dnses, daemonsets, err := h.getOperatorState()
 	if err != nil {
 		logrus.Errorf("syncOperatorStatus: getOperatorState: %v", err)
-
 		return
 	}
 
@@ -102,7 +100,6 @@ func (h *Handler) syncOperatorStatus() {
 	unstructObj, err := k8sutil.UnstructuredFromRuntimeObject(co)
 	if err != nil {
 		logrus.Errorf("syncOperatorStatus: k8sutil.UnstructuredFromRuntimeObject: %v", err)
-
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


pravisankar:

pkg/stub/status.go
```diff
@@ -83,12 +81,10 @@ func (h *Handler) syncOperatorStatus() {
...
 	if isNotFound {
 		if err := sdk.Create(co); err != nil {
-			logrus.Errorf("syncOperatorStatus: failed to create ClusterOperator %s: %v",
-				co.Name, err)
-		} else {
-			logrus.Infof("syncOperatorStatus: created ClusterOperator %s (UID %v)",
-				co.Name, co.UID)
+			logrus.Errorf("syncOperatorStatus: failed to create ClusterOperator %s: %v", co.Name, err)
 		}
+		logrus.Infof("syncOperatorStatus: created ClusterOperator %s (UID %v)", co.Name, co.UID)
```
Miciah: Need a `return` after the `logrus.Errorf`.

test/e2e/operator_test.go
```diff
@@ -64,3 +68,108 @@ func TestDefaultClusterDNSExists(t *testing.T) {
...
+			}
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to observe updated version reported in dns clusteroperator status: %v", err)
+	}
+}
+
+func TestCoreDNSImageUpgrade(t *testing.T) {
```
ironcladlou: Although I think this probably works for the intended case, I'm afraid it's possible that it _can_ break the deployment for any future subsequent tests because the `openshift/origin-coredns:latest` may not be valid in all test contexts (plus who knows what image we'll actually get). If the only assertion is that the deployment renders the new image, maybe a unit test would make more sense?

ironcladlou: @Miciah had the idea to restore the original image in a defer to avoid having to refactor for unit testing.

pravisankar:
I will create a follow up pr to address the feedback on https://github.com/openshift/cluster-dns-operator/pull/81#pullrequestreview-211777706

test/e2e/operator_test.go
pravisankar: https://github.com/openshift/cluster-dns-operator/pull/83

pkg/stub/status.go
pravisankar: https://github.com/openshift/cluster-dns-operator/pull/83

PR #82, opened 2019-03-08
DNS operator runs after the apiserver and so doesn't need to skip admission/SCC checks

pravisankar:
Same as https://github.com/openshift/cluster-ingress-operator/pull/160


pravisankar:
Looks like we need the privileged runlevel for DNS operator

PR #83, opened 2019-03-13
Restore old coreDNS image and release version in operator e2e tests

pravisankar:


pravisankar:

Miciah:
error: error pulling image registry.svc.ci.openshift.org/ci/clonerefs@sha256:835268c14fbcad843bc9752ba579ec3be2e7d19c94f6dfc627a3dacfeaf1afce: received unexpected HTTP status: 500 Internal Server Error

/retest

test/e2e/operator_test.go
```diff
@@ -91,7 +92,31 @@ func TestVersionReporting(t *testing.T) {
 		t.Errorf("failed to get deployment %s/%s: %v", deployment.Namespace, deployment.Name, err)
 	}
 
-	patch := []byte(`{"spec": {"template": {"spec": {"containers": [{"name":"dns-operator","env":[{"name":"RELEASE_VERSION","value":"0.0.1-test"}]}]}}}}`)
+	numContainers := len(deployment.Spec.Template.Spec.Containers)
+	if numContainers != 1 {
+		t.Errorf("expected operator deployment to contain 1 container but found %d", numContainers)
+	}
```
Miciah: This doesn't hurt but shouldn't be necessary—the API says there must be at least one container.

test/e2e/operator_test.go
```diff
@@ -91,7 +92,31 @@ func TestVersionReporting(t *testing.T) {
...
+	if len(curVersion) == 0 {
+		t.Errorf("env RELEASE_VERSION not found in the operator deployment")
+	}
+
+	defer func() {
+		patch := []byte(fmt.Sprintf(`{"spec": {"template": {"spec": {"containers": [{"name":"dns-operator","env":[{"name":"RELEASE_VERSION","value":"%s"}]}]}}}}`, curVersion))
+		if err := sdk.Patch(deployment, types.StrategicMergePatchType, patch); err != nil {
+			t.Fatalf("failed to restore dns operator to old release version: %v", err)
+		}
+	}()
```
Miciah: Could the `sdk.Patch` below mutate the object and then return a non-nil error value? If not, we could put this `defer` statement with the restorative `sdk.Patch` after the initial `sdk.Patch`.  But what you have is safer if `sdk.Patch` could possibly fail even after patching the object.

pravisankar:
@Miciah updated, PTAL

PR #86, opened 2019-03-28
Cluster DNS repo migration to controller-runtime and operator/v1 DNS API

pravisankar:


pravisankar:

assets/dns/daemonset.yaml
```diff
@@ -1,18 +1,10 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  # Name is set at runtime
+  # name and lables are set at runtime
```
Miciah: "lables"→"labels"

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -0,0 +1,186 @@
...
+	daemonset := &appsv1.DaemonSet{}
+	name := DNSDaemonSetName(dns)
+	daemonset.Name = name.Name
+	daemonset.Namespace = name.Namespace
+	if err := r.client.Delete(context.TODO(), daemonset); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	logrus.Infof("deleted dns daemonset: %s", dns.Name)
```
Miciah: I kind of feel like this log message should be in the `else` for the preceding `if`:
```go
	if err := r.client.Delete(context.TODO(), daemonset); err != nil {
		if !errors.IsNotFound(err) {
			return err
		}
	} else {
		logrus.Infof("deleted dns daemonset: %s", dns.Name)
	}
	return nil
```
What do you think?

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Ensure the daemonset adopts only its own pods.
+	daemonset.Spec.Selector = DNSDaemonSetPodSelector(dns)
+	daemonset.Spec.Template.Labels = daemonset.Spec.Selector.MatchLabels
+
+	coreFileVolumeFound := false
+	for i := range daemonset.Spec.Template.Spec.Volumes {
+		// TODO: remove hardcoding of volume name
+		if daemonset.Spec.Template.Spec.Volumes[i].Name == "config-volume" {
+			daemonset.Spec.Template.Spec.Volumes[i].ConfigMap.Name = dns.Name
```
Miciah: Use `DNSConfigMapName`.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		for i, c := range expected.Spec.Template.Spec.Containers {
+			if name == c.Name {
+				expImage = expected.Spec.Template.Spec.Containers[i].Image
+				break
+			}
+		}
+
+		if len(curImage) == 0 {
+			return true, expected
```
Miciah: How can `curImage` be empty? Should we add a error return value for this contingency, or at least log it?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,377 @@
...
+			logrus.Infof("cluster dns not found; reconciliation will be skipped for request: %v", request)
+		} else {
+			errs = append(errs, fmt.Errorf("failed to get cluster dns %s: %v", request, err))
+		}
+		dns = nil
+	}
+
+	if dns != nil {
+		if dns.Name != DefaultDNSController {
+			// We do not want to return this error to avoid re-triggering the event
```
Miciah: Maybe clearer: "Return a nil error value to avoid re-triggering the event."

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,377 @@
...
+			logrus.Errorf("skipping unexpected cluster dns %s", dns.Name)
+			return result, utilerrors.NewAggregate(errs)
```
Miciah: This is technically correct because `errs` must be empty so `utilerrors.NewAggregate(errs)` will return nil, but using `return result, nil` would be more obvious.

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,377 @@
...
+			// resource that got deleted async).
+			logrus.Infof("cluster dns not found; reconciliation will be skipped for request: %v", request)
+		} else {
+			errs = append(errs, fmt.Errorf("failed to get cluster dns %s: %v", request, err))
+		}
+		dns = nil
+	}
+
+	if dns != nil {
+		if dns.Name != DefaultDNSController {
```
Miciah: Why not do this check before the `Get`? It would simplify the logic a good deal.

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,377 @@
...
+
+	// TODO: Should this be another controller?
+	if err := r.syncOperatorStatus(); err != nil {
+		errs = append(errs, fmt.Errorf("failed to sync operator status: %v", err))
+	}
+
+	if len(errs) > 0 {
+		logrus.Errorf("failed to reconcile request %s: %v", request, utilerrors.NewAggregate(errs))
+	}
+	return result, utilerrors.NewAggregate(errs)
```
Miciah: Logging and returning `utilerrors.NewAggregate(errs)` seems redundant—are we doing that because the controller's logs get eaten? If so, please explain in a comment.

pkg/operator/controller/controller_dns_daemonset.go
pravisankar: Thanks, this fixed the dns container stuck on ContainerCreating issue.

pkg/operator/controller/controller_dns_daemonset.go
pravisankar: We do not expect this to happen, added logging error in this case.

assets/dns/daemonset.yaml
pravisankar: fixed

pkg/operator/controller/controller_dns_daemonset.go
pravisankar: done

pkg/operator/controller/controller.go
pravisankar: Yes, controller's logs get eaten in this case. Added a comment for this.

pravisankar:
Removed WIP tag, tested basic functionality and unit/e2e tests returned success.
@openshift/sig-network-edge ready for review/merge PTAL

pkg/operator/controller/controller_dns_daemonset.go
Miciah: The logging is an improvement.  However, I think that returning `expected` to `updateDNSDaemonSet` will cause the latter to fail when it tries to update the daemonset, because `expected` came from `desiredDNSDaemonSet` and does not have the resource version etc., right?  Probably the best thing to do here is to copy the full containers definition from `expected` to `updated` and then return `updated`:
```go
		if len(curImage) == 0 {
			logrus.Errorf("current daemonset %s/%s did not contain expected %s container", current.Namespace, current.Name, name)
			updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
			changed = true
			break
```
Setting `changed = true`and using `break` means we continue with the usual control flow, which means we'll get the logic for updating volumes too, if and when we implement that, before we reach the final `return true, updated`.

pkg/operator/operator.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	// Create additional controller event sources from informers in the managed
+	// namespace. Any new managed resources outside the operator's namespace
+	// should be added here.
+	mapper, err := apiutil.NewDiscoveryRESTMapper(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get API Group-Resources")
+	}
+	operandCache, err := cache.New(kubeConfig, cache.Options{Namespace: "openshift-dns", Scheme: scheme, Mapper: mapper})
```
Miciah: I think we could configure the manager to watch the `openshift-dns` namespace instead of creating a separate cache. 

I am guessing you are creating `operandCache` because cluster-ingress-operator does.  In the case of cluster-ingress-operator, the operator watches ingresscontrollers in `openshift-ingress-operator` and deployments and services in the `openshift-ingress` namespace, so because it is watching namespace-scoped resources in two different namespaces, it needs two different caches.  However, because dnses are cluster-scoped, cluster-dns-operator only needs to watch one namespace (`openshift-dns`), so I think we can do without `operandCache` in cluster-dns-operator if we configure the manager for `openshift-dns`.

Miciah: We can do this in a follow-up though.

pkg/operator/controller/controller_dns_daemonset.go
pravisankar: fixed

pkg/operator/operator.go
pravisankar: I will submit a follow-up pr for this.

pkg/manifests/manifests.go
```diff
@@ -2,182 +2,92 @@ package manifests
...
+	DNSClusterRoleAsset        = "assets/dns/cluster-role.yaml"
+	DNSClusterRoleBindingAsset = "assets/dns/cluster-role-binding.yaml"
+	DNSConfigMapAsset          = "assets/dns/configmap.yaml"
+	DNSDaemonSetAsset          = "assets/dns/daemonset.yaml"
+	DNSServiceAsset            = "assets/dns/service.yaml"
+
+	// OwningClusterDNSLabel should be applied to any objects "owned by" a
+	// cluster dns to aid in selection (especially in cases where an ownerref
+	// can't be established due to namespace boundaries).
+	OwningClusterDNSLabel = "dns.openshift.io/cluster-dns"
```
Miciah: This is an entirely new label, right? Can we use `operator.openshift.io/dns`?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,378 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// DefaultDNSController is the name of the default DNS instance.
+	DefaultDNSController = "default"
+
+	// DNSControllerFinalizer is applied to an DNS before being considered for processing;
+	// this ensures the operator has a chance to handle all states.
+	DNSControllerFinalizer = "dns.openshift.io/dns-controller"
```
Miciah: Since we are changing this from `dns.openshift.io/default-cluster-dns`, can we change it to `operator.openshift.io/dns`?

assets/dns/daemonset.yaml
```diff
@@ -1,18 +1,10 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  # Name is set at runtime
+  # name and labels are set at runtime
   namespace: openshift-dns
```
Miciah: `desiredDNSDaemonSet` sets the namespace, so it can be removed from the asset.

Miciah:
Any reason to keep the receivers on `DNSServiceAccount`, `DNSClusterRole`, and `DNSClusterRoleBinding`? Looks like we could do away with `manifests.Factory` entirely, and you mad some changes in that direction but did not update those three methods, and I wonder why.  This can be a follow-up PR though.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -0,0 +1,190 @@
...
+	if !coreFileVolumeFound {
+		return nil, fmt.Errorf("volume 'config-volume' is not found")
+	}
+
+	for i, c := range daemonset.Spec.Template.Spec.Containers {
+		switch c.Name {
+		case "dns":
+			daemonset.Spec.Template.Spec.Containers[i].Image = r.CoreDNSImage
+		case "dns-node-resolver":
+			daemonset.Spec.Template.Spec.Containers[i].Image = r.OpenshiftCLIImage
```
Miciah: Ideally `r.CoreDNSImage` and `r.OpenshiftCLIImage` would be inputs to `desiredDNSDaemonSet`, and `desiredDNSDaemonSet` would not take the reconciler as a receiver.  Not sure it's worth worrying about in this PR.

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,378 @@
...
+			if err := r.ensureOpenshiftExternalNameServiceDeleted(); err != nil {
+				errs = append(errs, fmt.Errorf("failed to delete external name for openshift service: %v", err))
+			}
+			if err := r.ensureDNSDeleted(dns); err != nil {
+				errs = append(errs, fmt.Errorf("failed to ensure deletion for cluster dns %s: %v", dns.Name, err))
+			}
+
+			if len(errs) == 0 {
+				// Clean up the finalizer to allow the cluster dns to be deleted.
+				updated := dns.DeepCopy()
```
Miciah: Can move this inside the `if`.

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,378 @@
...
+				if slice.ContainsString(dns.Finalizers, DNSControllerFinalizer) {
+					updated.Finalizers = slice.RemoveString(updated.Finalizers, DNSControllerFinalizer)
+					if err := r.client.Update(context.TODO(), updated); err != nil {
+						errs = append(errs, fmt.Errorf("failed to remove finalizer from cluster dns %s: %v", dns.Name, err))
+					}
+				}
+			}
+		} else if err := r.enforceDNSFinalizer(dns); err != nil {
```
Miciah: I think this fixes an unreported defect, similar to [BZ#1682922](https://bugzilla.redhat.com/show_bug.cgi?id=1682922) for cluster-ingress-operator: if someone had created the dns without a finalizer, we would have failed to clean up related resources.

pkg/manifests/manifests.go
pravisankar: done

pkg/operator/controller/controller.go
pravisankar: done

assets/dns/daemonset.yaml
pravisankar: done

pravisankar:
> Any reason to keep the receivers on `DNSServiceAccount`, `DNSClusterRole`, and `DNSClusterRoleBinding`? Looks like we could do away with `manifests.Factory` entirely, and you made some changes in that direction but did not update those three methods, and I wonder why. This can be a follow-up PR though.

I was planning to remove Factory struct entirely from manifests but forgot about that thought while fixing some other stuff for this pr. Thanks!

pkg/operator/controller/controller_dns_daemonset.go
pravisankar: done

pkg/operator/controller/controller.go
pravisankar: done

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,377 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// DefaultDNSController is the name of the default DNS instance.
+	DefaultDNSController = "default"
+
+	// DNSControllerFinalizer is applied to an DNS before being considered for processing;
```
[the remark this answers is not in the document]
pravisankar: fallout from s/IngressController/DNS/g ...fixed!

pkg/operator/controller/names.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// controllerDaemonSetLabel identifies a daemonset as an dns
```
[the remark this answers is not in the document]
pravisankar: fixed

Miciah:
> I was planning to remove Factory struct entirely from manifests but forgot about that thought while fixing some other stuff for this pr. Thanks!

You're welcome, and thank you for cleaning this up! One more thing: Any reason `DNSServiceAccount`, `DNSClusterRole`, and `DNSClusterRoleBinding` need to return error values?

pravisankar:
> > I was planning to remove Factory struct entirely from manifests but forgot about that thought while fixing some other stuff for this pr. Thanks!
> 
> You're welcome, and thank you for cleaning this up! One more thing: Any reason `DNSServiceAccount`, `DNSClusterRole`, and `DNSClusterRoleBinding` need to return error values?

No specific reason. I wasn't clear about benefits of using panic over returning error in this case. Both will throw errors in the log and will retry (panic restarts container and in case of error we will retry with exponential backoff). I have converted to panic for consistency and probably we could record an event or something before bailout in the future so that the operator admin can get notified and can take actionable task.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -0,0 +1,190 @@
...
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return fmt.Errorf("failed to update dns daemonset %s/%s: %v", updated.Namespace, updated.Name, err)
+	}
+	logrus.Infof("updated dns daemonset: %s/%s", updated.Namespace, updated.Name)
+	return nil
+}
+
+// daemonsetConfigChanged checks if current config matches the expected config
+// for the cluster dns daemonset and if not returns the updated config.
+func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.DaemonSet) {
```
ironcladlou: Note for the future: like with ingress, this function could be doing a more generic diff on `.spec` and if there's any diff simply overwrite the current `.spec.` with a copy of the expected `.spec`.

pkg/operator/operator.go
ironcladlou: Covered by https://github.com/openshift/cluster-dns-operator/pull/87

PR #87, opened 2019-04-04
deps: integrate with latest openshift version

ironcladlou:
* Upgrade to latest openshift/api and libary-go
* Upgrade to kube 1.13
* Upgrade to latest compatible controller-runtime

Fixes openapi spec generation, which as of the previous version was not picking
up field descriptions.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693127


Makefile
```diff
@@ -17,9 +17,19 @@ build:
 	$(GO_BUILD_RECIPE)
 
 .PHONY: generate
-generate:
+generate: crd
 	hack/update-generated-bindata.sh
 
+# Generate ClusterIngress CRD from vendored API spec.
```
Miciah: `ClusterIngress`→`DNS` here and also for the `verify-crd` target below.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -52,6 +52,7 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string, daemonsetRef
 	name := DNSConfigMapName(dns)
 	cm.Namespace = name.Namespace
 	cm.Name = name.Name
+	cm.SetOwnerReferences([]metav1.OwnerReference{dnsOwnerRef(dns)})
```
Miciah: I thought we couldn't do owner references between cluster-scoped and namespace-scoped resources; is it valid to do so but only a problem for garbage collection?

In any case, I think this owner reference gets overridden by `cm.SetOwnerReferences([]metav1.OwnerReference{daemonsetRef})` below.  Assuming the dns owner reference is valid, I think we want both owner references: dns owner reference so that `EnqueueRequestForOwner` enqueues the dns rather than the configmap, and a daemonset owner reference so that garbage collection deletes the configmap when the daemonset is deleted.  Also, we should set `Controller` to false in the `daemonsetRef`.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -61,6 +62,7 @@ func desiredDNSDaemonSet(dns *operatorv1.DNS, clusterIP, clusterDomain, coreDNSI
 	name := DNSDaemonSetName(dns)
 	daemonset.Name = name.Name
 	daemonset.Namespace = name.Namespace
+	daemonset.SetOwnerReferences([]metav1.OwnerReference{dnsOwnerRef(dns)})
```
Miciah: Can we do an owner reference between the cluster-scoped dns and the namespace-scoped daemonset?

pkg/operator/controller/controller_dns_service.go
```diff
@@ -51,6 +51,7 @@ func desiredDNSService(dns *operatorv1.DNS, clusterIP string, daemonsetRef metav
 	name := DNSServiceName(dns)
 	s.Namespace = name.Namespace
 	s.Name = name.Name
+	s.SetOwnerReferences([]metav1.OwnerReference{dnsOwnerRef(dns)})
```
Miciah: Similar comment as for `desiredDNSConfigMap`.

pravisankar:
Need to fix test/e2e/operator_test.go

```
test/e2e/operator_test.go:193:68: cannot use "github.com/openshift/cluster-dns-operator/vendor/sigs.k8s.io/controller-runtime/pkg/client".ListOptions literal (type *"github.com/openshift/cluster-dns-operator/vendor/sigs.k8s.io/controller-runtime/pkg/client".ListOptions) as type runtime.Object in argument to cl.List:
	*"github.com/openshift/cluster-dns-operator/vendor/sigs.k8s.io/controller-runtime/pkg/client".ListOptions does not implement runtime.Object (missing DeepCopyObject method)
test/e2e/operator_test.go:193:68: cannot use podList (type *"github.com/openshift/cluster-dns-operator/vendor/k8s.io/api/core/v1".PodList) as type "github.com/openshift/cluster-dns-operator/vendor/sigs.k8s.io/controller-runtime/pkg/client".ListOptionFunc in argument to cl.List
?   	github.com/openshift/cluster-dns-operator/cmd/cluster-dns-operator	[no test files]
```

pkg/operator/controller/controller_dns_configmap.go
ironcladlou: >I thought we couldn't do owner references between cluster-scoped and namespace-scoped resources; is it valid to do so but only a problem for garbage collection?

Cluster scoped owners are okay. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#ownerreference-v1-meta

>In any case, I think this owner reference gets overridden by cm.SetOwnerReferences([]metav1.OwnerReference{daemonsetRef}) below. 

Good catch

>Assuming the dns owner reference is valid, I think we want both owner references: dns owner reference so that EnqueueRequestForOwner enqueues the dns rather than the configmap, and a daemonset owner reference so that garbage collection deletes the configmap when the daemonset is deleted.

The Corefile's existence doesn't depend upon the existence of the daemonset. In fact, the daemonset container depends on the ConfigMap.

>Also, we should set Controller to false in the daemonsetRef.

Based on the API docs I'm not sure I totally understand what the controller flag even means here... help?

pkg/operator/controller/controller_dns_daemonset.go
ironcladlou: Yes, cluster-scoped owners are okay: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#ownerreference-v1-meta

Makefile
ironcladlou: Fixed

pkg/operator/controller/controller_dns_configmap.go
ironcladlou: Fixed the overwriting but didn't change the controller bit.

Miciah: > Based on the API docs I'm not sure I totally understand what the controller flag even means here... help?

See the godoc for `EnqueueRequestForOwner`: "IsController if set will only look at the first OwnerReference with Controller: true."  I infer from that that leaving `Controller` to true on the daemonset owner reference could cause the watch to enqueue the daemonset instead of the dns.

Makefile
```diff
@@ -17,9 +17,19 @@ build:
...
-generate:
+generate: crd
 	hack/update-generated-bindata.sh
 
+# Generate CRD from vendored API spec.
+.PHONY: crd
+crd:
+	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --apis-dir vendor/github.com/openshift/api
+
+# Do not write the ClusterIngress CRD, only compare and return (code 1 if dirty).
```
Miciah: Missed this "ClusterIngress".

ironcladlou: D'oh, thanks. Fixed.

pkg/operator/controller/controller_dns_configmap.go
ironcladlou: Interesting, thanks

pkg/operator/operator.go
```diff
@@ -68,57 +65,17 @@ func New(config operatorconfig.Config) (*Operator, error) {
...
-				} else {
-					return []reconcile.Request{}
-				}
-			}),
-		})
-	}
+	operatorController.Watch(&source.Kind{Type: &operatorv1.DNS{}}, &handler.EnqueueRequestForObject{})
+	operatorController.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, &handler.EnqueueRequestForOwner{})
+	operatorController.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{})
+	operatorController.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForOwner{})
```
Miciah: This has been nagging me for a while: Shouldn't these watches be established in `operatorcontroller.New`?

ironcladlou: Fixed

pkg/operator/controller/controller.go
```diff
@@ -64,6 +65,15 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &operatorv1.DNS{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return nil, err
 	}
+	if err := c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, &handler.EnqueueRequestForOwner{}); err != nil {
+		return nil, err
+	}
+	if err := c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, &handler.EnqueueRequestForOwner{}); err != nil {
```
Miciah: You establish two watches on `DaemonSet`.

ironcladlou: Fixed and squashed

Miciah:
With the dns owning the daemonset, do we need `ensureDNSDeleted`?

We could almost delete the finalizer, but for `ensureOpenshiftExternalNameServiceDeleted`.

ironcladlou:
>With the dns owning the daemonset, do we need ensureDNSDeleted?
>We could almost delete the finalizer, but for ensureOpenshiftExternalNameServiceDeleted.

Yeah, there's actually probably a lot we can do to refactor but I was just trying to do the minimum to upgrade libs for now

Miciah:
> Yeah, there's actually probably a lot we can do to refactor but I was just trying to do the minimum to upgrade libs for now

All right.  We are leaving cruft in the code and performing a spurious `Delete` in `ensureDNSDaemonSetDeleted`, but we do not log if we get a not-found error, so it is mostly harmless.

/lgtm

ironcladlou:
Sorry for the CI spam — there was actually a bug in the refactored code which I just got around to looking into and fixing this morning.

Miciah:
The godoc for `EnqueueRequestForOwner` doesn't say that `OwnerType` is required; documentation defect?

PR #88, opened 2019-04-04
Bug 1690342: Initial DNS status conditions implementation

danehans:
Previously, status conditions for the `dns.operator.openshift.io/v1` resource were not implemented.

This commit does the following:

- Implements `Degraded`, `Progressing` and `Available` status conditions for `dns.operator.openshift.io/v1`.
- Refactors cluster-dns-operator `Degraded` status condition to be calculated from the the availability of `ClusterIP`.
- Refactors cluster-dns-operator `Progressing` and `Available` status conditions to be calculated from the `dns.operator.openshift.io/v1` `Available` condition.
- Replaces deployment logic from cluster-dns-operator unit tests with `dns.operator.openshift.io/v1` `Available` conditions.
- Adds `dns.operator.openshift.io/v1` status condition unit tests.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -382,31 +382,12 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
 			errs = append(errs, fmt.Errorf("failed to integrate metrics with openshift-monitoring for dns %s: %v", dns.Name, err))
 		}
 
-		if err := r.syncDNSStatus(dns, clusterIP, clusterDomain); err != nil {
+		if err := r.syncDNSStatus(dns, clusterIP, clusterDomain, daemonset); err != nil {
 			errs = append(errs, fmt.Errorf("failed to sync status of dns %s/%s: %v", daemonset.Namespace, daemonset.Name, err))
 		}
 	}
 
 	return utilerrors.NewAggregate(errs)
-}
-
-// syncDNSStatus updates the status for a given dns.
-func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain string) error {
-	current := &operatorv1.DNS{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: dns.Name}, current); err != nil {
-		return fmt.Errorf("failed to get dns %s: %v", dns.Name, err)
-	}
-	if current.Status.ClusterIP == clusterIP &&
-		current.Status.ClusterDomain == clusterDomain {
-		return nil
-	}
-	current.Status.ClusterIP = clusterIP
-	current.Status.ClusterDomain = clusterDomain
-
-	if err := r.client.Status().Update(context.TODO(), current); err != nil {
-		return fmt.Errorf("failed to update status for dns %s: %v", current.Name, err)
-	}
-	return nil
 }
 
 // getClusterIPFromNetworkConfig will return 10th IP from the service CIDR range
--- a/pkg/operator/controller/dns_status.go
+++ b/pkg/operator/controller/dns_status.go
@@ -1,0 +1,170 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// syncDNSStatus computes the current status of dns and
+// updates status upon any changes since last sync.
+func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain string, ds *appsv1.DaemonSet) error {
+	updated := dns.DeepCopy()
+	updated.Status.ClusterIP = clusterIP
+	updated.Status.ClusterDomain = clusterDomain
+	updated.Status.Conditions = r.computeDNSStatusConditions(dns.Status.Conditions, clusterIP, ds)
+	if !dnsStatusesEqual(updated.Status, dns.Status) {
+		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+			return fmt.Errorf("failed to update dns status: %v", err)
+		}
+	}
+
+	return nil
+}
+
+// computeDNSStatusConditions computes dns status conditions based on
+// the status of ds and clusterIP.
+func (r *reconciler) computeDNSStatusConditions(oldConditions []operatorv1.OperatorCondition, clusterIP string,
+	ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
+	var oldDegradedCondition, oldProgressingCondition, oldAvailableCondition *operatorv1.OperatorCondition
+	for i := range oldConditions {
+		switch oldConditions[i].Type {
+		case operatorv1.OperatorStatusTypeDegraded:
+			oldDegradedCondition = &oldConditions[i]
+		case operatorv1.OperatorStatusTypeProgressing:
+			oldProgressingCondition = &oldConditions[i]
+		case operatorv1.OperatorStatusTypeAvailable:
+			oldAvailableCondition = &oldConditions[i]
+		}
+	}
+
+	conditions := []operatorv1.OperatorCondition{
+		computeDNSDegradedCondition(oldDegradedCondition, clusterIP, ds),
+		computeDNSProgressingCondition(oldProgressingCondition, ds),
+		computeDNSAvailableCondition(oldAvailableCondition, clusterIP, ds),
+	}
+
+	return conditions
+}
+
+// computeDNSDegradedCondition computes the dns Degraded status condition
+// based on the status of clusterIP and ds.
+func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clusterIP string,
+	ds *appsv1.DaemonSet) operatorv1.OperatorCondition {
+	degradedCondition := &operatorv1.OperatorCondition{
+		Type: operatorv1.OperatorStatusTypeDegraded,
+	}
+	switch {
+	case len(clusterIP) == 0 && ds.Status.NumberAvailable == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "NoClusterIPAndDaemonSet"
+		degradedCondition.Message = "No ClusterIP assigned to DNS Service and no DaemonSet pods running"
+	case len(clusterIP) == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "NoClusterIP"
+		degradedCondition.Message = "No ClusterIP assigned to DNS Service"
+	case ds.Status.NumberAvailable == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "DaemonSetUnavailable"
+		degradedCondition.Message = "DaemonSet pod not running on any Nodes"
+	case ds.Status.NumberAvailable != ds.Status.DesiredNumberScheduled:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "DaemonSetDegraded"
+		degradedCondition.Message = "Not all Nodes running DaemonSet pod"
+	default:
+		degradedCondition.Status = operatorv1.ConditionFalse
+		degradedCondition.Reason = "AsExpected"
+		degradedCondition.Message = "ClusterIP assigned to DNS Service and minimum DaemonSet pods running"
+	}
+
+	return setDNSLastTransitionTime(degradedCondition, oldCondition)
+}
+
+// computeDNSProgressingCondition computes the dns Progressing status condition
+// based on the status of ds.
+func computeDNSProgressingCondition(oldCondition *operatorv1.OperatorCondition, ds *appsv1.DaemonSet) operatorv1.OperatorCondition {
+	progressingCondition := &operatorv1.OperatorCondition{
+		Type: operatorv1.OperatorStatusTypeProgressing,
+	}
+	if ds.Status.NumberAvailable == ds.Status.DesiredNumberScheduled {
+		progressingCondition.Status = operatorv1.ConditionFalse
+		progressingCondition.Reason = "AsExpected"
+		progressingCondition.Message = "All expected Nodes running DaemonSet pod"
+	} else {
+		progressingCondition.Status = operatorv1.ConditionTrue
+		progressingCondition.Reason = "Reconciling"
+		progressingCondition.Message = fmt.Sprintf("%d Nodes running a DaemonSet pod, want %d",
+			ds.Status.NumberAvailable, ds.Status.DesiredNumberScheduled)
+	}
+
+	return setDNSLastTransitionTime(progressingCondition, oldCondition)
+}
+
+// computeDNSAvailableCondition computes the dns Available status condition
+// based on the status of clusterIP and ds.
+func computeDNSAvailableCondition(oldCondition *operatorv1.OperatorCondition, clusterIP string, ds *appsv1.DaemonSet) operatorv1.OperatorCondition {
+	availableCondition := &operatorv1.OperatorCondition{
+		Type: operatorv1.OperatorStatusTypeAvailable,
+	}
+	switch {
+	case len(clusterIP) == 0 && ds.Status.NumberAvailable == 0:
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "NoClusterIPAndDaemonSet"
+		availableCondition.Message = "No ClusterIP assigned to DNS Service and no running DaemonSet pods"
+	case len(clusterIP) == 0:
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "NoDNSService"
+		availableCondition.Message = "No ClusterIP assigned to DNS Service"
+	case ds.Status.NumberAvailable == 0:
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "DaemonSetUnavailable"
+		availableCondition.Message = "DaemonSet pod not running on any Nodes"
+	default:
+		availableCondition.Status = operatorv1.ConditionTrue
+		availableCondition.Reason = "AsExpected"
+		availableCondition.Message = "Minimum number of Nodes running DaemonSet pod"
+	}
+
+	return setDNSLastTransitionTime(availableCondition, oldCondition)
+}
+
+// setDNSLastTransitionTime sets LastTransitionTime for the given condition.
+// If the condition has changed, it will assign a new timestamp otherwise keeps the old timestamp.
+func setDNSLastTransitionTime(condition, oldCondition *operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+	if oldCondition != nil && condition.Status == oldCondition.Status &&
+		condition.Reason == oldCondition.Reason && condition.Message == oldCondition.Message {
+		condition.LastTransitionTime = oldCondition.LastTransitionTime
+	} else {
+		condition.LastTransitionTime = metav1.Now()
+	}
+
+	return *condition
+}
+
+// dnsStatusesEqual compares two DNSStatus values.  Returns true
+// if the provided values should be considered equal for the purpose of determining
+// whether an update is necessary, false otherwise.
+func dnsStatusesEqual(a, b operatorv1.DNSStatus) bool {
+	conditionCmpOpts := []cmp.Option{
+		cmpopts.EquateEmpty(),
+		cmpopts.SortSlices(func(a, b operatorv1.OperatorCondition) bool { return a.Type < b.Type }),
+	}
+	if !cmp.Equal(a.Conditions, b.Conditions, conditionCmpOpts...) {
+		return false
+	}
+	if a.ClusterIP != b.ClusterIP {
+		return false
+	}
+	if a.ClusterDomain != b.ClusterDomain {
+		return false
+	}
+
+	return true
+}
--- a/pkg/operator/controller/dns_status_test.go
+++ b/pkg/operator/controller/dns_status_test.go
@@ -1,0 +1,304 @@
+package controller
+
+import (
+	"fmt"
+	"testing"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var r reconciler
+
+func TestDNSStatusConditions(t *testing.T) {
+	type testInputs struct {
+		haveClusterIP bool
+		avail, desire int32
+	}
+	type testOutputs struct {
+		degraded, progressing, available bool
+	}
+	testCases := []struct {
+		description string
+		inputs      testInputs
+		outputs     testOutputs
+	}{
+		{
+			description: "no cluster ip, 0/0 pods available",
+			inputs:      testInputs{false, 0, 0},
+			outputs:     testOutputs{true, false, false},
+		},
+		{
+			description: "no cluster ip, 0/2 pods available",
+			inputs:      testInputs{false, 0, 2},
+			outputs:     testOutputs{true, true, false},
+		},
+		{
+			description: "no cluster ip, 1/2 pods available",
+			inputs:      testInputs{false, 1, 2},
+			outputs:     testOutputs{true, true, false},
+		},
+		{
+			description: "no cluster ip, 2/2 pods available",
+			inputs:      testInputs{false, 2, 2},
+			outputs:     testOutputs{true, false, false},
+		},
+		{
+			description: "daemonset pod available on 0/0 nodes",
+			inputs:      testInputs{true, 0, 0},
+			outputs:     testOutputs{true, false, false},
+		},
+		{
+			description: "daemonset pod available on 0/2 nodes",
+			inputs:      testInputs{true, 0, 2},
+			outputs:     testOutputs{true, true, false},
+		},
+		{
+			description: "daemonset pod available on 1/2 nodes",
+			inputs:      testInputs{true, 1, 2},
+			outputs:     testOutputs{true, true, true},
+		},
+		{
+			description: "daemonset pod available on 2/2 nodes",
+			inputs:      testInputs{true, 2, 2},
+			outputs:     testOutputs{false, false, true},
+		},
+	}
+
+	for i, tc := range testCases {
+		var (
+			clusterIP string
+
+			degraded, progressing, available operatorv1.ConditionStatus
+		)
+		if tc.inputs.haveClusterIP {
+			clusterIP = "1.2.3.4"
+		}
+		ds := &appsv1.DaemonSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("dns-%d", i+1),
+			},
+			Status: appsv1.DaemonSetStatus{
+				DesiredNumberScheduled: tc.inputs.desire,
+				NumberAvailable:        tc.inputs.avail,
+			},
+		}
+		if tc.outputs.degraded {
+			degraded = operatorv1.ConditionTrue
+		} else {
+			degraded = operatorv1.ConditionFalse
+		}
+		if tc.outputs.progressing {
+			progressing = operatorv1.ConditionTrue
+		} else {
+			progressing = operatorv1.ConditionFalse
+		}
+		if tc.outputs.available {
+			available = operatorv1.ConditionTrue
+		} else {
+			available = operatorv1.ConditionFalse
+		}
+		expected := []operatorv1.OperatorCondition{
+			{
+				Type:   operatorv1.OperatorStatusTypeDegraded,
+				Status: degraded,
+			},
+			{
+				Type:   operatorv1.OperatorStatusTypeProgressing,
+				Status: progressing,
+			},
+			{
+				Type:   operatorv1.OperatorStatusTypeAvailable,
+				Status: available,
+			},
+		}
+		actual := r.computeDNSStatusConditions([]operatorv1.OperatorCondition{}, clusterIP, ds)
+		gotExpected := true
+		if len(actual) != len(expected) {
+			gotExpected = false
+		}
+		for _, conditionA := range actual {
+			foundMatchingCondition := false
+
+			for _, conditionB := range expected {
+				if conditionA.Type == conditionB.Type &&
+					conditionA.Status == conditionB.Status {
+					foundMatchingCondition = true
+					break
+				}
+			}
+
+			if !foundMatchingCondition {
+				gotExpected = false
+			}
+		}
+		if !gotExpected {
+			t.Fatalf("%q: expected %#v, got %#v", tc.description,
+				expected, actual)
+		}
+	}
+}
+
+func TestDNSStatusesEqual(t *testing.T) {
+	testCases := []struct {
+		description string
+		expected    bool
+		a, b        operatorv1.DNSStatus
+	}{
+		{
+			description: "nil and non-nil slices are equal",
+			expected:    true,
+			a: operatorv1.DNSStatus{
+				Conditions: []operatorv1.OperatorCondition{},
+			},
+		},
+		{
+			description: "empty slices should be equal",
+			expected:    true,
+			a: operatorv1.DNSStatus{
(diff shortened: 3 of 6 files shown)


pkg/operator/controller/status_test.go
```diff
@@ -7,15 +7,14 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TestComputeStatusConditions(t *testing.T) {
+func TestComputeOperatorStatusConditions(t *testing.T) {
 	type testInputs struct {
-		haveNamespace                           bool
-		numWanted, numAvailable, numUnavailable int
+		haveNamespace           bool
+		numWanted, numAvailable int
```
Miciah: It feels like we have lost some test coverage here, because whereas before we were counting things that had a positive number in some particular field, now we are counting things that have a slice of things where one of those things has a specific value in one particular field and another specific value in another particular field.  That is to say, there are many more ways the logic could go wrong.  What do you think about adding inputs for the number of dnses without an `Available` condition, number with `Available` set to `Unknown`, and number with `Available` set to `False`, as well as a test case for each of these and for a combination of several?

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,258 @@
...
+func TestDNSStatusConditions(t *testing.T) {
+	testCases := []struct {
+		description   string
+		avail, desire int32
+		condType      string
+		condStatus    operatorv1.ConditionStatus
+	}{
+		{"daemonset pod available on 0/2 nodes", 0, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionFalse},
+		{"daemonset pod available on 1/2 nodes", 1, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionTrue},
+		{"daemonset pod available on 2/2 nodes", 2, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionTrue},
```
Miciah: We should decide what 0/0 means and add a test case.

pkg/operator/controller/controller.go
```diff
@@ -316,34 +316,14 @@ func (r *reconciler) ensureClusterDNS(dns *operatorv1.DNS) error {
 		if _, err := r.ensureDNSService(dns, clusterIP, daemonsetRef); err != nil {
 			errs = append(errs, fmt.Errorf("failed to create service for cluster dns %s: %v", dns.Name, err))
 		}
-
-		if err := r.syncClusterDNSStatus(dns, clusterIP, clusterDomain); err != nil {
-			errs = append(errs, fmt.Errorf("failed to sync status of cluster dns %s/%s: %v", daemonset.Namespace, daemonset.Name, err))
+		if err := r.syncDNSStatus(daemonset, dns); err != nil {
+			errs = append(errs, fmt.Errorf("failed to sync dns status: %v", err))
```
Miciah: `syncDNSStatus` should set `.status.clusterIP` and `.status.clusterDomain`.

pkg/operator/controller/status_test.go
danehans: Thank you for the review. I have addressed all your comments in my latest commit, except this one. My understanding is `numUnavailable` was removed because the `Progressing` and `Available` status conditions are now based on the `DNS` status condition instead of comparing the number of available and unavailable `DaemonSets`.

danehans:
Updated PR to also fix [bug 1695210](https://bugzilla.redhat.com/show_bug.cgi?id=1695210). __Note:__ This bug references cluster-ingress-operator, but cluster-dns-operator is affected by the same bug.

danehans:
Holding until https://github.com/openshift/cluster-ingress-operator/pull/201 merges since dns operator requires the same bug fix.

danehans:
Holding until https://github.com/openshift/cluster-ingress-operator/pull/201 merges since dns operator requires the same bug fix.

knobunc:
@danehans openshift/cluster-ingress-operator#201 has merged... should the hold be lifted?

danehans:
@knobunc I would like to see https://github.com/openshift/cluster-dns-operator/pull/92 and https://github.com/openshift/cluster-ingress-operator/pull/200 merge. Both are close.

danehans:
```
Apr 25 03:56:25.007 E clusteroperator/machine-config changed Degraded to True: error pool master is not ready, retrying. Status: (pool degraded: false total: 3, ready 2, updated: 2, unavailable: 1): Unable to apply 0.0.1-2019-04-25-025503: error pool master is not ready, retrying. Status: (pool degraded: false total: 3, ready 2, updated: 2, unavailable: 1)
Apr 25 04:01:19.105 E clusterversion/version changed Failing to True: ClusterOperatorDegraded: Cluster operator kube-apiserver is reporting a failure
```
/test e2e-aws-upgrade

danehans:
@Miciah @pravisankar I pulled the hold and this PR is ready for your review.

ironcladlou:
I think this is good. @Miciah PTAL and tag when ready

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,146 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// syncDNSStatus computes the current status of dns and
+// updates status upon any changes since last sync.
+func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain string, ds *appsv1.DaemonSet) error {
+	updated := dns.DeepCopy()
+	updated.Status.ClusterIP = clusterIP
+	updated.Status.ClusterDomain = clusterDomain
+	updated.Status.Conditions = computeDNSStatusConditions(updated.Status.Conditions, clusterIP, ds)
```
pravisankar: We do want to pass the old status conditions, `computeDNSStatusConditions(dns.Status.Conditions,..`  might be clear?

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+// computeDNSStatusConditions computes dns status conditions
+// based on the status of ds.
+func computeDNSStatusConditions(conditions []operatorv1.OperatorCondition, clusterIP string, ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
+	conditions = computeDNSDegradedCondition(conditions, clusterIP)
+	conditions = computeDNSProgressingCondition(conditions, ds)
+	conditions = computeDNSAvailableCondition(conditions, ds)
+
```
pravisankar: `computeDNSStatusConditions` does not depend on old status conditions so we don't have to pass this as an argument?
May be we could do:
```
func computeDNSStatusConditions(ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
        conditions := []operatorv1.OperatorCondition{}
	conditions = append(conditions, computeDNSDegradedCondition(clusterIP))
	conditions = append(conditions, computeDNSProgressingCondition(ds))
	conditions = append(conditions, computeDNSAvailableCondition(ds))
        return conditions
}
```
Each of the `computeDNSConditions` methods could return a condition and we no longer need the `setDNSStatusCondition` method.

Miciah: `computeDNSFooCondition` needs the old condition as input so that it can return the old condition with the old last transition time if the `Foo` condition did not change.

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,146 @@
...
+		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+			return fmt.Errorf("failed to update dns status: %v", err)
+		}
+	}
+
+	return nil
+}
+
+// computeDNSStatusConditions computes dns status conditions
+// based on the status of ds.
```
Miciah: Technically, that should be "based on the status of ds and clusterIP"... but maybe I am being excessively fussy.

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,146 @@
...
+// whether an update is necessary, false otherwise.
+func dnsStatusesEqual(a, b operatorv1.DNSStatus) bool {
+	conditionCmpOpts := []cmp.Option{
+		cmpopts.IgnoreFields(operatorv1.OperatorCondition{}, "LastTransitionTime"),
+		cmpopts.EquateEmpty(),
+		cmpopts.SortSlices(func(a, b operatorv1.OperatorCondition) bool { return a.Type < b.Type }),
+	}
+	if !cmp.Equal(a.Conditions, b.Conditions, conditionCmpOpts...) || a.ClusterIP != b.ClusterIP || a.ClusterDomain != b.ClusterDomain {
+		return false
+	}
```
Miciah: In `statusesEqual`, we do the following:
```go
	if foo is different {
		return false
	}
	if bar is different {
		return false
	}
	// ...
	return true
```
I find that approach easier to read.  I don't object if you prefer a single `if foo is different || if bar different || ... {` statement though.  However, if you do go in that direction, you could simplify further by doing away with the `if` and simply returning a Boolean expression:
```go
	return foo is the same && bar is the same && ...
```
Concretely:
```go
	return cmp.Equal(a.Conditions, b.Conditions, conditionCmpOpts...) && a.ClusterIP == b.ClusterIP && a.ClusterDomain == b.ClusterDomain
```

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,379 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestDNSStatusConditions(t *testing.T) {
+	testCases := []struct {
+		description                                string
+		clusterIP                                  string
+		avail, desire                              int32
+		degradeType, progressType, availType       string
```
Miciah: Why are `degradeType`, `progressType`, and `availType` inputs? They are all the same for all test cases, and each one has only one reasonable value.

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,379 @@
...
+			description:    "no cluster ip, 1/2 pods available",
+			clusterIP:      "",
+			avail:          1,
+			desire:         2,
+			degradeType:    operatorv1.OperatorStatusTypeDegraded,
+			progressType:   operatorv1.OperatorStatusTypeProgressing,
+			availType:      operatorv1.OperatorStatusTypeAvailable,
+			degradeStatus:  operatorv1.ConditionTrue,
+			progressStatus: operatorv1.ConditionTrue,
+			availStatus:    operatorv1.ConditionTrue,
```
Miciah: This looks odd—is DNS really available if it has no IP address?

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,379 @@
...
+			description:    "daemonset pod available on 1/2 nodes",
+			clusterIP:      "1.2.3.4",
+			avail:          1,
+			desire:         2,
+			degradeType:    operatorv1.OperatorStatusTypeDegraded,
+			progressType:   operatorv1.OperatorStatusTypeProgressing,
+			availType:      operatorv1.OperatorStatusTypeAvailable,
+			degradeStatus:  operatorv1.ConditionFalse,
+			progressStatus: operatorv1.ConditionTrue,
+			availStatus:    operatorv1.ConditionTrue,
```
Miciah: Would it make sense to say that the service is degraded if it is available on some but not all nodes? It is still available, but for pods on a node where the daemonset is not running, DNS requests will be routed via the service proxy to another node, so performance and possibly reliability may be slightly degraded.

pkg/operator/controller/status.go
```diff
@@ -128,58 +129,77 @@ func computeStatusConditions(conditions []configv1.ClusterOperatorStatusConditio
 		degradedCondition.Message = "DNS namespace does not exist"
 	} else {
 		degradedCondition.Status = configv1.ConditionFalse
+		degradedCondition.Reason = "NamespaceExists"
```
Miciah: It looks like most operators use "AsExpected" for the `Degraded` condition's `Reason` field when its `Status` field is false:
```console
% oc get clusteroperators -o json | jq -r '.items|.[].status.conditions|.[]|select(.type=="Degraded")|[.reason][0]' | sort | uniq -c | sort -n
      1 NoCredentialsFailing
      5 null
     10 AsExpected
```
I'm not one to advocate blind conformity in all things, but I submit the above for your consideration.

pkg/operator/controller/status.go
```diff
@@ -128,58 +129,77 @@ func computeStatusConditions(conditions []configv1.ClusterOperatorStatusConditio
...
+			if c.Type == operatorv1.DNSAvailable &&
+				c.Status == operatorv1.ConditionTrue {
+				dnsesAvailable++
+				break
+			}
+		}
+	}
+	if numDNSes == dnsesAvailable {
 		progressingCondition.Status = configv1.ConditionFalse
+		progressingCondition.Reason = "AllDaemonSetPodsRunning"
```
Miciah: Since I provided data on how other operators set the `Degraded` condition, I'll provide similar data on the `Progressing` condition, for your consideration:
```console
% oc get clusteroperators -o json | jq -r '.items|.[].status.conditions|.[]|select(.type=="Progressing")|[.reason][0]' | sort |
uniq -c | sort -n
      1 Ready
      1 ReconcilingComplete
     10 AsExpected
     13 null
```

pkg/operator/controller/status.go
```diff
@@ -128,58 +129,77 @@ func computeStatusConditions(conditions []configv1.ClusterOperatorStatusConditio
...
+				c.Status == operatorv1.ConditionTrue {
+				dnsesAvailable++
+				break
+			}
 		}
 	}
-	if len(unavailable) == 0 {
+	if dnsesAvailable > 0 {
 		availableCondition.Status = configv1.ConditionTrue
+		availableCondition.Reason = "DNSAvailable"
```
Miciah:
```console
% oc get clusteroperators -o json | jq -r '.items|.[].status.conditions|.[]|select(.type=="Available")|[.reason][0]' | sort | uniq -c | sort -n
      1 MinimumAvailability
     10 AsExpected
     14 null
```

pkg/operator/controller/dns_status.go
Miciah: On further thought, we could do something very close to Ravi's suggestion:
```go
func computeDNSStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
	conditions = setOperatorStatusCondition(conditions, computeDNSDegradedCondition(clusterIP))
	conditions = setOperatorStatusCondition(conditions, computeDNSProgressingCondition(ds))
	conditions = setOperatorStatusCondition(conditions, computeDNSAvailableCondition(ds))
        return conditions
}
```
I kind of like that, and it should correctly preserve or as appropriate update `LastTransitionTime`.

pkg/operator/controller/dns_status.go
danehans: `setDNSStatusCondition` creates a slice of conditions. The func has logic to append conditions to the new slice and return it. I thought we should be operating on a copy (i.e. `updated`) of the object under these circumstances?

pkg/operator/controller/dns_status.go
danehans: I agree with your recommendation and find the code easier to read that way. I'll update.

pkg/operator/controller/dns_status_test.go
danehans: Good point. The previous "available" condition logic only considered having 1 running pod in the daemonset for the condition to be true. Let me update that logic to consider the clusterIP too.

danehans: I will update the available condition to consider the ClusterIP status.

pkg/operator/controller/dns_status_test.go
danehans: Yes, I will update the condition to consider the ClusterIP status.

pkg/operator/controller/status.go
danehans: I'll use `AsExpected`.

pkg/operator/controller/status.go
danehans: I'll use `AsExpected`.

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,173 @@
...
+	switch {
+	case len(clusterIP) == 0 && ds.Status.NumberAvailable == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "NoClusterIPAndDaemonSet"
+		degradedCondition.Message = "No ClusterIP for DNS Service and no DaemonSet pods running"
+	case len(clusterIP) == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "NoClusterIP"
+		degradedCondition.Message = "No Cluster IP for DNS Service available"
```
Miciah: Should be "ClusterIP" for consistency (or alternatively we could change the "ClusterIP" to "Cluster IP" above and in `computeDNSAvailableCondition`).  Also, maybe "available"→"assigned".

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,173 @@
...
+	case ds.Status.NumberAvailable == 0:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "DaemonSetUnavailable"
+		degradedCondition.Message = "DaemonSet pod not running on any nodes"
+	case ds.Status.NumberAvailable != ds.Status.DesiredNumberScheduled:
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = "DaemonSetDegraded"
+		degradedCondition.Message = "All nodes not running DaemonSet pod"
```
Miciah: This could be misinterpreted as saying that no nodes are running the DaemonSet pod.  "Not all nodes running" or "Some nodes not running" would be clearer.

pkg/operator/controller/status.go
```diff
@@ -95,29 +90,35 @@ func (r *reconciler) syncOperatorStatus() error {
...
-	return ns, dnsList.Items, daemonsetList.Items, nil
+	return ns, dnsList.Items, nil
+}
+
+// computeOperatorStatusConditions computes the operator's current state.
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace,
+	dnses []operatorv1.DNS) []configv1.ClusterOperatorStatusCondition {
+	conditions = computeOperatorDegradedCondition(conditions, ns)
+	conditions = computeOperatorProgressingCondition(conditions, dnses)
+	conditions = computeOperatorAvailableCondition(conditions, dnses)
```
Miciah: Could we use the same `conditions = setStatusCondition(conditions, computeFooCondition(...))` pattern here that you implemented in `computeDNSStatusConditions`?

danehans: Sure. I'll push an update to the PR in a few minutes. Thanks for all the great feedback!

Miciah:
/lgtm
/hold
@pravisankar, do you want to give this another look before we merge? If you're fine with the changes, feel free to cancel the hold.

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,354 @@
...
+		degradeStatus, progressStatus, availStatus operatorv1.ConditionStatus
+	}{
+		{
+			description:    "no cluster ip, 0/0 pods available",
+			clusterIP:      "",
+			avail:          0,
+			desire:         0,
+			degradeStatus:  operatorv1.ConditionTrue,
+			progressStatus: operatorv1.ConditionFalse,
+			availStatus:    operatorv1.ConditionFalse,
```
pravisankar: As @ironcladlou suggested on the ingress pr https://github.com/openshift/cluster-ingress-operator/pull/200 , this can be simplified:
```
type conditions struct {
	degraded, progressing, available bool
}
And then in the test case, we could do
...
expectedConditions: conditions{true, false, false}
...
```

pkg/operator/controller/status_test.go
```diff
@@ -25,48 +24,92 @@ func TestComputeStatusConditions(t *testing.T) {
...
+			testInputs{
+				haveNamespace: false,
+				numWanted:     1,
+				numAvailable:  0,
+			},
+			testOutputs{
+				degraded:    true,
+				progressing: true,
+				available:   false,
+			},
```
pravisankar: This can be converted to compact format.

pkg/operator/controller/dns_status.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return nil
+}
+
+// computeDNSStatusConditions computes dns status conditions
+// based on the status of ds and clusterIP.
+func computeDNSStatusConditions(conditions []operatorv1.OperatorCondition, clusterIP string, ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
+	conditions = setDNSStatusCondition(conditions, computeDNSDegradedCondition(clusterIP, ds))
+	conditions = setDNSStatusCondition(conditions, computeDNSProgressingCondition(ds))
+	conditions = setDNSStatusCondition(conditions, computeDNSAvailableCondition(clusterIP, ds))
```
pravisankar: We could get rid of `setDNSStatusCondition`. Instead of looping all conditions for every `computeDNSXYZCondition` to check for changes, we could compute actual condition once and can patch the LastTransitionTime. Also did not like the fact that LastTransitionTime for the condition is set outside of `computeDNSXYZCondition` method.
WDYT of  https://github.com/openshift/cluster-ingress-operator/pull/200/files#diff-31d813e7a1281abc26f3e03bc7686530R163 ?

pkg/operator/controller/status.go
```diff
@@ -95,29 +90,35 @@ func (r *reconciler) syncOperatorStatus() error {
...
-	return ns, dnsList.Items, daemonsetList.Items, nil
+	return ns, dnsList.Items, nil
+}
+
+// computeOperatorStatusConditions computes the operator's current state.
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace,
+	dnses []operatorv1.DNS) []configv1.ClusterOperatorStatusCondition {
+	conditions = setOperatorStatusCondition(conditions, computeOperatorDegradedCondition(ns))
+	conditions = setOperatorStatusCondition(conditions, computeOperatorProgressingCondition(dnses))
+	conditions = setOperatorStatusCondition(conditions, computeOperatorAvailableCondition(dnses))
```
pravisankar: We could get rid of setOperatorStatusCondition and can set LastTransitionTime as part of computeOperatorXYZCondition.

pkg/operator/controller/dns_status_test.go
```diff
@@ -0,0 +1,354 @@
...
+		}
+	}
+}
+
+func TestSetDNSStatusCondition(t *testing.T) {
+	testCases := []struct {
+		description   string
+		oldConditions []operatorv1.OperatorCondition
+		newCondition  *operatorv1.OperatorCondition
+		expected      []operatorv1.OperatorCondition
```
pravisankar: Instead of using operatorv1.OperatorCondition, we could use conditions{available: true}, conditions{degraded: false, progressing: true} and build the actual conditions based on this concise input before calling the test method.

danehans:
@pravisankar thank you for the review. I have updated the PR based on your comments.

pkg/operator/controller/status.go
```diff
@@ -95,123 +99,184 @@ func (r *reconciler) syncOperatorStatus() error {
...
+				break
+			}
+		}
+		if !available {
+			return false
+		}
+	}
 
-	newConditions := []configv1.ClusterOperatorStatusCondition{}
+	return true
```
Miciah: We decided that a DaemonSet is available if 0/0 replicas are available (https://github.com/openshift/cluster-dns-operator/pull/88#discussion_r272673377, https://github.com/openshift/cluster-dns-operator/commit/5ac65e290667f001e1029fb0f69dc359875451c6#diff-c7469ee2a512d7eff942a36312cb4c50R20), and I see we are extending that reasoning by saying that all DNSes are available if 0/0 DNSes are available.

However, consider that the operator will automatically recreate the `default` DNS if it does not exist.  This consideration leads me to think that having 0/0 DNSes implies that `Progressing` is true, and possibly that `Available` is false.  We have discussed the analogous case of 0/0 ingress controllers in relation to the ingress operator: https://github.com/openshift/cluster-ingress-operator/pull/198#issuecomment-481058325.

pkg/operator/controller/status.go
```diff
@@ -95,123 +99,184 @@ func (r *reconciler) syncOperatorStatus() error {
...
-	found := false
-	for _, c := range oldConditions {
-		if condition.Type == c.Type {
-			if condition.Status == c.Status &&
-				condition.Reason == c.Reason &&
-				condition.Message == c.Message {
-				return oldConditions
+// checkAnyDNSesAvailable checks if all DNS resources are available.
+func checkAnyDNSesAvailable(dnses []operatorv1.DNS) bool {
+	var dnsesAvailable int
```
Miciah: Note for follow-up: This can be a Boolean value.  (Alternatively, if we decide not to combine `checkAnyDNSesAvailable` and `checkAllDNSesAvailable` as I suggested in another comment, we can do away with the variable and just `dnsesAvailable++`→`return true` below.)

pkg/operator/controller/status.go
```diff
@@ -95,123 +99,184 @@ func (r *reconciler) syncOperatorStatus() error {
...
+	if anyDNSesAvailable {
 		availableCondition.Status = configv1.ConditionTrue
+		availableCondition.Reason = "AsExpected"
+		availableCondition.Message = "At least 1 dns resource is available"
 	} else {
 		availableCondition.Status = configv1.ConditionFalse
-		availableCondition.Reason = "DaemonSetNotAvailable"
-		availableCondition.Message = strings.Join(unavailable, "\n")
+		availableCondition.Reason = "DNSUnavailable"
+		availableCondition.Message = fmt.Sprintf("No dns resources available")
```
Miciah: Note for follow-up: Use consistent capitalization of "dns".

pkg/operator/controller/status.go
ironcladlou: Given our invariant that `default` must always exist, agree that 0 `dns` resources means progressing=true.

IMO available should always be false if the actual replicas of `default` is 0 regardless of what the "desired". I don't understand how, from our operator's perspective, DNS could be considered available if no CoreDNS pods are running.

pkg/operator/controller/status.go
```diff
@@ -44,13 +45,16 @@ func (r *reconciler) syncOperatorStatus() error {
...
 	}
 
-	ns, dnses, daemonsets, err := r.getOperatorState()
+	ns, dnses, err := r.getOperatorState()
 	if err != nil {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
 
+	allDNSesAvailable := checkAllDNSesAvailable(dnses)
+	anyDNSesAvailable := checkAnyDNSesAvailable(dnses)
```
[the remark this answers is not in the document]
danehans: I converted both funcs into a `numDNSesAvailable` func and compare the number of dnses available to the total number of dnses in the status conditions calculation.

pkg/operator/controller/status.go
```diff
@@ -95,123 +99,184 @@ func (r *reconciler) syncOperatorStatus() error {
...
 	}
 
-	daemonsetList := &appsv1.DaemonSetList{}
-	if err := r.client.List(context.TODO(), daemonsetList, client.InNamespace(ns.Name)); err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to list daemonsets: %v", err)
+	return ns, dnsList.Items, nil
+}
+
+// computeOperatorStatusConditions computes the operator's current state.
+func (r *reconciler) computeOperatorStatusConditions(oldConditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace,
```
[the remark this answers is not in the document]
danehans: I have removed the receiver. It was added to include status versions as part of the `Progressing` condition calculation. The version calculation can be included in https://github.com/openshift/cluster-dns-operator/pull/92

pkg/operator/controller/status.go
danehans: >Given our invariant that default must always exist, agree that 0 dns resources means progressing=true.

Done, latest PR includes `Progressin=true` if `if dnsesAvailable == 0 || dnsesAvailable != dnsesTotal`.

pkg/operator/controller/status.go
```diff
@@ -95,123 +94,147 @@ func (r *reconciler) syncOperatorStatus() error {
...
+		case configv1.OperatorAvailable:
+			oldAvailableCondition = &oldConditions[i]
+		}
 	}
-	return ns, dnsList.Items, daemonsetList.Items, nil
+
+	conditions := []configv1.ClusterOperatorStatusCondition{}
+	conditions = append(conditions, computeOperatorDegradedCondition(oldDegradedCondition, dnsesAvailable, dnses, ns))
+	conditions = append(conditions, computeOperatorProgressingCondition(oldProgressingCondition, dnsesAvailable, dnses))
+	conditions = append(conditions, computeOperatorAvailableCondition(oldAvailableCondition, dnsesAvailable))
```
Miciah: In my earlier comment about using a slice literal, I meant to suggest the following:
```go
	conditions := []configv1.ClusterOperatorStatusCondition{
		computeOperatorDegradedCondition(oldDegradedCondition, dnsesAvailable, dnses, ns),
		computeOperatorProgressingCondition(oldProgressingCondition, dnsesAvailable, dnses),
		computeOperatorAvailableCondition(oldAvailableCondition, dnsesAvailable),
	}
￼```

File: pkg/operator/controller/status.go [hunk 28]
```diff
@@ -95,123 +94,147 @@ func (r *reconciler) syncOperatorStatus() error {
...
+func computeOperatorDegradedCondition(oldCondition *configv1.ClusterOperatorStatusCondition, dnsesAvailable, dnsesTotal int,
+	ns *corev1.Namespace) configv1.ClusterOperatorStatusCondition {
 	degradedCondition := &configv1.ClusterOperatorStatusCondition{
-		Type:   configv1.OperatorDegraded,
-		Status: configv1.ConditionUnknown,
+		Type: configv1.OperatorDegraded,
 	}
-	if ns == nil {
+	switch {
+	case dnsesAvailable == 0 || dnsesAvailable != dnsesTotal && ns == nil:
```
Miciah: `&&` is higher precedence than `||`, so this expression groups as follows: `dnsesAvailable == 0 || (dnsesAvailable != dnsesTotal && ns == nil)`.

Miciah: Since the expression starts to get hairy with the three subexpressions, I'd suggest simplifying by putting `case ns == nil:` first so that subsequent cases can assume the `ns` is not nil.  If the namespace does not exist, then DNSes cannot exist, so I believe it is sufficient just to say that the namespace does not exist if that is the case.

pkg/operator/controller/status.go
pravisankar: When there are no dns resources, I was expected these conditions:
Availabe = False --> No DNS is running and serving the requests
Progressing = False --> Operator is not actually doing any work or trying to transition from one state to another.
Degraded = True --> Operator expects default dns to be present but does not exist which is a failure.
@ironcladlou @Miciah What do you think?

pkg/operator/controller/status.go
pravisankar: Good catch

pkg/operator/controller/status.go
Miciah: > Availabe = False --> No DNS is running and serving the requests

I agree.

> Progressing = False --> Operator is not actually doing any work or trying to transition from one state to another.

I disagree because the operator calls `ensureDefaultDNS` in a periodic goroutine, so the operator is doing work (or will be doing work in less than 1 minute) to create the default resource.

> Degraded = True --> Operator expects default dns to be present but does not exist which is a failure.

I agree.

pravisankar: That makes sense. Okay, progressing=true when there are no dns resources.

pkg/operator/controller/status.go
```diff
@@ -15,24 +13,23 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
-
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	DNSClusterOperatorName = "dns"
+	DNSOperatorName     = "dns"
+	OperatorVersionName = "operator"
+	DNSVersionName      = DNSOperatorName
```
Miciah: This is no longer used in this PR, but no harm in leaving it in since #92 will most likely use it.

Miciah:
level=error msg="\t* module.vpc.aws_route.to_nat_gw[4]: 1 error occurred:"
    level=error msg="\t* aws_route.to_nat_gw.4: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"

/test e2e-aws-upgrade

PR #89, opened 2019-04-04
Removes static GOOS reference from Makefile

danehans:
Previously, `make build` would only build the cluster dns operator bin for linux. This prevents dev's from building and running the operator on a non-Linux machine.

This commit removes static `GOOS=linx` Makefile references, allowing the build to use the local OS. Additionally, this provides parity with the cluster-ingress-operator.

/hold


ironcladlou:
/hold cancel

I consider this a bug in the Makefile, thanks

PR #90, opened 2019-04-10
 Rename 'Cluster DNS' references to 'DNS'

pravisankar:
- Also rename DNS owning and daemonset labels


pravisankar:

ironcladlou:
lgtm, will let @Miciah review/tag to double check

assets/dns/daemonset.yaml
```diff
@@ -77,7 +77,7 @@ spec:
           while true; do
             declare -A svc_ips
             for svc in "${services[@]}"; do
-              # Fetch service IP from cluster dns if present
+              # Fetch service IP from dns if present
```
Miciah: It makes sense to say "cluster" DNS here because the code in question is resolving a cluster-internal name using the cluster-internal nameserver.

README.md
```diff
@@ -1,6 +1,6 @@
-# Cluster DNS Operator
+# DNS Operator
```
Miciah: How about "The DNS Operator" here and below?

assets/dns/daemonset.yaml
pravisankar: fixed

README.md
pravisankar: fixed

pravisankar:
Also renamed binary and manifest files.

pravisankar:
@Miciah updated, PTAL

ironcladlou:
Thanks!

/lgtm

PR #92, opened 2019-04-12
Bug 1705032: Fix updating versions for dns cluster operator

pravisankar:
- Always report 'operator' version irrespective of available condition.

- For upgrades, report old versions until the new version is fully applied.

### DIFF
--- a/manifests/0000_70_dns-operator_03-cluster-operator.yaml
+++ b/manifests/0000_70_dns-operator_03-cluster-operator.yaml
@@ -1,7 +1,8 @@
+# This resource is not created by CVO, it is only used as a communication
+# mechanism to tell the CVO which ClusterOperator resource to wait for.
+# DNS operator creates and updates this resource.
 apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: dns
 spec: {}
-status:
- # status is set at runtime
--- a/pkg/operator/controller/status_test.go
+++ b/pkg/operator/controller/status_test.go
@@ -17,16 +17,19 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
 		degraded, progressing, available bool
 	}
 	type versions struct {
-		operator, operand string
+		operator, coreDNSOperand, openshiftCLIOperand string
+	}
+	type dnsesCount struct {
+		available, total int
 	}
 
 	testCases := []struct {
-		description       string
-		noNamespace       bool
-		dnsesAvail, dnses int
-		reportedVersions  versions
-		curVersions       versions
-		expected          conditions
+		description      string
+		noNamespace      bool
+		dnses            dnsesCount
+		reportedVersions versions
+		curVersions      versions
+		expected         conditions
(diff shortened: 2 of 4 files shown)


ironcladlou:
@Miciah PTAL

pkg/operator/controller/status_test.go
```diff
@@ -347,3 +405,142 @@ func TestOperatorStatusesEqual(t *testing.T) {
...
+
+	testCases := []struct {
+		description       string
+		oldVersions       versions
+		curVersions       versions
+		allDNSesAvailable bool
+		expectedVersions  versions
+	}{
+		{
+			description:       "initialize versions, operator is avaiable",
```
Miciah: Typo: "avaiable"→"available".

Miciah: Here and below, it should be "DNSes available" (or "DNSes not all available", as appropriate).  Saying "operator is available" could be confusing because the operator's `Available` condition will be true if _any_ DNS is available, but `versions` should be updated only if _all_ of the DNSes are available.  (The actual test inputs and expectations look correct; I'm commenting only on the descriptions.)

manifests/0000_70_dns-operator_03-cluster-operator.yaml
```diff
@@ -1,7 +1,8 @@
+# This resource is not created by CVO, it is only used as a communication
+# mechanism to tell the CVO which ClusterOperator resource to wait for.
+# DNS operator creates and updates this resource.
 apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: dns
 spec: {}
```
Miciah: We can delete the `spec` line too.

Miciah: Oh, I see you deleted this earlier and added it back in the latest push.  Do we need it after all? We deleted it in the ingress operator's clusteroperator manifest file.

pravisankar: Yes, `spec` is not really needed but this will break hack/release-local.sh
 hack/release-local.sh when it runs `oc apply -f manifests/` will fail with this below error if spec is not present:
```
The ClusterOperator "dns" is invalid: 
...
.spec in body is required
```
We do not really want to create ClusterOpertor resource as part of hack/release-local.sh. I will remove `spec` from this file along with patch to hack/release-local.sh in a followup pr.

pravisankar:
@Miciah updated, PTAL

manifests/0000_70_dns-operator_03-cluster-operator.yaml
Miciah: Thanks for explaining! As far as I can see, the same issue applies to cluster-ingress-operator, so we will want the same change there.

pravisankar: Follow up pr: https://github.com/openshift/cluster-dns-operator/pull/105

PR #95, opened 2019-04-16
Bug 1699488: Fixes DNS ClusterOperator Initial Status

danehans:
Previously, `clusteroperator/dns` was being created without `status`. Even though the resource was being created with status fields properly constructed and included in the `client.create` call, the resource is initially created without `status`. `clusteroperator/dns` must be created first, then have the status updated.


Miciah:
/hold
Just noticed the title is wrong (it references the Bugzilla report for ingress operator).

Miciah:
Oops.

Probably the correct thing to do would be to clone BZ#1699488, assign the clone to DNS, and put the clone's number in this PR's title.  Might not be too late to do that still.

PR #98, opened 2019-04-18
Bug 1688969: Integrate coredns metrics with openshift-monitoring

pravisankar:
- CoreDNS provides metrics plugin to report prometheus metrics.
This change will enable coredns prometheus metrics to be integrated with
openshift-monitoring component to get better visibility into the dns queries.
- Change details:
   * Create clusterroles/roles needed to allow openshift-monitoring to access DNS metrics endpiont.
   * Add annotation on openshift-dns namesapce to inform openshift-monitoring to look for ServiceMonitor resources.
   * Create ServiceMonitor resource with coredns metrics endpoint that will report prometheus metrics.


pravisankar:

knobunc:
@pravisankar can you add a description to the PR please.

pravisankar:
@knobunc @Miciah @danehans  Updated the description. Ready for review/merge, PTAL

hack/uninstall.sh
```diff
@@ -12,6 +12,8 @@ oc delete namespaces/openshift-dns
 oc delete clusteroperator dns
 oc delete clusterroles/openshift-dns-operator
 oc delete clusterroles/openshift-dns
+oc delete clusterroles/dns-monitoring
```
danehans: I see that clusterroles and clusterrolebindings are deleted but not roles and rolebindings. Is there a reason for this?

pkg/operator/controller/controller.go
```diff
@@ -298,6 +298,59 @@ func (r *reconciler) ensureDNSNamespace() error {
 	return nil
 }
 
+// ensureMetricsIntegration ensures that dns prometheus metrics is integrated with openshift-monitoring for the given DNS.
```
danehans: nit: s/is/are

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -0,0 +1,112 @@
...
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return nil, fmt.Errorf("failed to create servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
+		}
+		logrus.Infof("created servicemonitor %s/%s", desired.GetNamespace(), desired.GetName())
+		return desired, nil
+	}
+	return current, nil
+}
+
+func serviceMonitorName(dns *operatorv1.DNS) types.NamespacedName {
```
danehans: Should this func go in `pkg/operator/controller/names.go`?

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -0,0 +1,112 @@
...
+	return types.NamespacedName{Namespace: "openshift-dns", Name: "dns-" + dns.Name}
+}
+
+func desiredServiceMonitor(dns *operatorv1.DNS, svc *corev1.Service, daemonsetRef metav1.OwnerReference) *unstructured.Unstructured {
```
danehans: Why is an unstructured object being created/returned instead of a servicemonitor object?

hack/uninstall.sh
pravisankar: Roles and RoleBindings are namespace scoped, so deleting 'openshift-dns-operator' namespace will delete these resources as well (https://github.com/openshift/cluster-dns-operator/blob/ed047bd2a2a6d74d45f5505275a5e7f5d1671874/hack/uninstall.sh#L10)

pkg/operator/controller/controller_service_monitor.go
pravisankar: Dynamic client was used instead of ServiceMonitor Client to avoid transitive package dependencies.

pkg/operator/controller/controller_service_monitor.go
pravisankar: fixed.

pkg/operator/controller/controller.go
pravisankar: fixed

pravisankar:
@ironcladlou @danehans updated, PTAL

Gopkg.toml
```diff
@@ -23,6 +23,11 @@ required = [
   name = "github.com/openshift/library-go"
   revision = "dab26bb3a8dc7fccde7227194af755bbff30ce5d"
 
+# coreos
+[[override]]
+  name = "github.com/coreos/prometheus-operator"
+  version = "v0.28.0"
```
Miciah: What is this for?

pkg/operator/controller/controller.go
```diff
@@ -298,6 +298,59 @@ func (r *reconciler) ensureDNSNamespace() error {
...
+			return fmt.Errorf("failed to get dns metrics role binding %s/%s: %v", mrb.Namespace, mrb.Name, err)
+		}
+		if err := r.client.Create(context.TODO(), mrb); err != nil {
+			return fmt.Errorf("failed to create dns metrics role binding %s/%s: %v", mrb.Namespace, mrb.Name, err)
+		}
+		logrus.Infof("created dns metrics role binding %s/%s", mrb.Namespace, mrb.Name)
+	}
+
+	if _, err := r.ensureServiceMonitor(dns, svc, daemonsetRef); err != nil {
+		return fmt.Errorf("failed to ensure servicemonitor for %s: %v", dns.Name, err)
```
Miciah: `ensureDNS` decorates the error message it gets from `ensureMetricsIntegration` with the dns's name, so including the name here is redundant.

pkg/operator/controller/controller_service_monitor.go
danehans: I see that `unstructured` is being used by cluster-ingress-operator too. Thank you.

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (r *reconciler) ensureServiceMonitor(dns *operatorv1.DNS, svc *corev1.Service, daemonsetRef metav1.OwnerReference) (*unstructured.Unstructured, error) {
+	desired := desiredServiceMonitor(dns, svc, daemonsetRef)
+
+	current, err := r.currentServiceMonitor(dns)
+	if err != nil {
+		return nil, err
+	}
+
+	if desired != nil && current == nil {
```
danehans: What about the case where `desired != nil && current != nil`? Should we cover this case and have the client make an `update` call? Something similar to [this](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/controller_router_deployment.go#L39-L42).

danehans: I see that ^ is not implemented in service monitor for the ingress operator.

danehans:
Thanks for addressing my feedback.

/lgtm

PR #99, opened 2019-04-23
Moves Operator Failing status to Degraded

danehans:
Now that https://github.com/openshift/api/pull/286 merged, operators should report `Degraded` instead of `Failing` status. This PR does the following:

- Bumps deps.
- Updates `computeStatusConditions` and supporting unit test to use `Degraded` condition type instead of `Failing`.


danehans:
`level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"`

/test e2e-aws

Miciah:
/lgtm
The new name does however throw into relief that we need to reexamine our interpretation of the condition.

Miciah:
/lgtm
But we need to reexamine our interpretation of the condition at some point.

PR #100, opened 2019-04-25
Add node-selector annotation to namespace

ravisantoshgudimetla:
**Why this change?**
When https://github.com/openshift/cluster-kube-apiserver-operator/pull/394 merges, all the pods running openshift cluster will have a [defaultNodeSelector](https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#setting-the-cluster-wide-default-node-selector) if it has been set by cluster-admin including pods running in `openshift-*` namespace. The main advantage of having that feature is in a multi-tenant environment, any new project created can be steered towards compute nodes(non-master nodes).

**What should I do?**
If you think, the pods in your namespace shouldn't have defaultNodeSelector set, please review this PR. If not feel free to close this. The annotation added as part of this PR lets all the pods created within your project to not have the defaultNodeSelector set.

/cc @deads2k @sjenning


Miciah:
Need to run `make generate` to regenerate bindata.  However, I am not sure we need this at all—we already specify `nodeSelector` on the operator deployment and operand daemonset, so we will not be getting the default node selector anyway, right?

deads2k:
> Need to run `make generate` to regenerate bindata. However, I am not sure we need this at all—we already specify `nodeSelector` on the operator deployment and operand daemonset, so we will not be getting the default node selector anyway, right?

You will and you'll perma-fail

deads2k:
> Need to run `make generate` to regenerate bindata.

you need a verify job.  @ironcladlou @Miciah the library-go build helpers have it built in I think

Miciah:
> You will and you'll perma-fail

Thanks, I didn't realize that the node selector was the union of the two selectors.  (The statement in the documentation, that the default node selector "is applied to the pods created in all projects without a specified nodeSelector value", is ambiguous, but I'll concede that the more likely interpretation is the correct one.)

> you need a verify job. @ironcladlou @Miciah the library-go build helpers have it built in I think

We have a verify job, which is why prow/unit failed.

/lgtm

deads2k:
> We have a verify job, which is why prow/unit failed.

Great.  Thanks.

PR #102, opened 2019-04-30
Fixes operator deployment filename in release-local.sh

danehans:
https://github.com/openshift/cluster-dns-operator/pull/101 introduced a bug by updating the operator deployment manifest filename without updating the reference in release-local.sh. This PR updates the filename in release-local.sh, allowing local releases to properly build


ironcladlou:
Thanks!
/lgtm

PR #104, opened 2019-05-03
Bug 1690342: Update Progressing condition if upgrade happened

Miciah:
* `pkg/operator/controller/status.go`:
* `pkg/operator/controller/status_test.go` (`syncOperatorStatus`): Pass old versions to `computeOperatorStatusConditions`.
(`computeOperatorStatusConditions`): Pass old versions to `computeOperatorProgressingCondition`.
(`computeOperatorProgressingCondition`): Add "Upgraded x to y" message if old and current versions of an image do not match.
(`TestComputeOperatorStatusConditions`): Add test cases.

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -60,7 +60,7 @@ func (r *reconciler) syncOperatorStatus() error {
 
 	co.Status.Versions = r.computeOperatorStatusVersions(oldStatus.Versions, allDNSesAvailable)
 	co.Status.Conditions = r.computeOperatorStatusConditions(oldStatus.Conditions, ns, dnsesTotal,
-		dnsesAvailable, co.Status.Versions)
+		dnsesAvailable, oldStatus.Versions, co.Status.Versions)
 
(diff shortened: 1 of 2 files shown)


danehans:
@Miciah it's interesting that the ingress operator does not need this fix: https://bugzilla.redhat.com/show_bug.cgi?id=1690333

Miciah:
> @Miciah it's interesting that the ingress operator does not need this fix: https://bugzilla.redhat.com/show_bug.cgi?id=1690333

Also, https://bugzilla.redhat.com/show_bug.cgi?id=1695210.  I'm not entirely confident that the problem does not still exist in the ingress operator; it may only be more difficult to trigger.

Miciah:
Note that in both https://bugzilla.redhat.com/show_bug.cgi?id=1690333#c6 and https://bugzilla.redhat.com/show_bug.cgi?id=1695210#c2, the available condition changes, which means that at some point, not all ingress controllers are available, and that is sufficient to cause the progressing condition to change.  If an upgrade happens without causing ingress controllers to become unavailable, then the ingress operator may fail to update the progressing condition.

PR #106, opened 2019-05-06
Bug 1690342: Fix Progressing condition upgrade check

Miciah:
Fix-up to https://github.com/openshift/cluster-dns-operator/pull/104.

* `pkg/operator/controller/status.go` (`computeOperatorProgressingCondition`): Move the check for a mismatch between old version and reported version before the check for a mismatch between current version and reported version.  Make sure to use the right string for the condition's message.

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -227,6 +227,9 @@ func (r *reconciler) computeOperatorProgressingCondition(oldCondition *configv1.
 	}
 
 	for _, opv := range curVersions {
+		if oldVersion, ok := oldVersionsMap[opv.Name]; ok && oldVersion != opv.Version {
+			messages = append(messages, fmt.Sprintf("Upgraded %s to %q.", opv.Name, opv.Version))
+		}
 		switch opv.Name {
 		case OperatorVersionName:
 			if opv.Version != r.OperatorReleaseVersion {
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/status.go
```diff
@@ -243,10 +246,6 @@ func (r *reconciler) computeOperatorProgressingCondition(oldCondition *configv1.
 				messages = append(messages, fmt.Sprintf("Moving to openshift-cli image version %q.", r.OpenshiftCLIImage))
 				progressing = true
 			}
-		default:
-			if oldVersion, ok := oldVersionsMap[opv.Name]; ok && oldVersion != opv.Version {
-				messages = append(messages, fmt.Sprintf("Upgraded %s to %q.", opv.Name, opv.Version))
-			}
```
pravisankar: https://github.com/openshift/cluster-dns-operator/blob/7fc741b2e4abe724457ecee73d8297dc4c02fd09/pkg/operator/controller/status.go#L252 may not be sufficient to print 'Upgraded ...' msg in all cases. I think we need:
```
if progressing {
	progressingCondition.Status = configv1.ConditionTrue
	progressingCondition.Reason = "Reconciling"
} else {
	progressingCondition.Status = configv1.ConditionFalse
	progressingCondition.Reason = "AsExpected"
}

if len(messages) > 0 {
	progressingCondition.Message = strings.Join(messages, "\n")
} else {
	progressingCondition.Message = "Desired and available number of DNS DaemonSets are equal"
}
```

Miciah: Thanks! Fixed.

PR #109, opened 2019-05-17
Bug 1711364: Report operator is degraded if any DNS is degraded

Miciah:
If any DNS resource's degraded condition is true, set the clusteroperator's degraded condition to be true as well.

Note: An implication of this change is that the clusteroperator will report degraded true during upgrades.

* `pkg/operator/controller/status.go` (`dnsStatusConditionsCounts`): New struct type.  Define a field for each status condition, to count how many DNSes have that condition, as well as field for the total number of DNSes.
(`syncOperatorStatus`): Use `computeDNSStatusConditionCounts` to get a `dnsStatusConditionsCounts` value, and pass this value to `computeOperatorStatusVersions` and `computeOperatorStatusConditions`.  Delete the `dnsesTotal`, `dnsesAvailable`, and `allDNSesAvailable` variables.
(`computeDNSStatusConditionCounts`): New function.  Take a slice of DNSes and return a `dnsStatusConditionsCounts` value.
(`computeOperatorStatusVersions`): Replace the `allDNSesAvailable` parameter with a `dnsStatusConditionsCounts` parameter.
(`computeOperatorStatusConditions`): Replace the `dnses` and `dnsesAvailable` parameters with a `dnsStatusConditionsCounts` parameter.
(`computeOperatorDegradedCondition`):
(`computeOperatorProgressingCondition`): Replace the `dnsesAvailable` and `dnsesTotal` parameters with a `dnsStatusConditionsCounts` parameter.
(`computeOperatorAvailableCondition`): Replace the `dnsesAvailable` parameter with a `dnsStatusConditionsCounts` parameter.
(`numDNSesAvailable`): Delete.
* `pkg/operator/controller/status_test.go` (`TestComputeOperatorStatusConditions`): Use `dnsStatusConditionsCounts`. Adjust test cases to expect the degraded condition to be true during upgrades.
(`TestComputeOperatorStatusVersions`): Use `dnsStatusConditionsCounts`.

### DIFF
--- a/pkg/operator/controller/status_test.go
+++ b/pkg/operator/controller/status_test.go
@@ -19,14 +19,11 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
 	type versions struct {
 		operator, coreDNSOperand, openshiftCLIOperand string
 	}
-	type dnsesCount struct {
-		available, total int
-	}
 
 	testCases := []struct {
 		description      string
 		noNamespace      bool
-		dnses            dnsesCount
+		dnses            dnsStatusConditionsCounts
 		reportedVersions versions
 		oldVersions      versions
 		curVersions      versions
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/status_test.go
```diff
@@ -39,97 +36,98 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
...
 		{
 			description: "0/0 dns resources available",
+			dnses:       dnsStatusConditionsCounts{0, 0, 0, 0},
 			expected:    conditions{true, true, false},
 		},
 		{
 			description: "1/2 dns resources available",
-			dnses:       dnsesCount{1, 2},
+			dnses:       dnsStatusConditionsCounts{1, 1, 1, 2},
 			expected:    conditions{true, true, true},
```
ironcladlou: Could we use field labels in these structs to visually hint at the value meanings?

Miciah: Yeah, at one point we tried to make them more compact, but I can expand them again.

Miciah: Fixed.  Does it look all right?

danehans:
```
fail [github.com/openshift/origin/test/extended/builds/new_app.go:56]: Expected error:
    <*errors.errorString | 0xc002db8ae0>: {
        s: "The build \"a234567890123456789012345678901234567890123456789012345678-1\" status is \"Failed\"",
    }
```

/test e2e-aws

ironcladlou:
@smarterclayton 
>Note: An implication of this change is that the clusteroperator will report degraded true during upgrades.

Given the DaemonSet's update strategy:

```
updateStrategy:
  rollingUpdate:
    maxUnavailable: 1
  type: RollingUpdate
```

A degraded status during an upgrade seems correct — agree?

Alternative would be to allow surge (which has some other considerations)?

smarterclayton:
Is this consistent with previous statements we have made about upgrades
being blocked by certain machines being down?  Updating one at a time seems
questionable.

@derekwaynecarr

smarterclayton:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/785/pull-ci-openshift-machine-config-operator-master-e2e-aws-upgrade/498

seeing you reporting an error here.

I'm not sure this is an error condition as such, especially during
upgrade.  I'm not sure you should start reporting degraded until a period
of time has passed or progress has halted.

PR #115, opened 2019-06-03
 Switch to Go modules, Go 1.12, and latest libraries 

ironcladlou:
Migrate to Go modules. Switch to Go 1.12. Upgrade to Kube 1.14 and the latest
controller-runtime.

Switch CRD generation away from the openshift fork and to the upstream
controller-tools generator.

Motivations for modules:

1. Modules resolves a transitive controller-tools conflict between
controller-runtime and library-go which `dep` could not accomodate.

2. Kubernetes will soon be using modules, and some related community projects on
which we depend are also migrating. The Kubernetes KEP has a good motivation
section which applies very broadly[1].

There are some important notes for contributors:

1. Vendoring is enabled for simple reproducible builds. It's therefore important
to explicitly enable vendoring in various Go execution contexts if
reproducibility is a concern. All the build tooling has been updated to enable
vendoring. Any end-user scripts or tooling should be audited for vendoring
considerations.[2]

2. If your project lives in a GOPATH, you may have to move your project or
perform extra steps to activate modules.

[1] https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md#motivation
[2] https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away
[3] https://github.com/golang/go/wiki/Modules#how-to-use-modules

Depends on:
- [x] https://github.com/openshift/release/pull/3946
- [x] https://github.com/openshift/api/pull/336


ironcladlou:
@Miciah updated to incorporate feedback from https://github.com/openshift/cluster-ingress-operator/pull/243, PTAL

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
```diff
@@ -1,20 +1,22 @@
...
   names:
     kind: DNS
-    listKind: DNSList
-    plural: dnses
-    singular: dns
-  scope: Cluster
-  version: v1
-  subresources:
-    status: {}
+    plural: dns
```
Miciah: Is this intentional?

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
```diff
@@ -1,20 +1,22 @@
...
+  scope: ""
```
Miciah: This looks totes wrong.

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
```diff
@@ -1,20 +1,22 @@
...
   group: operator.openshift.io
   names:
     kind: DNS
-    listKind: DNSList
-    plural: dnses
-    singular: dns
-  scope: Cluster
-  version: v1
-  subresources:
-    status: {}
```
Miciah: Won't this break status reporting?

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
ironcladlou: Not sure what's going on there, need to debug

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
ironcladlou: Will be fixed in https://github.com/openshift/api/pull/336

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
ironcladlou: This could be a problem: https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/crd.go#L39

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
ironcladlou: Figured this out, fixed in https://github.com/openshift/api/pull/336

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
ironcladlou: https://github.com/kubernetes-sigs/controller-tools/pull/220

ironcladlou: Will be fixed by https://github.com/openshift/api/pull/336 and https://github.com/kubernetes-sigs/controller-tools/pull/220

ironcladlou:
@Miciah @danehans okay, all the upstream and api work is done and this should be ready for review.

ironcladlou:
Will remove hold once all reviewers have tagged.

/hold

manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
```diff
@@ -68,6 +428,23 @@ spec:
...
+          required:
+          - clusterIP
+          - clusterDomain
           type: object
+      type: object
+  versions:
+  - name: v1
+    served: true
+    storage: true
+status:
```
Miciah: Seems like the `status:` stanza should be omitted, but it shouldn't hurt anything.

Dockerfile.rhel7
```diff
@@ -1,10 +1,10 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
```
Miciah: Oops!  Missed this.  https://github.com/openshift/cluster-dns-operator/pull/120

PR #116, opened 2019-06-10
Bug 1711373: Refactor client and cache handling

ironcladlou:
* Delete custom cache setup, no longer necessary
* Consolidate client usage and use dynamic discovery (see https://github.com/kubernetes-sigs/controller-runtime/issues/321 — hat tip to @danwinship for https://github.com/openshift/cluster-network-operator/pull/95). Fixes [bz1711373](https://bugzilla.redhat.com/show_bug.cgi?id=1711373).
* Plumb the cache through for future use


Miciah:
/lgtm
/hold
@danehans, would you like to give it a look over before we merge?

pkg/operator/operator.go
```diff
@@ -60,38 +65,37 @@ func New(config operatorconfig.Config) (*Operator, error) {
...
 // Start creates the default DNS and then starts the operator
 // synchronously until a message is received on the stop channel.
 // TODO: Move the default DNS logic elsewhere.
 func (o *Operator) Start(stop <-chan struct{}) error {
-	// Periodicaly ensure the default controller exists.
+	// Periodicaly ensure the default dns exists.
 	go wait.Until(func() {
-		if err := o.ensureDefaultDNS(); err != nil {
-			logrus.Errorf("failed to ensure default dns: %v", err)
+		if !o.manager.GetCache().WaitForCacheSync(stop) {
```
danehans: @Dan why is `GetCache()` called if `Manager` uses a non-caching client?

pkg/operator/operator.go
```diff
@@ -35,23 +35,28 @@ type Operator struct {
...
+		Namespace:      "openshift-dns",
+		MapperProvider: operatorutil.NewDynamicRESTMapper,
+		// Use a non-caching client everywhere. The default split client does not
+		// promise to invalidate the cache during writes (nor does it promise
+		// sequential create/get coherence), and we have code which (probably
+		// incorrectly) assumes a get immediately following a create/update will
+		// return the updated resource. All client consumers will need audited to
+		// ensure they are tolerant of stale data (or we need a cache or client that
+		// makes stronger coherence guarantees).
+		NewClient: func(_ cache.Cache, config *rest.Config, options client.Options) (client.Client, error) {
```
danehans: Is the operator caches field no longer needed? https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/operator.go#L33

ironcladlou: Nothing's using the cache yet, but soon will, so "reserved for future use" — if you'd rather I delete it until needed let me know

pkg/operator/operator.go
ironcladlou: See https://github.com/openshift/cluster-dns-operator/pull/116#discussion_r292221318

(p.s. you tagged the wrong dan)

ironcladlou: Also, the cache is actually still involved in the underlying event system — the rationale discussed in https://github.com/openshift/cluster-ingress-operator/pull/244#discussion_r290808716 applies here too. Not strictly necessary, but in terms of lifecycle it seemed sane to wait for the framework to hit something resembling a steady state before doing anything else.

pkg/operator/operator.go
danehans: Thanks for the explanation. I think it's fine to stay. Just trying to make sure I understand what's going on.

danehans:
@ironcladlou thanks for the additional background on the changes.

/lgtm

PR #117, opened 2019-06-10
metrics: fix metrics registration bug

ironcladlou:
Use a controller name which is always a valid metric name, allowing the
controller-runtime to successfully register all metrics which incorporate
the controller's name.

TODO:
- [x] Merge https://github.com/openshift/cluster-dns-operator/pull/116


ironcladlou:
@Miciah rebased, please re-tag

PR #120, opened 2019-06-11
Dockerfile.rhel7: Use golang-1.12

Miciah:
Follow-up to https://github.com/openshift/cluster-dns-operator/pull/115.

* `Dockerfile.rhel7`: Use the golang-1.12 builder image.


ironcladlou:
Good catch — wonder how that's working given absence of GOPATH

/lgtm

PR #123, opened 2019-07-08
corefile: switch deprecated proxy plugin to forward plugin

ironcladlou:
Switch the deprecated[1] `proxy` plugin to the `forward` plugin. The current
configuration used by the operator is drop-in compatible with the `forward`
plugin, and so this change should be completely transparent.

To support rolling out the Corefile change, add support for diffing and applying
changes to the Corefile configmap.

[1] https://github.com/coredns/coredns/pull/2744

Should unblock https://github.com/openshift/coredns/pull/15 (which removes the `proxy` plugin from CoreDNS).

/cc @celebdor


ironcladlou:
There may be an issue here with the daemonset image being updated before the configmap and th configmap change not coercing a rollout.

ironcladlou:
If the configmap changes at any time, should we coerce a rollout? Otherwise it becomes very unpredictable when the changes will be picked up (when CoreDNS pods happen to restart later on for various reasons).

ironcladlou:
@Miciah pointed out we're already using the [reload](https://coredns.io/plugins/reload/) plugin, so a change to the mounted contents should cause CoreDNS to restart internally. I'll test to make sure.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -63,3 +74,12 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string, daemonsetRef
 	}
 	return cm
 }
+
+func corefileChanged(current, expected *corev1.ConfigMap) (bool, *corev1.ConfigMap) {
+	if cmp.Equal(current.Data, expected.Data, cmpopts.EquateEmpty()) {
```
Miciah: This is fine, but out of curiosity, is there a reason not to use `reflect.DeepEqual` here?

ironcladlou:
Confirmed that mutating the Corefile configmap causes a reload:

```
$ oc logs -f -n openshift-dns pods/dns-default-5nnrf -c dns
2019-07-08T17:33:24.947Z [INFO] CoreDNS-1.3.1
2019-07-08T17:33:24.947Z [INFO] linux/amd64, go1.10.3,CoreDNS-1.3.1linux/amd64, go1.10.3,
2019-07-08T17:33:24.947Z [INFO] plugin/reload: Running configuration MD5 = 6dfacbfa08660b953611ad25ea5c84fc
[INFO] Reloading
2019-07-08T17:48:20.695Z [INFO] plugin/reload: Running configuration MD5 = d9f1b0da2003417fa038d3bb0a8ccef4
[INFO] Reloading complete
```

pkg/operator/controller/controller_dns_configmap.go
ironcladlou: I want to say it was actually you who started proliferating the use of [go-cmp](https://github.com/google/go-cmp) into the operators 😁 — for the most part I try to just use it by default now for consistency and flexibility

Miciah: Ravi introduced the first use of go-cmp, and mostly we use it where `reflect.DeepEqual` is not flexible enough, but yeah, we're now using go-cmp more than `reflect.DeepEqual`, so it is reasonable to choose go-cmp over `reflect.DeepEqual` by default.

PR #124, opened 2019-07-25
Add tolerations permissions

ravisantoshgudimetla:
As of now, tolerations to pod spec can be added by any user and the tolerations sometimes helps
in landing on master nodes. Using https://github.com/openshift/origin/pull/23427, we're moving to
a RBAC model to ensure that tolerations can be applied to a pod by service account with privileges to access them.


ironcladlou:
@Miciah @danehans PTAL

ravisantoshgudimetla:
@ironcladlou - I pushed the changes related to operand as well. PTAL

assets/dns/cluster-role.yaml
```diff
@@ -13,3 +13,13 @@ rules:
   verbs:
   - list
   - watch
+- apiGroups:
+  - "toleration.scheduling.openshift.io"
+  resources:
+  - ["*"]
```
Miciah: List of lists seems wrong; shouldn't that be the following?
```yaml
  resources:
  - "*"
```
or
```yaml
  resources: ["*"]
```

ironcladlou: ping @ravisantoshgudimetla

ironcladlou:
Still needs a rebase and response to https://github.com/openshift/cluster-dns-operator/pull/124/files#r307337877.

ravisantoshgudimetla:
Ahh this not needed anymore @ironcladlou :). 

Background: The main reason was to treat tolerations as resources that only certain roles can access. As of now, we're not this route.

So, I am closing this PR. Sorry for not being responsive.

PR #125, opened 2019-07-25
Bug 1726449: manifests: add version information

ironcladlou:
Add version info to manifest so the CVO can identify the operator's
version keys in the clusteroperator status.

### DIFF
--- a/manifests/0000_70_dns-operator_03-cluster-operator.yaml
+++ b/manifests/0000_70_dns-operator_03-cluster-operator.yaml
@@ -5,3 +5,7 @@ apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: dns
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"


manifests/0000_70_dns-operator_03-cluster-operator.yaml
```diff
@@ -5,3 +5,11 @@ apiVersion: config.openshift.io/v1
...
 metadata:
   name: dns
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
+    - name: coredns
+      version: "0.0.1-snapshot"
+    - name: operator-cli
+      version: "0.0.1-snapshot"
```
Miciah: [The CVO documentation](https://github.com/openshift/cluster-version-operator/blob/8402d219f36fc79e03edf45918785376113f2cc1/docs/dev/clusteroperator.md#what-should-be-the-contents-of-clusteroperator-custom-resource-in-manifests) says only to add an entry for the operator itself.

Miciah: Corroborating the documentation, Derek's original PR only added an entry for the operator: https://github.com/openshift/cluster-dns-operator/pull/78/files#diff-cc9f42cdd0bb164691de71f425fd6f6eR8

ironcladlou: Fixed

ironcladlou:
@Miciah tag?

PR #127, opened 2019-08-14
Bug 1740375: status/operator: improve availability reason

ironcladlou:
Improve the reason associated with an unavailable DNS operator condition.

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -332,8 +332,8 @@ func computeOperatorAvailableCondition(oldCondition *configv1.ClusterOperatorSta
 		availableCondition.Message = "At least 1 DNS DaemonSet available"
 	} else {
 		availableCondition.Status = configv1.ConditionFalse
-		availableCondition.Reason = "DNSUnavailable"
-		availableCondition.Message = "No DNS DaemonSets available"
+		availableCondition.Reason = "NoDaemonSetsAvailable"
+		availableCondition.Message = "No DNS DaemonSets are available"
 	}
 
 	setOperatorLastTransitionTime(&availableCondition, oldCondition)


pkg/operator/controller/status.go
```diff
@@ -332,8 +332,8 @@ func computeOperatorAvailableCondition(oldCondition *configv1.ClusterOperatorSta
 		availableCondition.Message = "At least 1 DNS DaemonSet available"
 	} else {
 		availableCondition.Status = configv1.ConditionFalse
-		availableCondition.Reason = "DNSUnavailable"
-		availableCondition.Message = "No DNS DaemonSets available"
+		availableCondition.Reason = "NoDaemonSetsAvailable"
+		availableCondition.Message = "No DNS DaemonSets are available"
```
Miciah: The reason and message may be misleading.  The problem is that no DNS resource is available.  The default DNS resource may be unavailable because the Service does not have an address assigned even while the DaemonSet is available.

ironcladlou: Since the DNS is a singleton, should the operator availability reason mirror the DNS reason in this case? @bparees

bparees: Seems plausible to me....

ironcladlou: Likewise for the Degraded condition?

bparees: yeah i think so

ironcladlou: Unfortunately we need to do a bit of refactoring here to propagate the reason, lifting newer techniques from ingress-operator. Will keep this PR open, but not sure if I'll get back to it in the 4.2 dev cycle.

ironcladlou:
Incomplete.

/close

PR #129, opened 2019-09-27
Bug 1756344: Add `policy sequential` to Corefile

tomassedovic:
This instructs CoreDNS to always try the first forwarder in the list
rather then a random one (the default behaviour).

OpenStack and BareMetal rely on this with their static coredns pods.
They set the first `resolv.conf` entry to the DNS serving the
necessary OpenShift records (`api`, `api-int`, `*.apps`, etcd SRV and
hostnames) and forward anything else to all the subsequent
nameservers (see[1][2] for more details about the architecture).

[1]: https://github.com/openshift/installer/blob/master/docs/design/openstack/networking-infrastructure.md
[2]: https://github.com/openshift/installer/blob/master/docs/design/baremetal/networking-infrastructure.md

When the cluster DNS (`openshift-dns/dns-default`) pods randomly
select anything but the first entry it will fail to resolve the
records above. This happens frequently for the `default-dns`
service (`172.30.0.10`) and results in issues such as the
`authentication-operator` becoming degraded every few minutes.


Miciah:
The manpage for `resolv.conf` states, "If there are multiple servers, the resolver library queries them in the order listed", so using the CoreDNS forward plugin's default random policy is surprising and contrary to the most obviously applicable documentation, and we indeed should rather use `policy sequential` with `resolv.conf`.

Miciah:
Provisioning failures.
/retest

Miciah:
@ironcladlou, this looks good to me; any objections?

ironcladlou:
If we grant that the default setting (random) was deliberately chosen upstream, it might be useful to understand why.

Default aren't always appropriate, but this PR implies that the upstream default is actually the wrong choice for OpenShift. The given justification seems based on an implementation detail for a certain set of use cases that I'm not sure represents a better default for everyone. Are there any implications of promoting assumptions about forwarding order on the platform? It feels brittle.

At the least, I'd expect some idea of what the expected impact would be to users who might actually prefer or rely on the current default.

Or maybe we need config API around the forwarding behavior?

Miciah:
The upstream documentation is defective.  Either

* the example at https://coredns.io/plugins/forward/ should not use `resolv.conf`;
* the example that uses `resolv.conf` should use `policy sequential`; or
* the documentation needs to explain why it is violating well established norms for the semantics of the ordering of nameservers in `resolv.conf`.

Should I file an issue upstream?

Miciah:
I found https://github.com/coredns/coredns/issues/2830, which was closed even after the reporter quoted the same statement regarding ordering in `resolv.conf` that I quoted above.

tomassedovic:
@ironcladlou I couldn't find any reasoning behind CoreDNS' default behaviour, but it is against what the Kubernetes documentation recommends:

> If a Pod’s dnsPolicy is set to “default”, it inherits the name resolution configuration from the node that the Pod runs on. The Pod’s DNS resolution should behave the same as the node.

https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#introduction

With cluster-dns in place we're not using the default `dnsPolicy` and I could not find any more info on such deployments. But the `sequential` policy seems to follow the spirit of the `resolv.conf` forwarding in Kubernetes.

For OpenStack and Bare Metal, the `random` policy is definitely incorrect. In addition, it goes against the expected semantics of the `resolv.conf` file (it's not supposed to be a load balancer, rather a sequence of fallbacks for graceful degradation).

I don't know enough about how AWS etc. work, but my assumption is that the random policy there happens to work because their resolv.conf `nameserver` entries are either equivalent or they've only got a single one there.

For OpenStack and Bare Metal (and other platforms without DNSaaS), we would have to work around this limitation by editing resolv.conf and removing the extra nameservers, leaving only our own entry there. In addition to that being more platform-specific code (and more code in general), it would make things more brittle: if this DNS went down, cluster-dns would stop serving all records, rather than just the internal ones.

I think switching the forward policy would not have any negative impact for the other platforms:

If they only have a single `nameserver`, the behaviour is identical. If they've got more, the `random` policy means the DNS requests will be spread out rather than hitting the first entry preferentially. But the current cluster-dns config already has `cache 30` so it should not be hitting the upstream nameservers that often.

Making this policy configurable would be a viable alternative, but I don't think there's a good reason to keep the current default.

/cc @celebdor and @yboaron (Bare Metal folks)

mandre:
It's perfectly fine for coredns to use a `random` policy by default in the forward plugin (in fact, at some point `random` was the only policy https://github.com/coredns/coredns/issues/1462), however it becomes problematic when forwarding to `/etc/resolv.conf` as it breaks the assumption that nameservers are queried in the order listed. In that case, a better default should be `sequential`.

IMO, the right thing to do is to set the policy to `sequential` in cluster-dns-operator (because we're forwarding to `/etc/resolv.conf`) and log an issue with coredns to change their default policy to `sequential` when forwarding to `/etc/resolv.conf`.

tomassedovic:
Yeah I was not suggesting we change the default policy in CoreDNS. Just in the DNS operator here.

ironcladlou:
Thanks for all the detailed feedback. I am now convinced that when forwarding to `/etc/resolv.conf` specifically, the `sequential` strategy is consistent with existing resolver semantics.

@Miciah, please tag if you are still in agreement.

/approve

ironcladlou:
P.S., this seems like a pretty easy backport candidate

tomassedovic:
Thanks, I'm on it!

ironcladlou:
If we want backports, opening the relevant bugzillas and using /cherrypick should be enough

tomassedovic:
I've cloned the BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=1757124

Will add it to the new PR (unless the cherry-pick bot does it automatically)

PR #131, opened 2019-10-02
Update to Kube 1.16

ironcladlou:


manifests/0000_70_dns-operator_00-custom-resource-definition.yaml
```diff
@@ -428,16 +75,14 @@ spec:
                     type: string
                   type:
                     type: string
-                required:
-                - type
-                - status
```
ironcladlou: Same as https://github.com/openshift/cluster-ingress-operator/pull/301#discussion_r327685214 (https://github.com/openshift/api/pull/445)

ironcladlou:
@Miciah this is ready for review,  PTAL

ironcladlou:
@Miciah 
>We can regenerate in a follow-up PR once openshift/api#445 and openshift/api#455 are merged.

https://github.com/openshift/api/pull/455 is already incorporated (https://github.com/openshift/cluster-dns-operator/pull/131/commits/fd88612ce7b79b386e3729d61b72414147946349#diff-994cc907b223d50dacb1ba617054962cR81)

Miciah:
You must have pushed an update between when I loaded the page and when I posted `/lgtm`.  GitHub is scary.

PR #134, opened 2019-10-11
status: prevent degraded status flapping on rollout

ironcladlou:
status: prevent degraded status flapping on rollout

The current DNS degraded status calculation doesn't incorporate time. If at any
time of observation the number of desired daemonset available replicas diverges
from the desired scheduled replicas, the CoreDNS daemonset is instantly
considered degraded. This results in degraded true/false flapping when the
operator aggregates the DNS status during a daemonset rollout.

This seems like a poor meausure of degraded, because most of the time this
condition arises during a rollout when QoS is not seriously affected. A better
way might be to incorprate progressing time into the calculation, and that is
something we should probably do, but the solution is a little more complex.

This patch should provide a good improvement in the meantime by only considering
DNS degraded if the number of unavailable replicas exceeds the max unavailable
count on the daemonset's rolling parameters, which seems like an objectively
poor state to be in.

And, as mentioned, we can further improve the situation later by trying to
incorporate progressing time. That later refinement could replace the logic
introduced in this commit.

### DIFF
--- a/pkg/operator/controller/dns_status.go
+++ b/pkg/operator/controller/dns_status.go
@@ -61,6 +61,7 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 	degradedCondition := &operatorv1.OperatorCondition{
 		Type: operatorv1.OperatorStatusTypeDegraded,
 	}
+	numberUnavailable := ds.Status.DesiredNumberScheduled - ds.Status.NumberAvailable
 	switch {
 	case len(clusterIP) == 0 && ds.Status.NumberAvailable == 0:
 		degradedCondition.Status = operatorv1.ConditionTrue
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/dns_status.go
```diff
@@ -70,14 +70,14 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 		degradedCondition.Status = operatorv1.ConditionTrue
 		degradedCondition.Reason = "NoClusterIP"
 		degradedCondition.Message = "No ClusterIP assigned to DNS Service"
-	case ds.Status.NumberAvailable == 0:
+	case ds.Status.DesiredNumberScheduled == 0:
```
wking: This should stay `NumberAvailable == 0`, right?

wking: i.e. wanting more than one pod is not enough, you actually have to *have* more than one pod to get out of `NoPodsScheduled`

ironcladlou: NoPodsScheduled is actually the misnomer, the check is intentional... if none are even _desired_ to be scheduled, then available seems irrelevant. I'm not even sure how we can get here in the wild...

ironcladlou: NoPodsDesired?

ironcladlou: Maybe DesiredNumberScheduled=0 if none of the nodes match the scheduling criteria due to taints or something.

pkg/operator/controller/dns_status.go
```diff
@@ -70,14 +70,14 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
 		degradedCondition.Status = operatorv1.ConditionTrue
-		degradedCondition.Reason = "DaemonSetUnavailable"
-		degradedCondition.Message = "DaemonSet pod not running on any Nodes"
-	case ds.Status.NumberAvailable != ds.Status.DesiredNumberScheduled:
+		degradedCondition.Reason = "NoPodsScheduled"
+		degradedCondition.Message = "No CoreDNS pods are desired to be scheduled"
+	case ds.Status.DesiredNumberScheduled > 0 && ds.Status.NumberAvailable <= 1:
```
wking: @Miciah suggested `ds.Status.DesiredNumberScheduled - ds.Status.NumberAvailable > ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable` for this case, and I like that better than `NumberAvailable <= 1`, because it means that Kubernetes is violating our DaemonSet rollout condition, and presumably you could grow a cluster to be large enough that a single CoreDNS could not serve the whole cluster.

jwforres:
@ironcladlou trying to understand @wking 's comment here https://github.com/openshift/machine-api-operator/pull/417#issuecomment-543846501

Does this mean this change introduces a regression, or did it just uncover another thing that should be cleaned up.

ironcladlou:
Not sure that I understand the question, can you try another way?

PR #137, opened 2019-10-29
Removes replace of controller-runtime in go.mod

danehans:
Replacing `sigs.k8s.io/controller-runtime` with `github.com/munnerz/controller-runtime` is no longer required. Bumps controller-runtime to be compatible with k8s 1.16 and bumps openshift/api to `abf28f9`.


danehans:
https://github.com/openshift/cluster-ingress-operator/pull/316 is the same bump for ingress-operator.

ironcladlou:
Thanks!

/approve
/lgtm

PR #140, opened 2019-10-31
Bug 1753059: Don't start DNS on NotReady nodes

ironcladlou:
Relax the tolerations for CoreDNS so that it doesn't get started on
nodes that aren't yet ready.


assets/dns/daemonset.yaml
```diff
@@ -137,5 +137,6 @@ spec:
           path: /etc/hosts
           type: File
       tolerations:
-      # tolerate all taints so that DNS is always present on all nodes
-      - operator: Exists
+      # DNS needs to run everywhere.
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
```
ironcladlou: I'm not sure if this is exactly what we need, but I wanted to start the convo.

ironcladlou:
@Miciah PTAL, tests seem good

PR #141, opened 2019-11-07
manifests: Use CRD from openshift/api

Miciah:
Copy the CRD from the vendored `openshift/api` into the operator's manifests instead of generating the CRD using `controller-gen` from controller-tools.

* `hack/update-generated-crd.sh`: Generate the CRD by copying it CRD from the vendored `openshift/api`.
* `hack/verify-generated-crd.sh`: Compare against the CRD in the vendored `openshift/api`.
* `tools/tools.go`: Drop `controller-gen`.
* `go.mod`:
* `go.sum`:
* `manifests/0000_70_dns-operator_00-custom-resource-definition.yaml`:
* `vendor/*`: Regenerate.


ironcladlou:
Nice, thanks

/lgtm

PR #142, opened 2019-11-20
WIP: metrics: Add metrics with number of total dnses, dnses using forwarding API

Miciah:
Add metrics with the number of dns resources and the number of dnses that are using the forwarding API.

* `pkg/operator/controller/metrics/metrics.go`: New file.
(`DNSesTotal`, `DNSesUsingForwardingAPI`): New metrics.
(`Update`): New function.  Update `DNSesTotal` and `DNSesUsingForwardingAPI` using the given slice of DNSes.
(`init`): Register `DNSesTotal` and `DNSesUsingForwardingAPI`.
* pkg/operator/controller/status.go (`syncOperatorStatus`): Call `metrics.Update`.

----

This PR has two purposes: (1) establish a pattern for adding operator metrics (@miheer, you were interested in this), and (2) provide visibility as to how many clusters actually use the new forwarding API (idea from @brenton).  @ironcladlou, what do you think?

I tested manually and verified in Prometheus that `dns_operator_dnses_using_forwarding_api` indeed shows 0 initially, 1 after I patch the `default` dns to use the new forwarding API, and 0 again if I subsequently patch the `default` dns resource's `spec.servers[0].forwardPlugin` field to `null`.  So it works.  Suggestions for better metrics names or other improvements are welcome.

### DIFF
--- a/pkg/operator/controller/metrics/metrics.go
+++ b/pkg/operator/controller/metrics/metrics.go
@@ -1,0 +1,46 @@
+package metrics
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	// DNSesTotal is a Prometheus gauge metric which holds the total number
+	// of DNSes.
+	DNSesTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "dns_operator_dnses_total",
+		Help: "Total number of DNSes",
+	}, []string{"dnses"})
+
+	// DNSesUsingForwardingAPI is a Prometheus gauge metric which indicates
+	// how many DNS operands are using the Forwarding API.
+	DNSesUsingForwardingAPI = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "dns_operator_dnses_using_forwarding_api",
+		Help: "Number of DNSes using the forwarding API",
+	}, []string{"dnses"})
+)
+
+func Update(dnses []operatorv1.DNS) {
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	// DNSesTotal is a Prometheus gauge metric which holds the total number
+	// of DNSes.
+	DNSesTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
Miciah: Not sure there's a need to export the metrics variables.

pkg/operator/controller/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+		Name: "dns_operator_dnses_total",
+		Help: "Total number of DNSes",
+	}, []string{"dnses"})
+
+	// DNSesUsingForwardingAPI is a Prometheus gauge metric which indicates
+	// how many DNS operands are using the Forwarding API.
+	DNSesUsingForwardingAPI = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
Miciah: I'd just as soon use a Boolean, but Prometheus does not appear to have a Boolean metric type.  My thinking is that we will add an additional metric for each major feature so as to provide insight as to whether that feature is being used; would it make more sense to have a single "features" metric with a label for each feature, or is there some more idiomatic approach?

ironcladlou: @smarterclayton can you give us a little guidance on the idioms for metric labelling here?

PR #147, opened 2019-12-04
Bug 1779490: dns-node-resolver sometimes exits with code 255

smarterclayton:
If the child processes are killed before trap exits, we can exit
with an error instead of zero:

```
Dec 04 01:43:12.495 E ns/openshift-dns pod/dns-default-954mg node/ip-10-0-133-153.ec2.internal container=dns-node-resolver container exited with code 255 (Error): kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]\n
```

Instead, use `xargs -r` like many of our other shell containers to
ignore when no children exist.

46 hits in the last 2 days in https://search.svc.ci.openshift.org/?search=usage%3A+kill+&maxAge=48h&context=2&type=all


smarterclayton:
/cherry-pick release-4.3

for bug 1779491

ironcladlou:
Thanks! This does need a bit more work to actually get the change to roll out. Superseded by https://github.com/openshift/cluster-dns-operator/pull/148.

/close

PR #148, opened 2019-12-04
Bug 1779490: Handle missing pids during sidecar teardown

ironcladlou:
Make the sidecar container script tolerate missing pids during cleanup
to prevent spurious errors on shutdown.


ironcladlou:
Need to add diffing logic to the operator to detect/apply container command changes

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -194,6 +194,14 @@ func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
 		changed = true
 	}
 
+	for i, a := range current.Spec.Template.Spec.Containers {
+		b := expected.Spec.Template.Spec.Containers[i]
```
Miciah: It would be prudent to verify that `len(current.Spec.Template.Spec.Containers)` equals `len(expected.Spec.Template.Spec.Containers)`.

ironcladlou: Good catch — I updated the logic and expanded the test cases

Miciah: The additional test case checks for a  change in the number of commands, not a change in the number of containers.  Did you mean to test the latter?

ironcladlou: The new command diffing and the existing image diffing are against the contents of container arrays which are assumed to be the same structurally.

Seems like we would want a separate check entirely to detect changes to length of containers and if so just take the expected containers and don't bother diffing the contents of the container arrays — agree?

Miciah: I suggested comparing lengths of the container arrays because you were using the iterator of one container array to index the other container array.  Detecting changes to the length of the containers arrays would be orthogonal but for that—we don't want the operator to panic if a naughty administrator or a future version of the operator changes the number of containers.

Miciah: I agree that we can just copy the containers array wholesale if any change is detected.

ironcladlou: >I suggested comparing lengths of the container arrays because you were using the iterator of one container array to index the other container array. Detecting changes to the length of the containers arrays would be orthogonal but for that—we don't want the operator to panic if a naughty administrator or a future version of the operator changes the number of containers.

That was my understanding as well, and I implemented that guard. Then you said

>The additional test case checks for a change in the number of words in the command, not a change in the number of containers. Did you mean to test the latter?

Which led to the discussion about the containers array itself (not the command array), which you now seem to say is orthogonal (which I would agree with).

I think the current patch is correct — agree?

Miciah: > Which led to the discussion about the containers array itself (not the command array), which you now seem to say is orthogonal (which I would agree with).
> 
> I think the current patch is correct — agree?

It is correct.  I asked about the intention of the additional test case because it seemed unnecessary to test both that `daemonsetConfigChanged` detects a change to the command and a that `daemonsetConfigChanged` detects a change to the length of the command when we're just using `cmp.Equal` to compare the command arrays, and since you were adding logic to detect changes to the length of the containers array, I thought maybe your intention was to add a test case for that.  If not, never mind.  We've probably spent more time discussion this matter than it deserves—sorry for the distraction!

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -194,6 +194,21 @@ func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
 		changed = true
 	}
 
+	// Detect changes to container commands
```
smarterclayton: The fact you are having to write this code is terrifying.  Is there a library-go function you should be using instead?

Miciah: Looks like library-go's resourceapply package might work.  However, the prospect of using library-go presents at least two salient challenges for us: (1) library-go is poorly documented and (2) library-go uses client-go whereas the DNS operator uses the controller-runtime client library.  Given these considerations, I don't see why using library-go would be less terrifying.

PR #150, opened 2019-12-05
[release-4.2] Bug 1780213: Don't start DNS on NotReady nodes

ironcladlou:
Relax the tolerations for CoreDNS so that it doesn't get started on
nodes that aren't yet ready.


ironcladlou:
@knobunc @smarterclayton I think we'll need this to help stabilize the 4.2 -> 4.3 upgrade path (https://bugzilla.redhat.com/show_bug.cgi?id=1778954)

PR #151, opened 2019-12-13
Node resolver: support IPv6 service IPs

ironcladlou:
This is a quick fix to support IPv6 service IPs when processing services
for /etc/hosts on the node. If a service exists but no A record is found,
check for an AAAA record, and if found, use that instead.

The assumption is that no A record will exist in the IPv6 environment,
which seems to hold for now.

This should enable the container runtime to resolve internal image
registry names for image pulls.


russellb:
seems OK to me

ironcladlou:
In a followup sometime during the release it might be nice to port this function into a subcommand of the Go binary, it's beyond hairy at this point.

ironcladlou:
Result on a node:

```
sh-4.2# cat /etc/hosts
# Kubernetes-managed hosts file (host network).
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
fd02::d2c8 image-registry.openshift-image-registry.svc image-registry.openshift-image-registry.svc.cluster.local # openshift-generated-node-resolver
```

assets/dns/daemonset.yaml
```diff
@@ -90,10 +90,16 @@ spec:
             declare -A svc_ips
             for svc in "${services[@]}"; do
               # Fetch service IP from cluster dns if present
-              ips=($(dig @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"))
+              ips=($(dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"))
```
Miciah: How about `dig "@$NAMESERVER" +short A "${svc}.${CLUSTER_DOMAIN}" AAAA "${svc}.${CLUSTER_DOMAIN}"`?

Miciah: Although that prints an extra line: ";; Warning, extra type option", which we would need to ignore (or maybe we shouldn't ignore it...).

Miciah: This should avoid the warning: `dig "@$NAMESERVER" +short "${svc}.${CLUSTER_DOMAIN}" AAAA "${svc}.${CLUSTER_DOMAIN}"`.

ironcladlou: Refactor the whole thing to Go in a followup?

ironcladlou:
I think something's up with the patch on AWS — seeing this in the node resolver container logs:

```
cmp: EOF on /etc/hosts
```

(That's the entirety of the logs.)

Miciah:
> I think something's up with the patch on AWS — seeing this in the node resolver container logs:
> 
> ```
> cmp: EOF on /etc/hosts
> ```
> 
> (That's the entirety of the logs.)

That suggests that `/etc/hosts` was initially empty, the script updated it, and no further changes were needed.  How about catting `/etc/hosts` before the `while` loop so we can see whether that is the case?  It would also make the output clearer to change `cmp` to `diff -u` (the exit codes are the same, but `diff` additionally gives output).

Another possibility is that something else is writing to `/etc/hosts.tmp` and is racing with the script and sometimes truncating `/etc/hosts.tmp` between when the script writes to it and when the script copies it over `/etc/hosts`.  This seems unlikely, and catting `/etc/hosts` on startup would help us to rule that possibility out.

ironcladlou:
Now I think we might be experiencing collateral damage from other CI issues. Need to go over the latest build cop communications, etc.

ironcladlou:
https://search.svc.ci.openshift.org/?search=failed.*proper+ImagePullPolicy&maxAge=24h&context=2&type=all

https://search.svc.ci.openshift.org/?search=failed.*58+characters&maxAge=24h&context=2&type=all

https://search.svc.ci.openshift.org/?search=failed.*Service+Catalog+should+check+basic+usages&maxAge=24h&context=2&type=all

https://search.svc.ci.openshift.org/?search=failed.*capture+build+stages+and+durations&maxAge=24h&context=2&type=all

https://search.svc.ci.openshift.org/?search=failed.*deploymentconfigs+keep+the+deployer+pod+invariant+valid&maxAge=24h&context=2&type=all

I don't think this patch is the cause of the aws-e2e failures.

russellb:
/cherry-pick release-4.3

@ironcladlou FYI

PR #153, opened 2020-01-30
Bug 1786024: Try TCP DNS in dns-node-resolver

dulek:
OpenShift deployments configured with Kuryr on older OpenStack
deployments are not exposing UDP in DNS Service as older OpenStack
Octavia does not support UDP listeners. This means that DNS queries need
to go over TCP. This commit makes dns-node-resolver retry with TCP DNS
queries when both UDP ones fail.

This can be removed once we stop supporting OpenStack Queens (OSP13).


assets/dns/daemonset.yaml
```diff
@@ -89,18 +89,23 @@ spec:
...
-              # Fetch service IP from cluster dns if present
-              ips=($(dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"))
+              # Fetch service IP from cluster dns if present. We make several tries
+              # to do it: IPv4, IPv6, IPv4 over TCP and IPv6 over TCP. The two last ones
+              # are for deployments with Kuryr on older OpenStack (OSP13) - those do not
+              # support UDP loadbalancers and require reaching DNS through TCP.
+              cmds=("dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}""
+                    "dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}""
+                    "dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}""
+                    "dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"")
```
Miciah: The quoting is getting weird—it looks like your intention is to have nested quoting, but you actually are removing the quoting around the parameter expansions.  I'd suggest either using single-quoting around the command and using `eval`:
```bash
    cmds=('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"')
    declare -p cmds
    for i in ${!cmds[*]}
    do
        ips=($(eval "${cmds[i]}"))
        if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
            svc_ips["${svc}"]="${ips[@]}"
            break
        fi
    done
    # ...
```
or using nested loops:
```bash
    for record_type in 'A' 'AAAA'; do
        for proto in '+notcp' '+tcp'; do
            ips=($(dig -t "$record_type" "@$NAMESERVER" "$proto" +short "${svc}.${CLUSTER_DOMAIN}"))
            if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
                svc_ips["${svc}"]="${ips[@]}"
                break 2
            fi
        done
    done
    # ...
```

dulek: I went with the former idea. Thanks for the code that works out of the box, that's a really wholesome review. :)

assets/dns/daemonset.yaml
```diff
@@ -89,18 +89,23 @@ spec:
...
-                ips=($(dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"))
+              # Fetch service IP from cluster dns if present. We make several tries
+              # to do it: IPv4, IPv6, IPv4 over TCP and IPv6 over TCP. The two last ones
+              # are for deployments with Kuryr on older OpenStack (OSP13) - those do not
+              # support UDP loadbalancers and require reaching DNS through TCP.
+              cmds=('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+                    'dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+                    'dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+                    'dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"')
+              declare -p cmds
```
gryf: Do we need this? Or maybe it's here for debugging purposes?

assets/dns/daemonset.yaml
```diff
@@ -89,18 +89,23 @@ spec:
...
+              for i in ${!cmds[*]}
```
gryf: Why not just
```bash
for cmd in "${cmds[@]}"
```
and use the command directly?

assets/dns/daemonset.yaml
```diff
@@ -89,18 +89,23 @@ spec:
...
+              do
+                ips=($(eval "${cmds[i]}"))
```
gryf: Why not just
```bash
ips=($("${cmd}"))
```
?

assets/dns/daemonset.yaml
dulek: @Miciah added it and yeah - I'd say it's useful to have it in the logs of the container.

assets/dns/daemonset.yaml
dulek: Right, I had some issues with that, but probably due to wrong escaping.

assets/dns/daemonset.yaml
dulek: It won't resolve vars:

```
[mdulko@dulek-workstation kuryr-kubernetes]$ for cmd in "${cmds[@]}"; do $("${cmd}"); done
bash: dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}": command not found...
bash: dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}": command not found...
bash: dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}": command not found...
bash: dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}": command not found...
[mdulko@dulek-workstation kuryr-kubernetes]$ for cmd in "${cmds[@]}"; do $(${cmd}); done
dig: couldn't get address for '"${NAMESERVER}"': not found
dig: couldn't get address for '"${NAMESERVER}"': not found
dig: couldn't get address for '"${NAMESERVER}"': not found
dig: couldn't get address for '"${NAMESERVER}"': not found
```

assets/dns/daemonset.yaml
dulek: Ah no, doing it the way you specify I'm getting elements of the array split by spaces. It needs to be like this.

```
[mdulko@dulek-workstation ~]$ for svc in "${services[@]}"; do
>   # Fetch service IP from cluster dns if present. We make several tries
>   # to do it: IPv4, IPv6, IPv4 over TCP and IPv6 over TCP. The two last ones
>   # are for deployments with Kuryr on older OpenStack (OSP13) - those do not
>   # support UDP loadbalancers and require reaching DNS through TCP.
>   cmds=('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
>         'dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
>         'dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
>         'dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"')
>   declare -p cmds
>   for cmd in ${cmds[@]}
>   do
>     echo $cmd
>   done
> done
declare -a cmds=([0]="dig -t A @\"\${NAMESERVER}\" +short \"\${svc}.\${CLUSTER_DOMAIN}\"" [1]="dig -t AAAA @\"\${NAMESERVER}\" +short \"\${svc}.\${CLUSTER_DOMAIN}\"" [2]="dig -t A +tcp @\"\${NAMESERVER}\" +short \"\${svc}.\${CLUSTER_DOMAIN}\"" [3]="dig -t AAAA +tcp @\"\${NAMESERVER}\" +short \"\${svc}.\${CLUSTER_DOMAIN}\"")
dig
-t
A
@"${NAMESERVER}"
+short
"${svc}.${CLUSTER_DOMAIN}"
dig
-t
AAAA
@"${NAMESERVER}"
+short
"${svc}.${CLUSTER_DOMAIN}"
dig
-t
A
+tcp
@"${NAMESERVER}"
+short
"${svc}.${CLUSTER_DOMAIN}"
dig
-t
AAAA
+tcp
@"${NAMESERVER}"
+short
"${svc}.${CLUSTER_DOMAIN}"

```

ironcladlou: I wonder if it would have taken less time to introduce a new command to the Go binary...

gryf: Ack, although I bet, there is a better way for doing this.

assets/dns/daemonset.yaml
gryf: That's because you are using single quotes in `cmds` array :) The can be unfloded during array declaration. I.e.
```bash
cmds = ('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
...
```
can be replaced by
```bash
cmds = ("dig -t A @${NAMESERVER} +short ${svc}.${CLUSTER_DOMAIN}"
...
```

dulek: Seems like it works, but only if I use `ips=($(${cmd}))` (no quotes). I can change it though I'm not sure why it was escaped like this in the first place.

dulek:
Alright, I just tested this with changes requested by @gryf and it seems to be working.

assets/dns/daemonset.yaml
Miciah: Sorry, it should be omitted.

assets/dns/daemonset.yaml
Miciah: I prefer to leave the quoting around the parameter expansions inside the commands.

dulek:
@Miciah: Okay, I reverted this back to version using `eval` and removed the debug bit.

assets/dns/daemonset.yaml
Miciah: If you want to get rid of the subscripting, the following should work:
```bash
    cmds=('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
          'dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"')
    for cmd in "${cmds[@]}"
    do
        ips=($(eval "$cmd"))
        # ...
```

assets/dns/daemonset.yaml
gryf: This is what I've meant:

```bash
    cmds=("dig -t A @${NAMESERVER} +short ${svc}.${CLUSTER_DOMAIN}"
        "dig -t AAAA @${NAMESERVER} +short ${svc}.${CLUSTER_DOMAIN}"
        "dig -t A +tcp @${NAMESERVER} +short ${svc}.${CLUSTER_DOMAIN}"
        "dig -t AAAA +tcp @${NAMESERVER} +short ${svc}.${CLUSTER_DOMAIN}")
    for cmd in "${cmds[@]}"
    do
        if ! mapfile -t ips < <(${cmd})
        then
            continue
        fi
        if [[ "${#ips[@]}" -ne 0 ]]; then
            svc_ips["${svc}"]="${ips[*]}"
            break
        fi
    done
    # …
done
```
By forcing of using apostrophes, one need to use `eval`, which seems unnecessary. Also note addresses concatenation in ` svc_ips["${svc}"]="${ips[*]}"` line.

Miciah: I like the use of `mapfile`!

> By forcing of using apostrophes, one need to use `eval`, which seems unnecessary.

My objection is that omitting double-quotes around `${NAMESERVER}` and `${svc}.${CLUSTER_DOMAIN}` tells Bash to perform word splitting after parameter expansion, which is semantically incorrect and gratuitously trusts that the input has the expected form (namely, that it does not have whitespace), which increases the probability that the code will be misunderstood, broken during refactoring, or used as a gadget in some larger security exploit (I cannot imagine how injecting additional parameters into the `dig` command could be used in an exploit, but avoiding word splitting would eliminate the possibility entirely).  Proper quoting avoids word splitting.  Using `eval` is ugly but gives us correct semantics.  (For what it's worth, using a nested loop, as in my second suggestion, avoids both `eval` and gratuitous word-splitting.)

> Also note addresses concatenation in ` svc_ips["${svc}"]="${ips[*]}"` line.

Agreed, `"${ips[*]}"` is clearer (I didn't say anything before because the existing code uses `"${ips[@]}"`, and the assignment forces scalar context, in which the two are equivalent).

gryf: > My objection is that omitting double-quotes around `${NAMESERVER}` and `${svc}.${CLUSTER_DOMAIN}` tells Bash to perform word splitting after parameter expansion, which is semantically incorrect and gratuitously trusts that the input has the expected form (namely, that it does not have whitespace), which increases the probability that the code will be misunderstood, broken during refactoring, or used as a gadget in some larger security exploit (I cannot imagine how injecting additional parameters into the `dig` command could be used in an exploit, but avoiding word splitting would eliminate the possibility entirely). […]

I see the issue now. So the only way is either use the `eval` or nested loops. I tend to like the latter :). Modified your example a little:

```bash
for svc in "${services[@]}"
do
    for record_type in "A" "AAAA"
    do
        for proto in "+notcp" "+tcp"
        do
            if ! mapfile -t ips < <(dig -t "${record_type}" "@${NAMESERVER}" \
                ${proto} +short "${svc}.${CLUSTER_DOMAIN}")
            then
                continue
            fi
            if [[ "${#ips[@]}" -ne 0 ]]; then
                svc_ips["${svc}"]="${ips[*]}"
                break 2
            fi
        done
    done
    #…
```

> Agreed, `"${ips[*]}"` is clearer (I didn't say anything before because the existing code uses `"${ips[@]}"`, and the assignment forces scalar context, in which the two are equivalent).

+1

dulek: `mapfile` won't work as apparently `dig` has a curious way of signaling errors:

```
[root@dns-default-4nvth /]# echo $SERVICES
image-registry.openshift-image-registry.svc
[root@dns-default-4nvth /]#           IFS=', ' read -r -a services <<< "${SERVICES}"
[root@dns-default-4nvth /]#             declare -A svc_ips
[root@dns-default-4nvth /]#             for svc in "${services[@]}"
>             do
>               # Fetch service IP from cluster dns if present. We make several tries
>               # to do it: IPv4, IPv4 over TCP, IPv6 and IPv6 over TCP. The TCP ones
>               # are for deployments with Kuryr on older OpenStack (OSP13) - those do not
>               # support UDP loadbalancers and require reaching DNS through TCP.
>               for record_type in "A" "AAAA"
>               do
>                 for proto in "+notcp" "+tcp"
>                 do
>                   if ! mapfile -t ips < <(dig -t "${record_type}" @"${NAMESERVER}" ${proto} +short "${svc}.${CLUSTER_DOMAIN}")
>                   then
>                     continue
>                   fi
>                   if [[ "${#ips[@]}" -ne 0 ]]; then
>                     svc_ips["${svc}"]="${ips[*]}"
>                     break 2
>                   fi
>                 done
>               done
>             done

[root@dns-default-4nvth /]# 
[root@dns-default-4nvth /]# echo ${svc_ips[*]}
;; connection timed out; no servers could be reached
```

Folks, while I can learn a lot from your discussions, we're now in the 2nd day of discussing syntax of a script that should get rewritten to golang anyway. Each time I need to get an env to test it and at the moment we have just one for whole team. Can we just agree on one solution that works? The current way works fine, isn't worse than previous code and doesn't suffer from escaping issue that @Miciah pointed out.

gryf: That's odd, since I did the testing yesterday and today morning. Even with nonexistent domains, which dig couldn't reach. So maybe it's not an issue with `dig` itself, but the nameserver and services?

ironcladlou:
I would have vastly preferred the psychic energy expended here be directed towards a Go implementation...  do we want to chalk this up to a sunk cost and use this bug as the reason to introduce a new command to the image?

assets/dns/daemonset.yaml
gryf: Ok. My bad. I foolishly assumed, that in case of redirection to `mapfile` all the errors from subshell will bubble up.

So in case of phony service *or* unresponsive name server, it should be something like this:

```bash
for svc in "${services[@]}"
do
    for record_type in "A" "AAAA"
    do
        for proto in "+notcp" "+tcp"
        do
            # First 'if' will check command exit code. The error usually means,
            # that we cannot reach name server, i.e.
            # $ dig -t A @192.168.0.255 +short google.com
            # ;; connection timed out; no servers could be reached
            if output=$(dig -t "${record_type}" @"${NAMESERVER}" "${proto}" \
                 +short "${svc}.${CLUSTER_DOMAIN}")
            then
                # This one will check if we have anything from dig. It might
                # happen, that its an empty string if nothing match the name.
                [[ -z "${output}" ]] && continue
                # Here should be some validation of the addresses returned by dig.
                # We naively assume that it should be address(ess)
                svc_ips["${svc}"]="${output}"
                break 2
            fi
        done
    done
done

# display svc_ip hashmap
declare -p svc_ips
```

Note, that from the beginning there was no validation at all, that name server is available/reachable, and there is nothing which parse possible non expected dig output.

ironcladlou:
I'm working on rewriting this container script as a new command in the Go project.

ironcladlou:
I'm working on the Go port in https://github.com/openshift/cluster-dns-operator/pull/154.

dulek:
@ironcladlou, @Miciah: Folks, I'd still insist on merging this. We'd like this to get backported to 4.3 as well and whole Dan's refactoring might be too big for a backport.

What do you think?

ironcladlou:
@dulek 

>@ironcladlou, @Miciah: Folks, I'd still insist on merging this. We'd like this to get backported to 4.3 as well and whole Dan's refactoring might be too big for a backport.
>What do you think?

No objection... I'll let @Miciah have final sign-off on this implementation. I wish we had some automated tests around this, but for now I guess we're probably going to have to just rely on manual tests and indirect testing via the image registry e2e.

PR #154, opened 2020-02-05
WIP: Refactor /etc/hosts management into a Go command

ironcladlou:
The Bash implementation of the /etc/hosts manager has started to
become more complex than can be reasoned about and tested easily.

Rewrite it in Go.


cmd/dns-operator/hosts_updater.go
```diff
@@ -0,0 +1,97 @@
...
+				os.Exit(1)
+			}
+		},
+	}
+
+	cmd.Flags().StringVarP(&updater.HostsFile, "hosts-file", "f", "/etc/hosts", "hosts file location")
+	cmd.Flags().StringArrayVarP(&updater.Services, "services", "s", DefaultServices, "service names to resolve and add to hosts file")
+	cmd.Flags().DurationVarP(&updater.Interval, "interval", "i", 1*time.Minute, "how often to sync")
+	cmd.Flags().StringVarP(&updater.Nameserver, "nameserver", "n", "", "the IP of the cluster DNS server (required)")
+	cmd.Flags().StringVarP(&updater.ClusterDomain, "cluster-domain", "d", "cluster.local", "the IP of the cluster DNS server (required)")
```
Miciah: The description needs to be fixed.

cmd/dns-operator/hosts_updater.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	// Read the existing hosts file, excluding lines previously generated
+	// from the hosts updater, so we can re-generate our entries.
+	file, err := ioutil.ReadFile(s.HostsFile)
+	if err != nil {
+		return err
+	}
+	var newFile string
```
Miciah: Use a buffered writer?

ironcladlou:
@Miciah @dulek @gryf this is getting very close. I'd like to clarify something about the tcp use case. Instead of blindly trying tcp as a fallback, is there some cluster state we could examine which would inform the update logic as to which protocol to use?

cmd/dns-operator/hosts_updater.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+const marker = "openshift-generated-node-resolver"
+
+func (s *HostsUpdater) Run() error {
+	s.resolver = &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			d := net.Dialer{}
+			return d.DialContext(ctx, "udp", net.JoinHostPort(s.Nameserver, "53"))
```
dulek: Nice, I'll be able to add TCP here too!

cmd/dns-operator/hosts_updater.go
```diff
@@ -0,0 +1,164 @@
...
+	if err := scanner.Err(); err != nil {
+		return "", err
+	}
+
+	// Append latest entries to the new hosts file contents.
+	var entries []string
+	for _, svc := range s.Services {
+		fqdn := strings.Join([]string{svc, s.ClusterDomain}, ".")
+		ctx, _ := context.WithTimeout(context.Background(), 5*time.Second)
+		ips, err := s.resolver.LookupIPAddr(ctx, fqdn)
```
dulek: Shouldn't this be LookupAddr?

ironcladlou: LookupAddr is for a reverse lookup, and we're building static hosts entries to IP addresses. Maybe I'm misunderstanding your q?

dulek:
> @Miciah @dulek @gryf this is getting very close. I'd like to clarify something about the tcp use case. Instead of blindly trying tcp as a fallback, is there some cluster state we could examine which would inform the update logic as to which protocol to use?

Short version - if you don't want to connect to OpenStack from CDO then no, there's no way as this depends on OpenStack Octavia version.

Long version: There's kinda way, but I don't really want to do this. Also you might want to sit down for this one. So with older Octavia not supporting UDP in its load balancers, the DNS queries in the cluster with Kuryr must go through TCP. And this is not a default DNS resolver behavior, so when Kuryr is deployed there's a mutating webhook that injects `use-vc` into DNS options of each pod.
DNS operator pod should have it too, but… apparently it gets created before the webhook is deployed, so it doesn't have it. We can force deletion of it to make webhook "catch" it as we do that with some other pod, but I'd really like to avoid it.
Also `use-vc` is not honored by golang resolver, so you'd need to make sure CGO resolver is used.

cmd/dns-operator/hosts_updater.go
dulek: Sorry, you're totally right, it's me misunderstanding golang's docs.

cmd/dns-operator/hosts_updater.go
```diff
@@ -34,10 +35,10 @@ func NewUpdateHostsCommand() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&updater.HostsFile, "hosts-file", "f", "/etc/hosts", "hosts file location")
-	cmd.Flags().StringArrayVarP(&updater.Services, "services", "s", DefaultServices, "service names to resolve and add to hosts file")
+	cmd.Flags().StringArrayVarP(&updater.Services, "services", "s", defaultServices, "service names to resolve and add to hosts file")
 	cmd.Flags().DurationVarP(&updater.Interval, "interval", "i", 1*time.Minute, "how often to sync")
 	cmd.Flags().StringVarP(&updater.Nameserver, "nameserver", "n", "", "the IP of the cluster DNS server (required)")
-	cmd.Flags().StringVarP(&updater.ClusterDomain, "cluster-domain", "d", "cluster.local", "the IP of the cluster DNS server (required)")
+	cmd.Flags().StringVarP(&updater.ClusterDomain, "cluster-domain", "d", "cluster.local", "the cluster domain suffix (required)")
```
Miciah: Omit " (required)".

cmd/dns-operator/hosts_updater.go
```diff
@@ -96,36 +114,26 @@ func (s *HostsUpdater) sync() error {
...
+	if err := newHostsFile.Close(); err != nil {
 		return err
 	}
 
-	// Append our entries to the hosts file.
-	for _, entry := range entries {
-		newFile += fmt.Sprintln(entry)
+	// Atomically replace the current hosts file with the new file.
+	if err := os.Rename(newHostsFile.Name(), s.HostsFile); err != nil {
+		return err
```
Miciah: Unlink the temporary file?

cmd/dns-operator/hosts_updater.go
```diff
@@ -96,36 +114,26 @@ func (s *HostsUpdater) sync() error {
...
 	}
+	for _, entry := range entries {
+		newContent += entry + "\n"
+	}
 
-	// Read the existing hosts file, excluding lines previously generated
-	// from the hosts updater, so we can re-generate our entries.
-	file, err := ioutil.ReadFile(s.HostsFile)
+	// Write a temporary hosts file with the updated contents.
+	newHostsFile, err := ioutil.TempFile("", "hosts")
```
Miciah: Need to copy the old file's attributes (owener, permissions, and SELinux context).

cmd/dns-operator/hosts_updater.go
ironcladlou: `os.Rename` is implemented with the [RENAMEAT syscall](https://golang.org/src/syscall/syscall_linux.go?s=3819:3874#L143), which I understand to [atomically replace an existing target](https://linux.die.net/man/2/renameat). Does that mean there's nothing to clean up?

cmd/dns-operator/hosts_updater.go
ironcladlou: Good catch...

cmd/dns-operator/hosts_updater.go
Miciah: If it succeeds, there is nothing to clean up.  If you get an error, the temporary file is probably still present.

ironcladlou: Oh! I totally misunderstood what you were suggesting. Agree!

gryf:
Looks good!
/lgtm

ironcladlou:
@Miciah pointed me to https://github.com/moby/moby/issues/17190, which rehashes a lot of the concerns with /etc/hosts file manipulations in this context. Attribute preservation, atomicity, and bind mount concerns are all serious problems which can't be easily solved in a platform agnostic way.

I wonder if instead of this refactor we should be getting out of the /etc/hosts file manipulation business altogether and moving towards something safer like a different/custom NSS module or pointing the hosts to a real DNS server (I'm not sure I remember the issues with pointing the nodes to CoreDNS).

ironcladlou:
cc @knobunc @frobware @danehans

ironcladlou:
I'm no longer convinced this is the right approach and think we should try to eliminate /etc/hosts management entirely.

PR #156, opened 2020-02-20
Bug 1805177: Prefer node-local endpoints

Miciah:
Use [service topology](https://kubernetes.io/docs/concepts/services-networking/service-topology/#prefer-node-local-endpoints) to prefer node-local endpoints for the cluster DNS service.

* `assets/dns/service.yaml`: Specify topologyKeys to prefer node-local endpoints.

----

Service topology was added in Kubernetes 1.17, so this feature can go into 4.5 and be backported to 4.4, if needed, but not to earlier versions.


ironcladlou:
What other implications are there to this change? For example, when is it desirable for a _non-local_ endpoint to be preferred?

Miciah:
> What other implications are there to this change? For example, when is it desirable for a _non-local_ endpoint to be preferred?

If the local one is down (the `"*"` key covers this case) or if the local CoreDNS instance is overloaded (not sure whether this is a realistic scenario).

ironcladlou:
/retitle Prefer node-local endpoint

Service Topology is alpha in 1.17 and it doesn't seem like we'll be enabling the feature gate in the foreseeable future. So this isn't a fix.

PR #159, opened 2020-03-19
Apply changes to service resource

ironcladlou:
Add support for applying changes to the service resource. Roll back 8ae4b02
since service topology feature gate isn't yet enabled, and add TODOs to
re-introduce support when the gate is finally enabled.

Part of, but not a complete fix for https://bugzilla.redhat.com/show_bug.cgi?id=1805177 because the service topology feature gate isn't enabled. Without that gate, service.spec.toplogyKey won't be persisted.


ironcladlou:
cc @Miciah

pkg/operator/controller/controller_dns_service.go
```diff
@@ -21,16 +23,21 @@ func (r *reconciler) ensureDNSService(dns *operatorv1.DNS, clusterIP string, dae
...
-	if err := r.client.Create(context.TODO(), desired); err != nil {
-		return nil, fmt.Errorf("failed to create dns service: %v", err)
+
+	switch {
+	case desired != nil && current == nil:
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return nil, fmt.Errorf("failed to create dns service: %v", err)
+		}
+		logrus.Infof("created dns service: %s/%s", desired.Namespace, desired.Name)
+		return desired, nil
```
Miciah: Could delete this line and fall out to the `return r.currentDNSService(dns)` statement below, which would return the actual API object.  Our `ensure` methods are inconsistent with respect to which value they return though.

pkg/operator/controller/controller_dns_service.go
```diff
@@ -64,3 +71,45 @@ func desiredDNSService(dns *operatorv1.DNS, clusterIP string, daemonsetRef metav
...
+	logrus.Infof("updated dns service: %s/%s", updated.Namespace, updated.Name)
+	return nil
+}
+
+func serviceChanged(current, expected *corev1.Service) (bool, *corev1.Service) {
+	changed := false
+	updated := current.DeepCopy()
+	if !cmp.Equal(current.Spec.TopologyKeys, expected.Spec.TopologyKeys, cmpopts.EquateEmpty()) {
+		updated.Spec.TopologyKeys = expected.Spec.TopologyKeys
+		changed = true
```
Miciah: Is it better to compare fields explicitly, or to use a broad comparison and ignore specific fields?  I have been leaning towards the latter, but the caveat there is that we might get spurious updates if we miss some field that the API sets:  https://github.com/openshift/cluster-ingress-operator/blob/1f3fbf43b4de34f0185271f894372f55252bee43/pkg/operator/controller/ingress/nodeport_service.go#L134

pkg/operator/controller/controller_dns_service.go
ironcladlou: Fixed

pkg/operator/controller/controller_dns_service.go
ironcladlou: Good idea, refactored

ironcladlou:
I still need to do e2e testing of this.

/hold

ironcladlou:
/hold cancel

@Miciah want to go ahead and merge this?

ironcladlou:
/hold

Wait for https://github.com/openshift/cluster-dns-operator/pull/159 and then rebase

Miciah:
Does `spec.topologyKeys` get persisted on update even with the feature gate disabled?  I don't see it in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/159/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-operator/408/artifacts/e2e-aws-operator/services.json, and I six service updates logged in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/159/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-operator/408/artifacts/e2e-aws-operator/pods/openshift-dns-operator_dns-operator-6bf8cf67bd-fblwc_dns-operator.log.  We don't want the operator to keep retrying the update if the desired service has `spec.topologyKeys` and the API does not persist it.

ironcladlou:
It doesn't get persisted — you're right, if we want to keep the general diff/apply logic I should remove serviceTopology entirely from the PR for now.

ironcladlou:
/hold cancel

@Miciah I rebased and updated the goal of this PR. PTAL.

pkg/operator/controller/controller_dns_service_test.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		{
+			description: "if .spec.sessionAffinity changes",
+			mutate: func(service *corev1.Service) {
+				service.Spec.SessionAffinity = corev1.ServiceAffinityClientIP
+			},
+			expect: true,
+		},
+		{
+			description: "if .spec.publishNotReadyAddresses changes",
```
frobware: Maybe add a test for field `Spec.ClusterIP` as we call that out explicitly in `serviceChanged()`?

```go
// Preserve fields that the API, other controllers, or user may have
// modified.
updated.Spec.ClusterIP = current.Spec.ClusterIP
```

ironcladlou: Done

ironcladlou:
>Might be better to split the rollback into a discrete commit, if possible.

Done

PR #160, opened 2020-03-20
daemonset: Decrease default CPU request relative to SDN

smarterclayton:
CPU is a compressible resource and we want to tune relative to
other infrastructure components on nodes. openshift-sdn requests
100m, and in practice uses between 100-300m. openshift-dns requests
50m, and in practice uses about 5m (on similar workloads). Tune
CoreDNS and the sidecar loop to be more proportionate (half of
SDN).

This brings CoreDNS more in line with other node components and
should still ensure it gets proportional time on nodes.


Miciah:
Thanks!

/lgtm

I am not aware of any issue with lowering the requests for CPU; the old request for CoreDNS was chosen based on the value that upstream sets, and I can find no discussion or documentation for how upstream chose them.  (In contrast, the memory request and limit do have an empirical basis that has been discussed upstream.)

PR #162, opened 2020-03-27
Bug 1813479: Tolerate all taints

ddreggors:
This was changed for master only, but when using taints on infra nodes 
as described in opeshift docs, dns pods will no longer schedule in infra 
nodes.

See:
https://bugzilla.redhat.com/show_bug.cgi?id=1780318#

and

https://bugzilla.redhat.com/show_bug.cgi?id=1813479

### DIFF
--- a/assets/dns/daemonset.yaml
+++ b/assets/dns/daemonset.yaml
@@ -147,6 +147,5 @@ spec:
           path: /etc/hosts
           type: File
       tolerations:
-      # DNS needs to run everywhere.
-      - key: "node-role.kubernetes.io/master"
-        operator: Exists
+      # DNS needs to run everywhere. Tolerate all taints
+      - operator: Exists


ddreggors:
Also this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1813479

Miciah:
I think we can accept this change.

One consequence of this change is that we may get some transient "Missing CNI default network" errors when a not is booting, and these transient errors may result in spurious Bugzilla reports, but ultimately that is a problem that needs to be fixed in the kubelet or elsewhere, but not in this operator.

Please run `make generate` to update `pkg/manifests/bindata.go`.

/approve
/retitle Bug 1813479: Tolerate all taints

Miciah:
Also, please reference https://bugzilla.redhat.com/show_bug.cgi?id=1813479 in the commit message (it currently references a report against MCD).

suridaddy:
@ddreggors will this be delivered to 4.2-release branch?

ddreggors:
> @ddreggors will this be delivered to 4.2-release branch?

Yes please

ironcladlou:
I think we decided we're okay with this, but we'll need to:

1. Be diligent about opening bugs to cover the cases where DNS pods are getting started on nodes where container networking is unavailable
2. For any cases like (1), add some sort of interim status reporting (e.g. event detection?) so the DNS operator can point the finger at the kubelet (for example)

/lgtm

ironcladlou:
@ddreggors please run `make generate` and amend your commit,  generated code is out of date

Miciah:
/hold
`manifests.go` still needs to be regenerated.

We're in a bad state because #167 has merged but https://github.com/openshift/release/pull/8350 has not, which means that CI is not running the verify step, so it is not catching this issue.

Miciah:
/close
Superseded by #171.

danehans:
> One consequence of this change is that we may get some transient "Missing CNI default network" errors when a node is booting, and these transient errors may result in spurious Bugzilla reports, but ultimately that is a problem that needs to be fixed in the kubelet or elsewhere, but not in this operator.

@Miciah FYI... https://bugzilla.redhat.com/show_bug.cgi?id=1821833 has been fixed and a CI search for the error shows no recent results.

PR #122, opened 2019-07-01
NE-76: Integrate operator metrics with Prometheus

Miciah:
The operator exposes metrics, but they are not collected.  Configure Prometheus to collect these metrics.

* `manifests/0000_70_dns-operator_00-namespace.yaml`: Add a label for `openshift-monitoring`.
* `manifests/0000_70_dns-operator_02-deployment.yaml`: Add the metrics port.
* `manifests/0000_70_dns-operator_04-service.yaml`: New file.  Define a metrics service.
* `manifests/0000_90_dns-operator_00_prometheusrole.yaml`:  New file.  Define a role that grants access to services, endpoints, and pods.
* `manifests/0000_90_dns-operator_01_prometheusrolebinding.yaml`:  New file.  Bind the `prometheus-k8s` service account to the new role.
* `manifests/0000_90_dns-operator_02_servicemonitor.yaml`: New file.  Define a service monitor for the new service.

PR #163, opened 2020-04-06
Bug 1809197: Secure CoreDNS metrics

sgreene570:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1809197

Follows along with https://github.com/openshift/cluster-dns-operator/pull/122 to add a `kube-rbac-proxy` sidecar container to the coreDNS daemonset. The ensures that metrics are securely scraped from port 9154, where requests are passed to insecure port 9153 within the pod. 

* `assets/dns/cluster-role.yaml` changes add necessary token cluster permissions for `kube-rbac-proxy`
* `assets/dns/daemonset.yaml` changes add deployment options for `kube-rbac-proxy` 
* `assets/dns/service.yaml` changes add necessary certificate annotation


assets/dns/service.yaml
```diff
@@ -1,5 +1,12 @@
 kind: Service
 apiVersion: v1
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: metrics-tls
+  labels:
+    name: dns
+  name: metrics
+  namespace: openshift-dns
 # name, namespace and labels are set at runtime
```
Miciah, line 10: `desiredDNSDaemonSet` still overrides labels, name, and namespace, so it does not make sense to add them to the yaml file.

cmd/dns-operator/main.go
```diff
@@ -32,10 +32,16 @@ func main() {
 		logrus.Fatalf("OPENSHIFT_CLI_IMAGE environment variable is required")
 	}
 
+	kubeRBACProxyImage := os.Getenv("KUBE_RBAC_PROXY_IMAGE")
+	if len(cliImage) == 0 {
```
Miciah, line 36: `cliImage` → `kubeRBACProxyImage`.

pkg/operator/config/config.go
```diff
@@ -11,4 +11,7 @@ type Config struct {
 
 	// OpenshiftCLIImage is the openshift client image to manage.
 	OpenshiftCLIImage string
+
+	// KubeRBACProxyImage is the kube-rbac-proxy image to manage.
```
Miciah, line 15: How about this?
```suggestion
	// KubeRBACProxyImage is the kube-rbac-proxy image to use to secure
	// the metrics endpoint.
```

assets/dns/service.yaml
sgreene570: fixed.

pkg/operator/config/config.go
sgreene570: good idea. Changed.

sgreene570:
Verified working with a 4.5 cluster.

Screenshot

assets/dns/service.yaml
```diff
@@ -1,5 +1,8 @@
 kind: Service
 apiVersion: v1
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: metrics-tls
```
Miciah, line 5: It seems strange that the name of the service associated with the `dns` resource is set at runtime, but the name of the associated secret is static.  Granted, in practice, the `dns` resource should be a singleton (the operator processes the `dnses.operator.openshift.io/default` object and ignores any other `dnses.operator.openshift.io` objects), and the singleton's name is always going to be the same (`default`), but as long as the operator is deriving the service's name from the dns's name, it seems like the secret's name should likewise be derived from the dns's name.  What do you think?

manifests/0000_70_dns-operator_00-cluster-role.yaml
```diff
@@ -72,6 +73,7 @@ rules:
   - get
   - list
   - watch
+  - update
```
Miciah, line 76: We really only need to add update permission for the `clusterroles` resource, right?  The more we can restrict access, the better.

pkg/operator/controller/controller_cluster_role_test.go
```diff
@@ -0,0 +1,49 @@
...
+	testCases := []struct {
+		description string
+		mutate      func(*rbacv1.ClusterRole)
+		expect      bool
+	}{
+		{
+			description: "if nothing changes",
+			mutate:      func(_ *rbacv1.ClusterRole) {},
+			expect:      false,
+		},
```
Miciah, line 20: Maybe add a test case where something that should be ignored changes, such as adding an annotation or label.

pkg/operator/controller/controller_dns_daemonset_test.go
```diff
@@ -104,6 +113,13 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if .spec.template.spec.volumes changes",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				daemonset.Spec.Template.Spec.Volumes = []corev1.Volume{volume}
```
Miciah, line 119: Is it necessary to use a named value rather than a literal?  If you prefer not to use a literal for stylistic reasons, consider moving the declaration of `volume` here as this is the only place that uses it.

pkg/operator/controller/controller_dns_service.go
```diff
@@ -93,12 +93,19 @@ func serviceChanged(current, expected *corev1.Service) (bool, *corev1.Service) {
 		cmpopts.IgnoreFields(corev1.ServiceSpec{}, "ClusterIP", "TopologyKeys"),
 		cmpopts.EquateEmpty(),
 	}
-	if cmp.Equal(current.Spec, expected.Spec, serviceCmpOpts...) {
+
+	servingCertAnnotationKey := "service.beta.openshift.io/serving-cert-secret-name"
+	currentServingCertAnnotation := current.ObjectMeta.Annotations[servingCertAnnotationKey]
+	expectedServingCertAnnotation := expected.ObjectMeta.Annotations[servingCertAnnotationKey]
+	annotationMatches := cmp.Equal(currentServingCertAnnotation, expectedServingCertAnnotation)
```
Miciah, line 100: As you are comparing just the single annotation's value, `cmp.Equal` should be unnecessary; `annotationMatches := currentServingCertAnnotation == expectedServingCertAnnotation` should suffice.

pkg/operator/controller/controller_service_monitor_test.go
```diff
@@ -0,0 +1,64 @@
...
+	testCases := []struct {
+		description string
+		mutate      func(*unstructured.Unstructured)
+		expect      bool
+	}{
+		{
+			description: "if nothing changes",
+			mutate:      func(_ *unstructured.Unstructured) {},
+			expect:      false,
+		},
```
Miciah, line 19: Maybe add a test case where something that should be ignored (such as a metadata field) is ignored.

pkg/operator/controller/controller_service_monitor_test.go
```diff
@@ -0,0 +1,64 @@
...
+		{
+			description: "if spec.endpoints.scheme changes",
+			mutate: func(serviceMonitor *unstructured.Unstructured) {
+				serviceMonitor.Object["spec"] = map[string]interface{}{
+					"selector": map[string]interface{}{},
+					"endpoints": []interface{}{
+						map[string]interface{}{
+							"scheme": "http",
```
Miciah, line 27: Instead of overriding `spec`, is it feasible to update just this field?  If it would be a hassle to do so, this is fine, but it does mean that we risk false positives if anything else in the definition of `original` below differs from the literal value here.

pkg/operator/controller/controller_cluster_role_test.go
sgreene570: done.

manifests/0000_70_dns-operator_00-cluster-role.yaml
sgreene570: done

pkg/operator/controller/controller_service_monitor_test.go
sgreene570: This is cumbersome, although I do see the risk of false positives. Maybe this is something that could be corrected with the potential switch to the prometheus operator's servicemonitor type?

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -84,6 +84,11 @@ func desiredDNSDaemonSet(dns *operatorv1.DNS, clusterIP, clusterDomain, coreDNSI
 			coreFileVolumeFound = true
 			break
 		}
+		if daemonset.Spec.Template.Spec.Volumes[i].Name == "metrics-tls" {
```
Miciah, line 87: Could replace these two `if` statements with a `switch daemonset.Spec.Template.Spec.Volumes[i].Name`.

Miciah:
/lgtm
I commented with one minor suggestion, but this is fine to merge as is.
/hold
Feel free to `/hold cancel` when you are ready.

ironcladlou:
Sorry, I totally missed this one. I think @smarterclayton was surprised we were still using the rbac proxy, but I can't remember what the suggested alternative was. Should we make sure the proxy is still the best approach?

Miciah:
Apparently oauth-proxy provides the same functionality (it isn't just for OAuth!).  We're still using kube-rbac-proxy for the operator metrics and for cluster-ingress-operator, so I don't see any urgency to switching to oauth-proxy, but it would be a good follow-up to switch both operators and the CoreDNS operand to use oauth-proxy.

danehans:
Since we did not hear back from @smarterclayton in https://github.com/openshift/cluster-dns-operator/pull/122 and the operator already uses kube-rbac-proxy, I think it's acceptable to proceed in this direction.

/lgtm

PR #166, opened 2020-04-15
Bug 1835725: computeOperatorDegradedCondition: Fix "NoDNS" check

Miciah:
Follow-up to #109.

* `pkg/operator/controller/status.go` (`computeOperatorDegradedCondition`): Check total number of `dns` resources, not the available number, before reporting no DNS resource exists.

---

I noticed the erroneous error reporting in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_router/113/pull-ci-openshift-router-master-e2e-aws/373:

```
level=error msg="Cluster operator dns Degraded is True with NoDNS: No DNS resource exists"
```

The operator's logs showed that it had created the resource, and CI artifacts showed that the daemonset existed but had no available replicas.


danehans:
`TestDNSForwarding` is the source of the `e2e-aws-operator` failure. Maybe the wait for the Corefile to get updated in each DNS pod should be greater than 2 minutes?

/test e2e-aws-operator

PR #170, opened 2020-04-22
Bug 1826885: DNS Service Monitor & Daemonset fixes

sgreene570:
Fixes the potential for a false positive unit test in `pkg/operator/controller/controller_service_monitor_test.go`.
Also adds the `kube-rbac-proxy` name to the DNS daemonset config changed function, as well as corresponding unit test.


sgreene570:
/retitle Bug 1826885 DNS Service Monitor & Daemonset fixes

Miciah:
Thanks! Sorry I didn't catch this when reviewing #163.  Failing to update the kube-rbac-proxy image would be bad, especially if there were a security update to the kube-rbac-proxy image in 4.5.z.

/lgtm

PR #171, opened 2020-04-25
Bug 1813479: Tolerate all taints

Miciah:
Configure the operand daemonset to tolerate all taints in order to accommodate user-set taints.

This PR reverts #140.

* `assets/dns/daemonset.yaml`: Tolerate all taints.

----

Supersedes #162.


smarterclayton:
/hold

@derekwaynecarr did we get agreement to allow tolerating all taints

Is there an enhancement / update that clarifies official stance of project on this?

runcom:
> /hold
> 
> @derekwaynecarr did we get agreement to allow tolerating all taints
> 
> Is there an enhancement / update that clarifies official stance of project on this?

sorry to introduce MCO comments here, but we have the very same BZ for the MCD where we're asked to tolerate all taints and we have a PR to fix it as well. Tho, I'm definitely not sure what's the discussion here is about wrt to other system daemons. Is it always ok for things like MCD/DNS to tolerate all taints?

Miciah:
Sorry, I should have posted an update to this PR.  Currently we are building consensus and codifying the OpenShift project's conventions around taints and tolerations:  https://github.com/openshift/enhancements/pull/344

However, because we have not gotten these conventions codified yet and we have code freeze for the 4.5 release this Friday, we have deferred this fix to 4.6 with plans to backport the fix to 4.5.z (https://bugzilla.redhat.com/show_bug.cgi?id=1813479#c13).  @runcom, maybe MCO can take the same course of action (defer to 4.6 and backport to 4.5.z).

runcom:
> @runcom, maybe MCO can take the same course of action (defer to 4.6 and backport to 4.5.z).

@Miciah thanks a lot for the update and yeah, I've deferred the BZ already to 4.6 - will keep an eye on the enhancement

derekwaynecarr:
@smarterclayton see conventions update:
https://github.com/openshift/enhancements/pull/344

davidkarlsen:
It's un-cool to need to run dns-pods on all nodes. This makes it difficult when you have a zoned environment (different nodes in different VLANs), as requests (via the service) might be attempted towards a pod the node should not be able to communicate with. Also - running DNS everywhere is overkill - it should be sufficient to place on a few selected nodes with a shared-infra purpose for the entire cluster. With the default toleration which effectively ignores all taints this is not possible.

PR #172, opened 2020-05-11
NE-169: Add simple CoreDNS Prometheus alert rules

sgreene570:
Adds a couple of CoreDNS alert rules.

Still deciding if using the CVO is the best way to apply the `PrometheusRule` object. For now I think it is. If anyone has any input for CVO vs ~OLM~ using the DNS operator for managing DNS alert rules, please feel free to chime in. 

- CoreDNSPanicCount
Triggers if CoreDNS panics
- CoreDNSHealthCheckSlow
Triggers if CoreDNS health checks are significantly slowing down

### DIFF
--- a/manifests/0000_90_dns-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_dns-operator_03_prometheusrules.yaml
@@ -1,0 +1,24 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: dns
+  namespace: openshift-dns-operator
+  labels:
+    role: alert-rules
+spec:
+  groups:
+    - name: openshift-dns.rules
+      rules:
+      - alert: CoreDNSPanicking
+        expr: coredns_panic_count_total > 0
+        labels:
+          severity: warning
+        annotations:
+          message: "{{ $value }} CoreDNS panics observed on {{ $labels.instance }}"
+      - alert: CoreDNSHealthCheckSlow
+        expr: histogram_quantile(.95, sum(rate(coredns_health_request_duration_seconds_bucket[5m])) by (instance, le)) > 10
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          message: "CoreDNS Health Checks are slowing down (instance {{ $labels.instance }})"


Miciah:
> If anyone has any input for CVO vs OLM for managing DNS alert rules, please feel free to chime in.

We don't use OLM, so it is a choice between CVO or the operator itself.  Using CVO makes sense to me.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: dns
+  namespace: openshift-dns-operator
+  labels:
+    role: alert-rules
+spec:
+  groups:
+    - name: openshift-dns.rules
+      rules:
+      - alert: CoreDNSPanicCount
```
Miciah, line 12: &lt;bikeshed&gt;Maybe `CoreDNSPanicking`?&lt;/bikeshed&gt;

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -0,0 +1,27 @@
...
+        expr: coredns_panic_count_total > 0
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          message: |
+            "{{ $value }} CoreDNS panics encountered on {{ $labels.instance }}"
```
Miciah, line 19: &lt;bikeshed&gt;"encountered" → "observed"&lt;/bikeshed&gt;

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: openshift-dns-operator
+  labels:
+    role: alert-rules
+spec:
+  groups:
+    - name: openshift-dns.rules
+      rules:
+      - alert: CoreDNSPanicCount
+        expr: coredns_panic_count_total > 0
+        for: 5m
```
Miciah, line 14: Why 5m?  Would it make more sense to report a panic immediately?

manifests/0000_90_dns-operator_03_prometheusrules.yaml
Miciah, line 19: Using `""` with `|` means you're quoting a quoting string; is that necessary?

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -0,0 +1,27 @@
...
+        labels:
+          severity: warning
+        annotations:
+          message: |
+            "{{ $value }} CoreDNS panics encountered on {{ $labels.instance }}"
+      - alert: CoreDNSHealthCheckSlow
+        expr: histogram_quantile(.95, sum(rate(coredns_health_request_duration_seconds_bucket[5m])) by (le)) > 10
```
Miciah, line 21: Is this saying that &ge;95% of health probes take &gt; 10 seconds?  Would you mind breaking down and explaining the expression?

manifests/0000_90_dns-operator_03_prometheusrules.yaml
sgreene570: Sounds better, will switch.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
sgreene570: Meant to remove this line, good catch! You're correct, there's no reason to wait on this one.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
sgreene570: It is not, will fix.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
sgreene570: Saw a similar expression in the cluster network operator alerts (https://github.com/openshift/cluster-network-operator/blob/e86bdfc82c02bdd8d6ad04d9cdd28470c49ed4fe/bindata/network/openshift-sdn/alert-rules.yaml#L53).
This `histogram_quantile` function is needed for histogram-based metrics. This expression determines the average time taken per health check request for all requests within the 95th percentile and below. Dropping the top 5% helps to remove some noise, but I suppose is not necessary. We could use 100th or 99th percentile instead if you'd prefer.

Miciah: Is this per-instance?  So if 5 out of 6 instances are healthy and 1 instance is taking &gt;10 seconds for &ge;95% of its health probes, then that raises an alert, right?

Miciah:
I think alerts should focus on infrastructure components that are broken (that is, things in namespaces that start with `openshift-`), or things that could break applications, but not applications themselves.  Some cluster administrators will be very annoyed if they are alerted because some user's throwaway test application is down, and we don't have a way to determine which routes belong to production workloads that the cluster administrator cares about and which do not, so I think we need to leave it to the user to configure alerts for non-infrastructure components.  Let's get the rest of the team's (and possibly some other teams') thoughts on this.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
sgreene570: Ah, I made a mistake. The sum should be `by (instance, le)` not just `by (le)`.

sgreene570: But yes, the query is per-instance.

sgreene570:
> I think alerts should focus on infrastructure components that are broken (that is, things in namespaces that start with `openshift-`), or things that could break applications, but not applications themselves. Some cluster administrators will be very annoyed if they are alerted because some user's throwaway test application is down, and we don't have a way to determine which routes belong to production workloads that the cluster administrator cares about and which do not.

I agree with this ideology. However, in the case of DNS, if a user's application is causing CoreDNS services to panic or otherwise degrade, wouldn't other (infra or user) applications be affected in some capacity? Unless I'm missing something, I don't think these alerts would fire if a user's application were to go down. @Miciah could you maybe elaborate more on what you think DNS alerts should be targeted towards?

Miciah:
I must have mixed up the two PRs.  https://github.com/openshift/cluster-dns-operator/pull/172#issuecomment-629262477 applies to the cluster-ingress-operator PR.  Sorry about that!  This PR looks good.
/lgtm

PR #173, opened 2020-05-20
Bug 1838266: Updates Progressing status condition message

danehans:
The DNS operator `Progressing` status condition message should reflect a "progressing", not an "available" message.


Miciah:
Thanks!
/lgtm

PR #174, opened 2020-06-02
Bug 1842741: Fix serviceChanged and daemonsetConfigChanged

Miciah:
#### `serviceChanged`: Fix session affinity, type

When comparing services to determine whether an update is required, treat implicit and explicit default values as equal.

Before this commit, the update logic would keep trying to revert the default value that the API set for the service's session affinity or type.

* `pkg/operator/controller/controller_dns_service.go` (`serviceChanged`): Use the new `cmpServiceAffinity` and `cmpServiceType` functions so that two services are considered equal if the only difference between them is that one does not specify the session affinity or type and the other service has the default values.
(`cmpServiceAffinity`, `cmpServiceType`): New function.
* `pkg/operator/controller/controller_dns_service_test.go`: (`TestDNSServiceChanged`): Verify that `serviceChanged` returns false if the only mutation is that the session affinity or type has been updated from empty to the default value.

#### `daemonsetConfigChanged`: Fix for volume default mode

When comparing daemonsets to determine whether an update is required, treat implicit and explicit default values as equal.

Before this commit, the update logic would keep trying to revert the default value that the API set for a volume's default mode.

* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Use the new `cmpConfigMapVolumeSource` and `cmpSecretVolumeSource` functions so that two daemonsets are considered equal if the only difference between them is that one does not specify a default mode value and the other daemonset specifies the default value.
(`volumeDefaultMode`): New constant with the default mode that the API uses for configmap and secret volume sources.
(`cmpConfigMapVolumeSource`, `cmpSecretVolumeSource`): New functions.
* `pkg/operator/controller/controller_dns_daemonset_test.go`: (`TestDaemonsetConfigChanged`): Verify that `daemonsetConfigChanged` returns false if the only mutation is that a volume's default mode value has been updated from empty to the default value.

### DIFF
--- a/pkg/operator/controller/controller_dns_daemonset.go
+++ b/pkg/operator/controller/controller_dns_daemonset.go
@@ -207,7 +207,7 @@ func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
 		updated.Spec.Template.Spec.Tolerations = expected.Spec.Template.Spec.Tolerations
 		changed = true
 	}
-	if !cmp.Equal(current.Spec.Template.Spec.Volumes, expected.Spec.Template.Spec.Volumes, cmpopts.EquateEmpty()) {
+	if !cmp.Equal(current.Spec.Template.Spec.Volumes, expected.Spec.Template.Spec.Volumes, cmpopts.EquateEmpty(), cmp.Comparer(cmpConfigMapVolumeSource), cmp.Comparer(cmpSecretVolumeSource)) {
 		updated.Spec.Template.Spec.Volumes = expected.Spec.Template.Spec.Volumes
 		changed = true
 	}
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -230,6 +230,58 @@ func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
...
+	if b.DefaultMode != nil {
+		bDefaultMode = *b.DefaultMode
+	}
+	if aDefaultMode != bDefaultMode {
+		return false
+	}
+	if !cmp.Equal(a.Optional, b.Optional, cmpopts.EquateEmpty()) {
+		return false
+	}
+	return true
```
sgreene570, line 256: nit: could condense lines 253-256 to `return cmp.Equal(a.Optional, b.Optional, cmpopts.EquateEmpty())`

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -230,6 +230,58 @@ func daemonsetConfigChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
...
+// cmpSecretVolumeSource compares two secret volume source values and returns a
+// Boolean indicating whether they are equal.
+func cmpSecretVolumeSource(a, b corev1.SecretVolumeSource) bool {
+	if a.SecretName != b.SecretName {
+		return false
+	}
+	if !cmp.Equal(a.Items, b.Items, cmpopts.EquateEmpty()) {
+		return false
+	}
+	aDefaultMode := int32(420)
```
sgreene570, line 268: Would it make sense to define a `defaultMode := int32(420)` literal somewhere, in the event the default mode were to  change?

Miciah: I think it is extremely unlikely that the default mode would change because that seems like it would be a breaking API change.  I could introduce a constant though.  Would using the same constant for both secret volume sources and configmap volume sources make sense?

pkg/operator/controller/controller_dns_daemonset.go
Miciah: Stylistically, I prefer to follow the pattern consistently:
```go
	if thing1 doesn't match {
		return false
	}
	if thing2 doesn't match {
		return false
	}
	// ...
	if thingn doesn't match {
		return false
	}
	return true
```
There is nothing special about thingn (in concrete terms, the `Optional` field), so I prefer not to deviate from the established pattern.  Does that seem reasonable?

pkg/operator/controller/controller_dns_daemonset.go
sgreene570: Ah, my mistake. I think using a constant as you mentioned would make sense.

pkg/operator/controller/controller_dns_daemonset.go
sgreene570: yes, it does. I see what you mean now!

pkg/operator/controller/controller_dns_daemonset.go
Miciah: Updated.  Thanks!

Miciah:
Rebased.

pkg/operator/controller/controller_dns_daemonset_test.go
```diff
@@ -166,6 +166,22 @@ func TestDaemonsetConfigChanged(t *testing.T) {
...
+			},
+			expect: false,
+		},
+		{
+			description: "if the config-volume default mode value changes",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				newVal := int32(0)
+				daemonset.Spec.Template.Spec.Volumes[1].Secret.DefaultMode = &newVal
+			},
+			expect: true,
```
sgreene570, line 183: These 2 test cases cover the `cmpSecretVolumeSource(...)` function. Perhaps there should be 2 similar test cases for `cmpConfigMapVolumeSource(...)`?

Miciah: Added.  I also noticed that the index was wrong (it should be 0 rather than 1 for the `config-volume` volume), so I fixed the index on the existing test and added a host-path volume to make the list of volumes consistent with the actual manifest.  Thanks!

PR #175, opened 2020-06-08
DNS Daemonset: Add unit test

sgreene570:
Adds a unit test to the DNS daemonset test to cover https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/controller/controller_dns_daemonset.go#L213. Adding a new container to the daemonset is something that has recently occurred (#163), so this test would be helpful.

### DIFF
--- a/pkg/operator/controller/controller_dns_daemonset_test.go
+++ b/pkg/operator/controller/controller_dns_daemonset_test.go
@@ -155,6 +155,17 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if an unexpected additional container is added",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				containers := daemonset.Spec.Template.Spec.Containers
+				daemonset.Spec.Template.Spec.Containers = append(containers, corev1.Container{
+					Name:  "foo",
+					Image: "bar",
+				})
+			},
+			expect: true,
+		},
 	}
 
 	for _, tc := range testCases {


Miciah:
/lgtm
You should be added to https://github.com/openshift/cluster-dns-operator/blob/master/OWNERS; do you want to make a PR to do so?

sgreene570:
> /refresh

Is `/refresh` a valid PR command? Haven't seen that one outside of the context of refreshing a bugzilla.

/retest

Miciah:
> Is `/refresh` a valid PR command? Haven't seen that one outside of the context of refreshing a bugzilla.

Yeah, `/refresh` tells the bot to check CI and update the jobs reported in GitHub.   I believe it's this Prow plugin:  https://godoc.org/k8s.io/test-infra/prow/external-plugins/refresh

The e2e-aws and e2e-aws-operator jobs are still reported to be pending, but it looks like e2e-aws at least actually failed, so I think `/refresh` is supposed to fix that.  However, since `/refresh` didn't do the trick, I'll try `/retest`.

/retest

sgreene570:
> Yeah, `/refresh` tells the bot to check CI and update the jobs reported in GitHub. I believe it's this Prow plugin: https://godoc.org/k8s.io/test-infra/prow/external-plugins/refresh

Ah, ok. I didn't see `/refresh` in https://prow.k8s.io/command-help so I wasn't sure. Thanks!

petr-muller:
`/retest` only kicks-off failed jobs, and `/refresh` makes Prow attempt to re-report finished jobs that failed to be reported for some reason, to the right thing here was to directly kick individual jobs.

PR #176, opened 2020-06-09
controller: Rework boolean logic for consistency

sgreene570:
Switches all `ensureDNS<thing>` functions in `pkg/operator/controller`
to use a `have<thing>` boolean rather than manually checking if
`<thing> != nil`.

@Miciah mentioned that this is the pattern the cluster-ingress-operator tries to follow, so the cluster-dns-operator should do the same.

### DIFF
--- a/pkg/operator/controller/controller_cluster_role.go
+++ b/pkg/operator/controller/controller_cluster_role.go
@@ -15,37 +15,37 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func (r *reconciler) ensureDNSClusterRole() (*rbacv1.ClusterRole, error) {
-	current, err := r.currentDNSClusterRole()
+func (r *reconciler) ensureDNSClusterRole() (bool, *rbacv1.ClusterRole, error) {
+	haveCR, current, err := r.currentDNSClusterRole()
 	if err != nil {
-		return nil, err
+		return false, nil, err
 	}
 	desired := desiredDNSClusterRole()
 
 	switch {
-	case desired != nil && current == nil:
+	case !haveCR:
 		if err := r.client.Create(context.TODO(), desired); err != nil {
-			return nil, fmt.Errorf("failed to create dns cluster role: %v", err)
+			return false, nil, fmt.Errorf("failed to create dns cluster role: %v", err)
 		}
 		logrus.Infof("created dns cluster role: %s/%s", desired.Namespace, desired.Name)
-	case desired != nil && current != nil:
+	case haveCR:
 		if err := r.updateDNSClusterRole(current, desired); err != nil {
-			return nil, err
+			return true, current, err
 		}
 	}
 	return r.currentDNSClusterRole()
 }
 
-func (r *reconciler) currentDNSClusterRole() (*rbacv1.ClusterRole, error) {
+func (r *reconciler) currentDNSClusterRole() (bool, *rbacv1.ClusterRole, error) {
 	current := &rbacv1.ClusterRole{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: manifests.DNSClusterRole().Name}, current)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			return nil, nil
+			return false, nil, nil
 		}
-		return nil, err
+		return false, nil, err
 	}
-	return current, nil
+	return true, current, nil
 }
 
 func desiredDNSClusterRole() *rbacv1.ClusterRole {
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -47,26 +47,26 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
 `))
 
 // ensureDNSConfigMap ensures that a configmap exists for a given DNS.
-func (r *reconciler) ensureDNSConfigMap(dns *operatorv1.DNS, clusterDomain string) (*corev1.ConfigMap, error) {
-	current, err := r.currentDNSConfigMap(dns)
+func (r *reconciler) ensureDNSConfigMap(dns *operatorv1.DNS, clusterDomain string) (bool, *corev1.ConfigMap, error) {
+	haveCM, current, err := r.currentDNSConfigMap(dns)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get configmap: %v", err)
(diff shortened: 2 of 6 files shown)


pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -20,23 +20,23 @@ import (
...
+	haveDS, current, err := r.currentDNSDaemonSet(dns)
 	if err != nil {
-		return nil, fmt.Errorf("failed to build dns daemonset: %v", err)
+		return false, nil, err
 	}
-	current, err := r.currentDNSDaemonSet(dns)
+	desired, err := desiredDNSDaemonSet(dns, clusterIP, clusterDomain, r.CoreDNSImage, r.OpenshiftCLIImage, r.KubeRBACProxyImage)
 	if err != nil {
-		return nil, err
+		return haveDS, nil, fmt.Errorf("failed to build dns daemonset: %v", err)
```
Miciah, line 30: Should return `haveDS, current, fmt.Errorf(...)`.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -20,23 +20,23 @@ import (
...
+	case !haveDS:
 		if err := r.createDNSDaemonSet(desired); err != nil {
-			return nil, err
+			return false, nil, err
 		}
-	case desired != nil && current != nil:
+	case haveDS:
 		if err := r.updateDNSDaemonSet(current, desired); err != nil {
-			return nil, err
+			return true, nil, err
```
Miciah, line 39: Return `true, current, err`.

pkg/operator/controller/controller_dns_service.go
```diff
@@ -18,37 +18,37 @@ import (
...
 		if err := r.client.Create(context.TODO(), desired); err != nil {
-			return nil, fmt.Errorf("failed to create dns service: %v", err)
+			return false, nil, fmt.Errorf("failed to create dns service: %v", err)
 		}
 		logrus.Infof("created dns service: %s/%s", desired.Namespace, desired.Name)
-	case desired != nil && current != nil:
+	case haveService:
 		if err := r.updateDNSService(current, desired); err != nil {
-			return nil, err
+			return true, nil, err
```
Miciah, line 36: Return `true, current, err`.

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -18,23 +18,23 @@ import (
...
 		if err := r.client.Create(context.TODO(), desired); err != nil {
-			return nil, fmt.Errorf("failed to create servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
+			return false, nil, fmt.Errorf("failed to create servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
 		}
 		logrus.Infof("created servicemonitor %s/%s", desired.GetNamespace(), desired.GetName())
-	case desired != nil && current != nil:
+	case haveSM:
 		if err := r.updateDNSServiceMonitor(current, desired); err != nil {
-			return nil, err
+			return true, nil, err
```
Miciah, line 37: Return `true, current, err`.

pkg/operator/controller/controller_cluster_role.go
```diff
@@ -15,37 +15,37 @@ import (
...
-	case desired != nil && current == nil:
+	case !haveCR:
 		if err := r.client.Create(context.TODO(), desired); err != nil {
-			return nil, fmt.Errorf("failed to create dns cluster role: %v", err)
+			return false, nil, fmt.Errorf("failed to create dns cluster role: %v", err)
 		}
 		logrus.Infof("created dns cluster role: %s/%s", desired.Namespace, desired.Name)
-	case desired != nil && current != nil:
+	case haveCR:
 		if err := r.updateDNSClusterRole(current, desired); err != nil {
```
Miciah, line 32: Sidenote:  A future cleanup could add a Boolean return value to the update methods that don't have one to indicate whether an update was performed.  If no update was performed, the caller can return `true, current, nil` instead of falling through to `return r.currentFoo` (which incurs an unnecessary API call).

pkg/operator/controller/controller_cluster_role.go
```diff
@@ -15,37 +15,37 @@ import (
...
-			return nil, err
+			return true, nil, err
```
Miciah, line 33: Should return `true, current, err`.

Miciah, line 33: Should return `true, current, err`.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -47,43 +47,43 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
...
+func (r *reconciler) ensureDNSConfigMap(dns *operatorv1.DNS, clusterDomain string) (bool, *corev1.ConfigMap, error) {
+	haveCM, current, err := r.currentDNSConfigMap(dns)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get configmap: %v", err)
+		return false, nil, fmt.Errorf("failed to get configmap: %v", err)
 	}
 	desired, err := desiredDNSConfigMap(dns, clusterDomain)
 	if err != nil {
-		return nil, fmt.Errorf("failed to build configmap: %v", err)
+		return haveCM, nil, fmt.Errorf("failed to build configmap: %v", err)
```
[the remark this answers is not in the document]
sgreene570: I would agree. Will update this, good catch!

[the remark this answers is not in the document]
sgreene570: Perhaps some of the `return true, nil, ...` lines in the cluster-ingress-operator could be cleaned up as well. I'm working a PR for the cluster-ingress-operator similar to this one, so I'll keep that in mind.

pkg/operator/controller/controller_cluster_role.go
sgreene570: Good idea! I suspect there are quite a few places that have these extra API calls, not just in the cluster-dns-operator but also in the cluster-ingress-operator, right?

Miciah: Probably, yeah.  This is an opportunity to make all of this logic more consistent, correct, and efficient across both operators.

Miciah:
Thanks!
/lgtm

PR #182, opened 2020-07-14
Bug 1813062: ensures service IP is allocatable

danehans:
`pkg/operator/controller/controller.go`: Adds `ensureServiceIP()` to ensure the DNS service IP is available for allocation before assigning it to `spec.ClusterIP`.

/assign @Miciah @knobunc

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -364,6 +364,9 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
 			errs = append(errs, fmt.Errorf("failed to create configmap for dns %s: %v", dns.Name, err))
 		}
 		if haveSvc, svc, err := r.ensureDNSService(dns, clusterIP, daemonsetRef); err != nil {
+			// Set clusterIP to an empty string to cause ClusterOperator to report
+			// Available=False and Degraded=True.
+			clusterIP = ""
 			errs = append(errs, fmt.Errorf("failed to create service for dns %s: %v", dns.Name, err))
 		} else if !haveSvc {
 			errs = append(errs, fmt.Errorf("failed to get service for dns %s", dns.Name))


pkg/operator/controller/controller.go
```diff
@@ -399,9 +399,41 @@ func (r *reconciler) getClusterIPFromNetworkConfig() (string, error) {
...
+// returning an error if addr is already allocated or failures occur
+// while checking service resources for addr.
+func (r *reconciler) ensureServiceIP(addr net.IP) error {
+	namespaces := &corev1.NamespaceList{}
+	if err := r.client.List(context.TODO(), namespaces); err != nil {
+		return fmt.Errorf("failed to list namespaces: %v", err)
+	}
+	for _, ns := range namespaces.Items {
+		services := &corev1.ServiceList{}
+		if err := r.client.List(context.TODO(), services); err != nil {
```
Miciah, line 419: You aren't specifying a namespace in this `List` call, so it should list all services in all namespaces; no need to iterate over namespaces explicitly.

pkg/operator/controller/controller.go
```diff
@@ -399,9 +399,41 @@ func (r *reconciler) getClusterIPFromNetworkConfig() (string, error) {
...
+			return fmt.Errorf("failed to list services for namespace %s: %v", ns.Name, err)
+		}
+		for _, svc := range services.Items {
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}, &svc); err != nil {
```
Miciah, line 423: You already have the services; why get it again?

You need to ignore the DNS service itself so that you don't report that the service that _should_ have the address is in conflict.

Keep in mind that this logic runs every iteration, so it should avoid excessive API use.

pkg/operator/controller/controller.go
```diff
@@ -399,9 +399,41 @@ func (r *reconciler) getClusterIPFromNetworkConfig() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("invalid service cidr %v: %v", serviceCIDR, err)
 	}
+	// Ensure dnsClusterIP is not allocated to an existing service.
+	if err := r.ensureServiceIP(dnsClusterIP); err != nil {
+		return "", fmt.Errorf("failed to ensure service IP %s: %v", dnsClusterIP, err)
```
Miciah, line 404: Returning an error here causes `ensureDNS` to return before it updates the DNS's status.  It might be better to perform this check in `ensureDNSService`.  (It might also make sense for `ensureDNS` to update the DNS's status even if `getClusterIPFromNetworkConfig` returns an error, but maybe as a separate change.)

danehans:
> Is the goal here to make it easier to diagnose problem and let the administrator know which service to delete to fix it?

@Miciah yes, I think it's a middle ground that allows the DNS service to continue using the .10 IP while making the issue apparent by surfacing an error through operator status conditions. If the BZ continues to be a problem, then we can look into extending the functionality to select a different IP when a conflict is detected. How does that sound to you?

pkg/operator/controller/controller.go
danehans: @Miciah thanks for the review! Commit `4ce15d5` addresses the feedback above. `ensureDNSService()` will now return an error if the DNS Service IP is allocated to another Service. Since `ensureDNS()` is aggregating the errors, `syncDNSStatus()` should still get called to update status.

pkg/operator/controller/controller.go
danehans: Commit `4ce15d5` addresses the feedback above.

Miciah:
> I think it's a middle ground that allows the DNS service to continue using the .10 IP while making the issue apparent by surfacing an error through operator status conditions.

The Bugzilla report mentions that the operator is already logging a failure to create the service:

    2020-02-27T21:24:19.449200823Z time="2020-02-27T21:24:19Z" level=error msg="failed to reconcile request /default: failed to ensure dns default: failed to create service for dns default: failed to create dns service: Service \"dns-default\" is invalid: spec.clusterIP: Invalid value: \"10.253.159.10\": provided IP is already allocated"

Why isn't the operator already reporting the problem in the clusteroperator?  How would this change cause the operator to do so?

Miciah:
Let's set `clusterIP` to the empty string in `ensureDNS` if `ensureDNSService` returns an error, so that we pass an empty `clusterIP` value to `syncDNSStatus`.  That should cause the operator to report `Available=False` and `Degraded=True`  in the clusteroperator.  Does that sound reasonable?

sgreene570:
I reviewed the changes in `429bd72` to verify that they align with [Miciah's suggestion](https://github.com/openshift/cluster-dns-operator/pull/182#issuecomment-664397498), which I also agree with. 
/lgtm

PR #184, opened 2020-07-20
Port coredns errors alert

yeya24:
This pr updates the existing `CoreDNSPanicking` metric and add alerts for `CoreDNSErrorsHigh`. For more context, please check https://coreos.slack.com/archives/CCH60A77E/p1595256269123600

### DIFF
--- a/manifests/0000_90_dns-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_dns-operator_03_prometheusrules.yaml
@@ -23,3 +23,14 @@ spec:
           severity: warning
         annotations:
           message: "CoreDNS Health Checks are slowing down (instance {{ $labels.instance }})"
+      - alert: CoreDNSErrorsHigh
+        expr: |
+          (sum(rate(coredns_dns_response_rcode_count_total{rcode="SERVFAIL"}[5m]))
+            /
+          sum(rate(coredns_dns_response_rcode_count_total[5m])))
+          > 0.01
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          message: "CoreDNS is returning SERVFAIL for {{ $value | humanizePercentage }} of requests."
--- a/manifests/0000_90_dns-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_dns-operator_03_prometheusrules.yaml
@@ -10,7 +10,8 @@ spec:
     - name: openshift-dns.rules
       rules:
       - alert: CoreDNSPanicking
-        expr: coredns_panic_count_total > 0
+        expr: increase(coredns_panic_count_total[10m]) > 0
+        for: 5m
         labels:
           severity: warning
         annotations:


manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -22,3 +23,25 @@ spec:
...
+          (sum(rate(coredns_dns_response_rcode_count_total{rcode="SERVFAIL"}[5m]))
+            /
+          sum(rate(coredns_dns_response_rcode_count_total[5m])))
+          > 0.03
+        for: 5m
+        labels:
+          severity: critical
+        annotations:
+          message: "CoreDNS is returning SERVFAIL for {{ $value | humanizePercentage }} of requests."
+      - alert: CoreDNSErrorsHigh
```
sgreene570, line 37: I think having only one `CoreDNSErrorsHigh` alert would be best, in order to remove duplicate alerts and thus minimize "alert fatigue". As far as which value to use (.03 vs .01), I don't have any strong opinions on which one to use. Thoughts?

yeya24: I prefer to keep the `0.01` one since I don't think it is really necessary to add a critical level alert here.

sgreene570: sounds good to me.

yeya24: @RiRa12621  Hello, would you mind taking a look at this? Which alert should we use? The critical one or the warning one?

RiRa12621: +1 for warning from me
/cc @jewzaam

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -22,3 +23,14 @@ spec:
           severity: warning
         annotations:
           message: "CoreDNS Health Checks are slowing down (instance {{ $labels.instance }})"
+      - alert: CoreDNSErrorsHigh
+        expr: |
+          (sum(rate(coredns_dns_response_rcode_count_total{rcode="SERVFAIL"}[5m]))
+            /
+          sum(rate(coredns_dns_response_rcode_count_total[5m])))
+          > 0.01
```
jewzaam, line 31: Trying to grok what this expr is attempting to do.  It's a % of SERFAIL rate change sums over 5 minute increments and alerts if it's over 1%?

yeya24: Correct

sgreene570:
Thanks @yeya24 !
Looks good.
/lgtm

manifests/0000_90_dns-operator_03_prometheusrules.yaml
jewzaam: @yeya24 not really clear what this is trying to do with the sum mixed in.  Can this be written with it and simplify the query?  If we're trying to say "if the rate of failures increases by more than 1% of total responses over a 5 minute period of time" that would be:

```yaml
        expr: |
          sum(coredns_dns_response_rcode_count_total{rcode="SERVFAIL"})
            /
          sum(coredns_dns_response_rcode_count_total)
          > 0.01
        for: 5m
```

The difference is we're not looking at the change in failures over time.  I think this is better unless we assume there's some high chance of 1% of responses failing in a 5 min period.  The original query is looking at the changes in rates over time.  You could exclude sum given the alert is "for: 5m".  So if your rate of errors goes up slow enough relative to the rate of total responses you can have an ever increasing number of failures as long as the change _over time_ isn't more than 1% in a 5 minute block.

sgreene570: I think there's some misunderstanding of what the `sum` function does here. From my understanding, the `sum` function is just combining the rates for each separate server/zone, so that way the `rate` function doesn't return several values. 
So I don't think we would want to exclude `sum` here. @yeya24 does that sound right?

jewzaam: I agree, I left `sum` in and removed `rate`.  I see risk with sum of rates over time in that subtle cumulative increases in failure rates over time will not trip the alert.

yeya24: Sorry for the late. 

```
        expr: |
          sum(coredns_dns_response_rcode_count_total{rcode="SERVFAIL"})
            /
          sum(coredns_dns_response_rcode_count_total)
          > 0.01
        for: 5m
```

This query calculates the failure ratio over all the responses, which is not what I want.

`rate` restricts the samples in 5min and IMO this makes sense in this context.

PR #185, opened 2020-07-23
Bug 1860142: Adds Prometheus rule for forwarding healthcheck metrics

danehans:
This PR adds Prometheus rules for CoreDNS forward plugin health checking metrics. Note that metric `coredns_forward_healthcheck_broken_count_total` requires https://github.com/coredns/coredns/pull/4021 to get ported to [openshift/coredns](https://github.com/openshift/coredns).

Partially fixes [BZ 1860142](https://bugzilla.redhat.com/show_bug.cgi?id=1860142)

/assign @knobunc @miabbott 
/cc @sgreene570 @nstielau @miheer

### DIFF
--- a/manifests/0000_90_dns-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_dns-operator_03_prometheusrules.yaml
@@ -34,3 +34,17 @@ spec:
           severity: warning
         annotations:
           message: "CoreDNS is returning SERVFAIL for {{ $value | humanizePercentage }} of requests."
+      - alert: CoreDNSForwardHealthcheckFailureCount
+        expr: rate(coredns_forward_healthcheck_failure_count_total[10s]) > 3
+        for: 30s
+        labels:
+          severity: warning
+        annotations:
+          message: '{{ $value }} CoreDNS health checks are failing to upstream server {{ $labels.to }} on {{ $labels.instance }}'
+      - alert: CoreDNSForwardHealthcheckBrokenCount
+        expr: rate(coredns_forward_healthcheck_broken_count_total[5s]) > 3
+        for: 15s
+        labels:
+          severity: warning
+        annotations:
+          message: '{{ $value }} CoreDNS health checks are failing for all upstream servers on {{ $labels.instance }}'


manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -34,3 +34,15 @@ spec:
           severity: warning
         annotations:
           message: "CoreDNS is returning SERVFAIL for {{ $value | humanizePercentage }} of requests."
+      - alert: CoreDNSForwardHealthcheckFailureCount
+        annotations:
+          message: '{{ $value }} A CoreDNS counter for the number of failed health checks for an upstream server on {{ $labels.instance }}'
+        expr: coredns_forward_healthcheck_failure_count_total > 0
```
sgreene570, line 40: These alerts will fire until the applicable CoreDNS instances are restarted, since the counter is monotonically increasing.
I am in favor of using `increase(<metric>) > 0` instead, so that the alert is not continuously firing after the CoreDNS issues have been resolved (Which may or may not involve CoreDNS pods being restarted). 
The `CoreDNSPanicking` alert in this file already follows this pattern. Thoughts?

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -34,3 +34,15 @@ spec:
...
+        labels:
+          severity: warning
+      - alert: CoreDNSForwardHealthcheckBrokenCount
+        annotations:
+          message: '{{ $value }} A CoreDNS counter for the number of health check failures for all upstream servers on {{ $labels.instance }}'
```
sgreene570, line 45: I think the message for both of these alerts should be more along the lines of alert information, rather than generic metric information. See the messages used for the other alerts in this file. Alert messages typically aren't read by a cluster administrator until an alert is firing.

manifests/0000_90_dns-operator_03_prometheusrules.yaml
RiRa12621, line 40: How often are those healthchecks executed? 
By having it fire for >0 this imposes a risk for flapping alerts

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -34,3 +34,15 @@ spec:
...
+        expr: coredns_forward_healthcheck_broken_count_total > 0
+        labels:
+          severity: critical
```
RiRa12621, line 48: Sould probably also not just be >0

manifests/0000_90_dns-operator_03_prometheusrules.yaml
lilic: Here we want to do an error rate based alert. Here are some examples:

https://github.com/openshift/cluster-monitoring-operator/blob/09180fe1a032cf9fe0c0d98d1feb35c3ee2c21a0/assets/thanos-ruler/thanos-ruler-prometheus-rule.yaml#L32-L37

https://github.com/coreos/kube-prometheus/blob/master/manifests/prometheus-rules.yaml#L1853-L1868

manifests/0000_90_dns-operator_03_prometheusrules.yaml
```diff
@@ -34,3 +34,15 @@ spec:
...
+      - alert: CoreDNSForwardHealthcheckFailureCount
+        annotations:
+          message: '{{ $value }} A CoreDNS counter for the number of failed health checks for an upstream server on {{ $labels.instance }}'
+        expr: coredns_forward_healthcheck_failure_count_total > 0
+        labels:
+          severity: warning
+      - alert: CoreDNSForwardHealthcheckBrokenCount
+        annotations:
+          message: '{{ $value }} A CoreDNS counter for the number of health check failures for all upstream servers on {{ $labels.instance }}'
+        expr: coredns_forward_healthcheck_broken_count_total > 0
```
lilic, line 46: Same for this one as above. Note this is a critical severity alert you will page people during the night to fix something, so we want to make sure this alert is correct. You also need to add a for, so the time you want this to be true (the error based alert) for at least 5, 10 or 15 minutes before this fires. (see examples above)

manifests/0000_90_dns-operator_03_prometheusrules.yaml
danehans: > How often are those healthchecks executed?

@RiRa12621 the checks are executed every 0.5 sec by default.

danehans:
@sgreene570 @lilic @RiRa12621 thanks for the review.  Commit `4324b5e` addresses the feedback you provided. @lilic although I updated the expression for the rules, I'm unable to get them to fire on my test cluster even though the value for metric `coredns_forward_healthcheck_failure_count_total` is increasing based on the specified rate for the specified period:

```
sh-4.2$ curl -H "Authorization: Bearer ${TOKEN}" -k https://10.129.2.3:9154/metrics | grep coredns_forward_healthcheck_failure_count_total
# HELP coredns_forward_healthcheck_failure_count_total Counter of the number of failed healthchecks.
# TYPE coredns_forward_healthcheck_failure_count_total counter
coredns_forward_healthcheck_failure_count_total{to="216.239.32.1:53"} 4297

sh-4.2$ curl -H "Authorization: Bearer ${TOKEN}" -k https://10.129.2.3:9154/metrics | grep coredns_forward_healthcheck_failure_count_total
# HELP coredns_forward_healthcheck_failure_count_total Counter of the number of failed healthchecks.
# TYPE coredns_forward_healthcheck_failure_count_total counter
coredns_forward_healthcheck_failure_count_total{to="216.239.32.1:53"} 4302

sh-4.2$ curl -H "Authorization: Bearer ${TOKEN}" -k https://10.129.2.3:9154/metrics | grep coredns_forward_healthcheck_failure_count_total
# HELP coredns_forward_healthcheck_failure_count_total Counter of the number of failed healthchecks.
# TYPE coredns_forward_healthcheck_failure_count_total counter
coredns_forward_healthcheck_failure_count_total{to="216.239.32.1:53"} 4305
```

lilic:
As mentioned on slack, these alerts should be done in https://github.com/povilasv/coredns-mixin and we can just import this via cluster monitoring operator like we do with other jsonnet mixins, some alerts might already exist there. What do you think of that option?

danehans:
Closing based on https://github.com/openshift/cluster-dns-operator/pull/185#issuecomment-664170942.

danehans:
@lilic per your request, I created https://github.com/povilasv/coredns-mixin/pull/6 in favor of this PR.

RiRa12621:
Probably needs an additional ticket with the monitoring folks now to add said mixin to the stack

danehans:
> Probably needs an additional ticket with the monitoring folks now to add said mixin to the stack

@lilic when you say "ticket", is this a Jira user story that I need to create? Also, https://github.com/povilasv/coredns-mixin/pull/6 has yet to be reviewed. Does OpenShift have a clone of this repo that I can submit a PR to or do you have any suggestions on how to get https://github.com/povilasv/coredns-mixin/pull/6 reviewed?

PR #190, opened 2020-08-22
Prefer node-local endpoints

Miciah:
Configure the cluster DNS service to send traffic to, in descending order of preference, a node-local, zone-local, region-local, or any available endpoint.

Reprise of #156, which was commented out in #159.

* `assets/dns/service.yaml`: Specify `topologyKeys`.


Miciah:
Service topology is still disabled: https://github.com/openshift/kubernetes/blob/release-4.6/pkg/features/kube_features.go#L729

/close

PR #192, opened 2020-08-28
Bug 1872080: Avoid a bash "unbound variable" error

danwinship:
Seen in https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/periodic-ci-openshift-ovn-kubernetes-master-e2e-metal-ipi-dualstack-periodic/1299134739436802048/. There is some larger failure mode here which is not yet diagnosed, but on the way to figuring that out I saw that the current error-handling code in the `dns-node-resolver` pod is not working, resulting in the container crashing with:

    2020-08-28T01:57:14.539502724Z /bin/bash: line 38: svc_ips: unbound variable

rather than sleeping and looping


danwinship:
lol wut

```
2020-08-28T17:49:11.033552409Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:50:11.058408167Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:51:11.080578023Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:52:11.111326141Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:53:11.127450483Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:54:11.142883458Z /bin/bash: line 51: cmp: command not found
2020-08-28T17:55:11.157156676Z /bin/bash: line 51: cmp: command not found
...
```

was the check broken before and always false? Or maybe this is caused by the RHEL 8 migration?

danwinship:
So yeah, the RHCOS 8 images appear to not have `cmp` (or `diff`; same package, `diffutils`), while the RHCOS 7 images did. Possible RHCOS bug? (@jupierce? @alvaroaleman?)

You could do something like:

    if [[ "$(cat ${TEMP_FILE} | sha256sum)" != "$(cat ${HOSTS_FILE} | sha256sum)" ]]; then
      cp -f "${TEMP_FILE}" "${HOSTS_FILE}"
    fi

(The cat+pipe is needed because otherwise `sha256sum` includes the filename in the output which isn't what we want.)

alvaroaleman:
Sorry, I know nothing about how rhcos 7 and 8 differs. @jupierce and @yselkowitz are the best candidates for questions around that.

yselkowitz:
In what container is the script which calls `cmp` actually run (asking because some operators provide a script but run it in a different container)?

danwinship:
oh, duh, I'm mixing up RHCOS and UBI... I was thinking we couldn't install extra RPMs by hand but obviously we can

danwinship:
> In what container is the script which calls `cmp` actually run (asking because some operators provide a script but run it in a different container)?

Looks like `"openshift/origin-cli:v4.0"` or whatever that gets magically turned into.

yselkowitz:
@jupierce should diffutils be explicitly added to base for this?

yselkowitz:
https://github.com/openshift/images/pull/35 should fix this via inheritance.

danwinship:
> If the script is run in cli, then installing cmp in this operator isn't going to help.

yeah, I hadn't realized it was using a different image when I pushed that commit

danwinship:
(I later un-pushed that commit)

yselkowitz:
/retitle Bug 1872080: Avoid a bash "unbound variable" error

sgreene570:
Ah, nevermind, after further reading, the `ips` variable shouldn't ever be unbound.

https://github.com/openshift/cluster-dns-operator/blob/5c7efff2a3288271a891748ac2919417d3c6a347/assets/dns/daemonset.yaml#L117-L119

/approve

yselkowitz:
/hold
This might not be enough after all, testing now.

sgreene570:
> /hold
> This might not be enough after all, testing now.

Could you elaborate on whats going wrong?

danwinship:
The logs are still showing `/bin/bash: line 51: cmp: command not found` (eg https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/192/pull-ci-openshift-cluster-dns-operator-master-e2e-aws/1300410258996334592/artifacts/e2e-aws/gather-extra/pods/openshift-dns_dns-default-2xq66_dns-node-resolver.log) but that's technically a separate problem from the one being fixed in this PR...

assets/dns/daemonset.yaml
```diff
@@ -127,7 +127,7 @@ spec:
             # Update /etc/hosts only if we get valid service IPs
             # We will not update /etc/hosts when there is coredns service outage or api unavailability
             # Stale entries could exist in /etc/hosts if the service is deleted
-            if [[ "${#svc_ips[@]}" -ne 0 ]]; then
+            if [[ -n "${svc_ips[*]}" ]]; then
```
Miciah, line 130: Why is this necessary?  This works for me on RHEL76:
```bash
$ echo $BASH_VERSION
4.2.46(2)-release
$ set -u
$ declare -A svc_ips
$ [[ "${#svc_ips[@]}" -ne 0 ]]
$ echo $?
1
$
```
Do we need to report a BZ against RHEL8's Bash?

yselkowitz:
missing cmp would be indicative of a stale cache, problem has already been fixed in 4.6:base

assets/dns/daemonset.yaml
yselkowitz: much newer bash, probably expected

sgreene570:
`cmp` binary present in most recent run: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/192/pull-ci-openshift-cluster-dns-operator-master-e2e-aws/1301524653289771008

Failures look unrelated.

/retest

assets/dns/daemonset.yaml
Miciah: I filed [BZ#1875566](https://bugzilla.redhat.com/show_bug.cgi?id=1875566), "Bash says a declared but empty array is unbound".

danwinship: The behavior in 4.2 was a bug. It was fixed incorrectly in 4.3 (making both `${#svc_ips[@]}` and `${svc_ips[@]}` an error) and then fully fixed in 4.4 (making the former an error but the latter not). This has to do with POSIX compliance or something.

sgreene570: This PR seems to have fixed clusters using the RHEL 8 base, but broken clusters using the prior RHEL 7 base:
eg
```
sh-4.2# echo $BASH_VERSION
4.2.46(2)-release
sh-4.2# set -u
sh-4.2# declare -A svc_ips
sh-4.2# [[-n "${svc_ips[*]}"]]
sh: svc_ips[*]: unbound variable
```

danwinship: Is there any reason for something to be building images with UBI7 base and cluster-dns-operator git master?

Not sure if there's any alternative that would work in both versions. Simplest fix, if needed, would probably be to just set an additional variable in the earlier loop to indicate "I added at least one IP to `svc_ips`", and then test that here instead of testing `svc_ips` directly.

sgreene570: Looks like 4.6 nightly builds are still using UBI7, but that should to change to UBI8 this week to match CI, so we should actually be able to leave this fix as is. 

https://coreos.slack.com/archives/CCH60A77E/p1599489624207700 for context .

Miciah: Would `[[ -n "${svc_ips[*]-}" ]]` work on both Bash 4.2 and 4.4?

sgreene570: > Would `[[ -n "${svc_ips[*]-}" ]]` work on both Bash 4.2 and 4.4?

I think so... what exactly does the `-` change?

Miciah: `${parameter-default}` means to substitute the value of the variable `parameter` if it is set, or the literal value `default` (which in our case is the empty string) if `parameter` is not set.

danwinship: Ah, I had tried changing the original `${#svc_ips[@]}` to `${#svc_ips[@]:-0}`, which does not work.

Miciah: https://github.com/openshift/cluster-dns-operator/pull/193 changes the test to `[[ -n "${svc_ips[*]-}" ]]`.

PR #195, opened 2020-09-11
[WIP] experiment, do not merge: Try out coredns autopath

sgreene570:
playing around with coredns autopath via clusterbot

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -32,8 +32,9 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
 .:5353 {
     errors
     health
+    autopath @kubernetes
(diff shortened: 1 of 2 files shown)


sgreene570:
Screenshot
Proof that it works if anyone else wants to test.

sgreene570:
```
	# Enabling autopath causes coredns to use more memory since it needs to
	# maintain a watch on all pods. If autopath and "pods verified" mode are
	# both enabled, they will share the same watch. I.e. enabling both options
	# should have an equivalent memory impact of just one.
```

sgreene570:
https://github.com/coredns/coredns/issues/2890#issuecomment-502843775

sgreene570:
https://github.com/kubernetes/enhancements/pull/967

sgreene570:
performance analysis
https://stackoverflow.com/questions/60793911/how-to-avoid-coredns-resolving-overhead-in-kubernetes

PR #198, opened 2020-09-18
[Release 4.5] Bug 1880311: Fix bug in reflector not recovering from "Too large resource version" 

sgreene570:
This bump imports the recent fix in reflector/informer that was made in upstream client-go.

kubernetes/kubernetes#92688

As per https://bugzilla.redhat.com/show_bug.cgi?id=1880311.

Note this will be fixed in 4.6 with the bump to 1.19 https://github.com/openshift/cluster-dns-operator/pull/197

/assign @Miciah @danehans
/cc @frobware @knobunc @miheer


Miciah:
/lgtm
Seems like an oversight that we shipped with 1.18.0-rc.1.  This bump picks up several nice fixes (for example, https://github.com/kubernetes/kubernetes/pull/89652/commits/343c1e7636fe5c75cdd378c0b170b26935806de5 and https://github.com/kubernetes/kubernetes/pull/90476/commits/397319ed2b13c9a4e49718a018924d2265f7396e).  Maybe we need automation to verify that components are all on a common kube version, or at least are all on recent release versions (as opposed to rc).

sgreene570:
> Seems like an oversight that we shipped with 1.18.0-rc.1. 

Yea, I was also surprised to see that the `release-4.5` branch was using `1.18.0-rc.1` dependencies. In general I think we need to be more vigilant about dependency versioning as we approach release code freezes. We should definitely also discuss automation, too!

PR #200, opened 2020-09-25
Bug 1882323: Fix operator namespace in clusteroperator related resources

sgreene570:
The dns cluster operator is fed the DNS operator's namespace in pkg/operator/controller/status/controller.go. This commit fixes a regression brought in by https://github.com/openshift/cluster-dns-operator/pull/187 in which the cluster operator is given a reference to the "dns" namespace, which does not exist. This commit instead gives the cluster operator a reference to the correct operator namespace set at run time.

--- 

This regression affects must-gather, but could also have potentially unknown consequences since the cluster operator resource is pointing to a nil namespace.

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -104,7 +104,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	related := []configv1.ObjectReference{
 		{
 			Resource: "namespaces",
-			Name:     operatorcontroller.DNSClusterOperatorName().Name,
+			Name:     r.Config.OperatorNamespace,
 		},
 		{
 			Group:    operatorv1.GroupName,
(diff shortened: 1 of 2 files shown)


danehans:
@sgreene570 can you update the e2e test so that it compares current>expected status related objects? It would be nice to avoid running into this again in the future.

sgreene570:
@danehans good idea. Commit `4b4cd8b` addresses your suggestion. PTAL when you get the chance.

sgreene570:
```
level=error msg="Cluster operator machine-config Degraded is True with MachineConfigControllerFailed: Unable to apply 4.6.0-0.ci.test-2020-09-25-164140-ci-op-hsvhf0dq: timed out waiting for the condition during waitForControllerConfigToBeCompleted: controllerconfig is not completed: status for ControllerConfig machine-config-controller is being reported for 0, expecting it for 1"
```
/retest

Miciah:
Thanks!
/lgtm

PR #204, opened 2020-10-08
WIP: status: Verify and report endpoints availability

Miciah:
* `pkg/operator/controller/controller.go` (`ensureDNS`): Get the DNS service's endpoints and pass it to `syncDNSStatus`.
* `pkg/operator/controller/dns_status.go` (`syncDNSStatus`): Add `endpoints` parameter.  Use endpoints and the new `checkEndpoints` function to verify endpoints, and pass the results to `computeDNSStatusConditions`.
(`checkEndpoints`): New function.  Verify the provided endpoints by sending a DNS query for the API service to each endpoint, and return a list of errors.
(`computeDNSStatusConditions`): Add parameter for resolver errors.  Pass the errors to `computeDNSDegradedCondition`.
(`computeDNSDegradedCondition`): Add parameter for resolver errors.  Report `Degraded=True` if the list of resolver errors is nonempty,
* `pkg/operator/controller/dns_status_test.go` (`TestDNSStatusConditions`): Add test case for resolver errors.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -351,6 +351,7 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
 		if _, _, err := r.ensureDNSConfigMap(dns, clusterDomain); err != nil {
 			errs = append(errs, fmt.Errorf("failed to create configmap for dns %s: %v", dns.Name, err))
 		}
+		endpoints := &corev1.Endpoints{}
 		if haveSvc, svc, err := r.ensureDNSService(dns, clusterIP, daemonsetRef); err != nil {
 			// Set clusterIP to an empty string to cause ClusterOperator to report
 			// Available=False and Degraded=True.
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/dns_status.go
```diff
@@ -31,10 +39,47 @@ func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain
...
-	ds *appsv1.DaemonSet) []operatorv1.OperatorCondition {
+// checkEndpoints verifies that the provided endpoints by sending a DNS query
+// for the API service host name to each endpoint and returns a list of errors.
+func checkEndpoints(endpoints *corev1.Endpoints) []error {
+	const apiHost = "kubernetes.default.svc.cluster.local."
+	var (
+		errs      []error
+		addresses []string
+	)
+	for _, ep := range endpoints.Subsets {
```
sgreene570, line 50: How should this function behave if  `endpoints` is empty? Similarly, should this function (or maybe a different function) check that the expected number of endpoints are being evaluated?

pkg/operator/controller/controller.go
```diff
@@ -351,18 +351,28 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
...
 			errs = append(errs, fmt.Errorf("failed to create service for dns %s: %v", dns.Name, err))
 		} else if !haveSvc {
 			errs = append(errs, fmt.Errorf("failed to get service for dns %s", dns.Name))
-		} else if err := r.ensureMetricsIntegration(dns, svc, daemonsetRef); err != nil {
-			errs = append(errs, fmt.Errorf("failed to integrate metrics with openshift-monitoring for dns %s: %v", dns.Name, err))
+		} else {
+			if err := r.ensureMetricsIntegration(dns, svc, daemonsetRef); err != nil {
+				errs = append(errs, fmt.Errorf("failed to integrate metrics with openshift-monitoring for dns %s: %v", dns.Name, err))
+			}
+			name := types.NamespacedName{
```
sgreene570, line 366: nit: Add some comments here to make it clear that the metrics integration and endpoint GET call are not related.

sgreene570 (commented):
This is awesome! Some initial things after a quick glance. :smiley:

pkg/operator/controller/controller.go
Miciah: Done.

pkg/operator/controller/dns_status.go
Miciah: I added a check that adds an error if the number of addresses does not match the number of available pods as reported by the daemonset.  I think this suffices as `computeDNSDegradedCondition` sets `Degraded=True` if the daemonset reports that no pods are available.

PR #205, opened 2020-10-09
Bug 1884053: Configure CoreDNS to shut down gracefully

Miciah:
* `assets/dns/daemonset.yaml`: Specify a termination grace period of 120 seconds.  Change the readiness probe to use `:8181/ready`.
* `pkg/operator/controller/controller_dns_configmap.go`: (`corefileTemplate`): Configure CoreDNS's health plugin to sleep 60 seconds when CoreDNS is shut down.  Enable CoreDNS's ready plugin in order to provide a readiness endpoint on `:8181/ready`, which stops responding when CoreDNS is shutting down.
* `pkg/operator/controller/controller_dns_configmap_test.go` (`TestDesiredDNSConfigmap`): Adjust for changes to `corefileTemplate`.
* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check if the readiness probe or termination grace period changed, and update them if they did.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Add test cases to verify that `deploymentConfigChanged` detects updates to `.spec.template.spec.containers[].readinessProbe.httpGet` and `.spec.template.spec.terminationGracePeriodSeconds`.

----

These changes make CoreDNS behave more correctly but do not seem to resolve [BZ#1884053](https://bugzilla.redhat.com/show_bug.cgi?id=1884053) completely:  In testing, I still see that the DNS pod becomes unresponsive before the endpoints controller removes the pod's address from the endpoints when I reboot a node.  It is not clear whether this is a problem with CoreDNS, the kubelet, the endpoints controller, MCO, or something else.


pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -31,7 +31,10 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
 {{end -}}
 .:5353 {
     errors
-    health
+    health {
+        lameduck 60s
```
sgreene570, line 35: I'm not trying to suggest that `60s` is an unacceptable lameduck period, but I'm curious to hear how you arrived at this value.

sgreene570 (commented):
@Miciah, is https://github.com/coredns/coredns/issues/4099 a concern for us? It was rectified in upstream CoreDNS via https://github.com/coredns/coredns/pull/4167, but it is currently not available in the openshift fork of CoreDNS.

pkg/operator/controller/controller_dns_configmap.go
Miciah: The readiness probe has a period of 10 seconds and a failure threshold  of 3, so I figure it might take 30 seconds for the kubelet to recognize that the pod is unresponsive, and I doubled that for good measure.

Miciah:
> @Miciah, is [coredns/coredns#4099](https://github.com/coredns/coredns/issues/4099) a concern for us? It was rectified in upstream CoreDNS via [coredns/coredns#4167](https://github.com/coredns/coredns/pull/4167), but the fix is currently not available in the openshift fork of CoreDNS.

Yeah, I saw that, but according to the issue, the issue only applies if the connection is re-used.  The issue report mentions that Azure re-uses the connection.  I assume the kubelet does not re-use the connection or else someone would have said something.

sgreene570:
> I assume the kubelet does not re-use the connection or else someone would have said something.

Alright, fair enough. Just wanted to double check and raise awareness. Thanks!

sgreene570:
Looks good!
/lgtm

cgwalters:
Likely fallout from this in https://bugzilla.redhat.com/show_bug.cgi?id=1893360

cgwalters:
Revert inbound in https://github.com/openshift/cluster-dns-operator/pull/213

PR #206, opened 2020-10-26
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#510 and doesn't change existing behavior


frobware:
Was the go mod bump due to switching from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`? (otherwise LGTM, but curious about the bump)

guillaumerose:
The openshift/api bumped is required because some CRD manifests are in (DNS CRD here). 
https://github.com/openshift/api/blob/master/operator/v1/0000_70_dns-operator_00-custom-resource-definition.yaml#L6

The annotation needs to be on both sides.

PR #207, opened 2020-11-05
Bug 1895147: Remove memory limit on the CoreDNS container

Miciah:
Remove the memory resource limit for the CoreDNS container.  Control plane components should specify resource requests but not limits.

* `assets/dns/daemonset.yaml`: Remove resources limits for the dns container.

----

@smarterclayton


lihongan:
/bugzilla cc-qa
(follow https://issues.redhat.com/browse/DPTP-660)

PR #208, opened 2020-11-08
WIP: Delete node resolver daemonset if it exists

Miciah:
In OCP 4.7, the node resolver will be split out of the DNS daemonset into a separate daemonset.  If the node resolver daemonset exists on a 4.6 cluster (which could happen during downgrades), it should be deleted to avoid conflicts with the node resolver container in the DNS daemonset.

* `pkg/operator/controller/controller.go` (`ensureDNS`): Call the new `ensureNodeResolverDaemonset` method before calling `ensureDNSDaemonSet` in order to delete any old node resolver daemonset before reconciling the DNS daemonset.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go`: New file.
(`ensureNodeResolverDaemonset`): New method.  Check for and delete any existing node resolver daemonset, using the `desiredNodeResolverDaemonSet` function and the `currentNodeResolverDaemonSet` and `deleteNodeResolverDaemonSet` methods.
(`desiredNodeResolverDaemonSet`): New function.  Return `false`.
(`currentNodeResolverDaemonSet`, `deleteNodeResolverDaemonSet`): New methods.
* `pkg/operator/controller/controller_dns_node_resolver_test.go`: New file.
(`TestDesiredNodeResolverDaemonset`): Verify that `desiredNodeResolverDaemonSet` returns `false` and a nil error value.
* `pkg/operator/controller/names.go` (`NodeResolverDaemonSetName`): New function.  Return the name of the node resolver daemonset.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -334,6 +334,17 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
 	}
 
 	errs := []error{}
+
+	// In 4.6 and earlier, the node resolver runs as a container in the
+	// daemonset that ensureDNSDaemonSet manages, and if a separate node
+	// resolver daemonset exists, then ensureNodeResolverDaemonset deletes
+	// it.  We want to delete the node resolver daemonset before updating
+	// the DNS daemonset to avoid having both the node resolver running in
+	// both daemonsets at the same time.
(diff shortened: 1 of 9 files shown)


Miciah:
/close
For now we are not planning to backport this change to 4.7; having the node-resolver container running in two daemonsets should not cause problems as each daemonset's container should be writing the same content in `/etc/hosts` and doing so using an atomic rename operation.

PR #209, opened 2020-11-08
Add node-resolver daemonset

Miciah:
#### Delete node-resolver daemonset if it exists

In OCP 4.8, the node resolver is be split out of the DNS daemonset into a separate daemonset.  If the node-resolver daemonset exists on a 4.7 cluster (which could happen during downgrades), it should be deleted.

* `pkg/operator/controller/controller.go` (`ensureDNS`): Call the new `ensureNodeResolverDaemonset` method in order to delete any old node-resolver daemonset.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go`: New file.
(`ensureNodeResolverDaemonset`): New method.  Check for and delete any existing node-resolver daemonset, using the `desiredNodeResolverDaemonSet` function and the `currentNodeResolverDaemonSet` and `deleteNodeResolverDaemonSet` methods.
(`desiredNodeResolverDaemonSet`): New function.  Return false.
(`currentNodeResolverDaemonSet`, `deleteNodeResolverDaemonSet`): New methods.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset_test.go`: New file.
(`TestDesiredNodeResolverDaemonset`): Verify that `desiredNodeResolverDaemonSet` returns false and a nil error value.
* `pkg/operator/controller/names.go` (`NodeResolverDaemonSetName`): New function.  Return the name of the node-resolver daemonset.

#### Fix error message for failed DNS status updates

* `pkg/operator/controller/controller.go` (`ensureDNS`): Use the dns's name rather than the dns daemonset's name.  Also, use `%q` for the name and `%w` for the error message.

#### status: Use `GetScaledValueFromIntOrPercent`

Replace a call to `GetValueFromIntOrPercent` with a call to `GetScaledValueFromIntOrPercent`, which properly rejects string values that are missing the percent sign.

* `pkg/operator/controller/dns_status.go` (`computeDNSDegradedCondition`): Use `GetScaledValueFromIntOrPercent`.
* `pkg/operator/controller/dns_status_test.go` (`TestComputeDNSDegradedCondition`): Enable test case where `MaxUnavailable` is a string of digits without a percent sign.

#### status: Check `MaxUnavailable` as late as possible

Check if the DNS daemonset has 0 pods, or if the DNS service does not have a cluster IP address, before checking if `MaxUnavailable` is valid.  We can definitively report that DNS is degraded if the DNS daemonset has 0 pods or if the DNS service has no cluster IP address, and by checking these conditions first, we avoid reporting that that the status is unknown when one of the first two conditions is met and `MaxUnavailable` is also invalid.

* `pkg/operator/controller/dns_status.go` (`computeDNSDegradedCondition`): Move the check for an invalid value for `MaxUnavailable` after the checks for having 0 pods or empty service's cluster IP address.
* `pkg/operator/controller/dns_status_test.go` (`TestComputeDNSDegradedCondition`): Add test case to verify that `computeDNSDegradedCondition` determines that the DNS is degraded if 0 pods are available and `MaxUnavailable` is invalid.

#### Add node-resolver daemonset

Split the node resolver out of the DNS daemonset into a separate daemonset.

The node-resolver daemonset needs to run on every node and therefore must tolerate all taints.  Because the node resolver runs in a separate daemonset, the blanket toleration can be removed from the DNS daemonset.

* `assets/dns/daemonset.yaml`: Remove the "dns-node-resolver" container, "hosts-file" volume, and blanket toleration.
* `assets/node-resolver/service-account.yaml`: New file.  Define a service account for the node-resolver daemonset.
* `assets/node-resolver/update-node-resolver.sh`: New file for the shell script that was previously in the DNS daemonset.
* `pkg/manifests/manifests.go` (`nodeResolverScriptAsset`): New const with the path to the node-resolver script asset.
(`nodeResolverServiceAccountAsset`): New const with the path to the node-resolver service account asset.
(`NodeResolverScript`): New function that returns the node-resolver script.
(`NodeResolverServiceAccount`): New function that returns the node-resolver service account.
* `pkg/manifests/manifests_test.go` (`TestManifests`): Verify `NodeResolverScript` and `NodeResolverServiceAccount`.
* `pkg/operator/controller/controller.go` (`ensureDNSNamespace`): Create the node-resolver service account.
(`ensureDNS`): Delete the `clusterIP` and `clusterDomain` arguments to `ensureDNSDaemonSet`.
* `pkg/operator/controller/controller_dns_daemonset.go` (`ensureDNSDaemonSet`): Delete the `clusterIP` and `clusterDomain` parameters, and delete the `clusterIP`, `clusterDomain`, and `openshiftCLIImage` arguments in the call to `desiredDNSDaemonSet`.
(`desiredDNSDaemonSet`): Delete the `clusterIP`, `clusterDomain`, and `openshiftCLIImage` parameters and logic to configure the node-resolver container.
(`daemonsetConfigChanged`): Delete logic for the node-resolver container.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`, `TestDaemonsetConfigChanged`): Delete logic related to the node-resolver container.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go` (`services`): New const with the list of service names for which entries should be added to `/etc/hosts`.
(`nodeResolverScript`): New variable with the shell script that updates `/etc/hosts`.
(`ensureNodeResolverDaemonset`): Add logic to create or update the node-resolver daemonset as needed, using the new `createNodeResolverDaemonSet` and `updateNodeResolverDaemonSet` methods.
(`desiredNodeResolverDaemonSet`): Return true and a daemonset.
(`createNodeResolverDaemonSet`): New method.
(`updateNodeResolverDaemonSet`): New method that updates the node-resolver daemonset using the new `nodeResolverDaemonSetConfigChanged` function.
(`nodeResolverDaemonSetConfigChanged`): New function that checks if the current daemonset needs to be updated, and if so, return true and an updated version of the daemonset.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset_test.go` (`TestDesiredNodeResolverDaemonset`): Verify that `desiredNodeResolverDaemonSet` returns true and a valid daemonset.
(`TestNodeResolverDaemonsetConfigChanged`): New test.  Verify that `nodeResolverDaemonSetConfigChanged` correctly detects changes to the node-resolver daemonset.
* `pkg/operator/controller/names.go` (`nodeResolverDaemonSetLabelName`): New const with the name of the label that identifies the node-resolver daemonset.
(`NodeResolverDaemonSetPodSelector`): New function that returns the label selector for node-resolver pods.

#### Use the host network for node-resolver pods

Configure node-resolver pods to use the host network in order to avoid consuming SR-IOV devices on nodes that have Smart NICs.

* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go` (`desiredNodeResolverDaemonSet`): Use the host network.

#### `syncDNSStatus`: Report status of node-resolver pods

* `pkg/operator/controller/controller.go` (`ensureDNS`): Always call `syncDNSStatus`.  Store the `haveDNSDaemonset` and `dnsDaemonset` return values from `ensureDNSDaemonSet` and the `haveNodeResolverDaemonset` and `nodeResolverDaemonset` return values from `ensureNodeResolverDaemonSet`, and pass these values to `syncDNSStatus`.
* `pkg/operator/controller/dns_status.go` (`syncDNSStatus`): Add parameters for the new arguments, and pass them to `computeDNSStatusConditions`.
(`computeDNSStatusConditions`): Add parameters for the new arguments, and pass them to `computeDNSDegradedCondition`, `computeDNSProgressingCondition`, and `computeDNSAvailableCondition`.
(`computeDNSDegradedCondition`, `computeDNSProgressingCondition`): Check if the DNS and node-resolver daemonsets exist, and check the status of the node-resolver daemonset.
(`computeDNSAvailableCondition`): Check if the DNS daemonset exists.
* `pkg/operator/controller/dns_status_test.go` (`TestDNSStatusConditions`): Streamline test case definitions.  Add inputs for `haveDNSDaemonset`, `haveNodeResolverDaemonset`, and `nodeResolverDaemonset`.  Add a test case where the daemonsets do not exist.  Add test cases to cover various node-resolver daemonset states.
(`TestComputeDNSDegradedCondition`): Add node-resolver daemonset input.  Add test cases to cover various node-resolver daemonset states.  Print the actual status condition if a test case fails.

#### Bump `openshift/api` for DNS pod placement API

Bump to github.com/Miciah/api@94c5cb42d9fa5c4bea24c720e900d368b0a9eb3a to get the DNS pod placement API.

* `go.mod`: Bump.
* `go.sum`:
* `manifests/0000_70_dns-operator_00-custom-resource-definition.yaml`:
* `vendor/github.com/openshift/api/*`:
* `vendor/modules.txt`: Regenerate.

#### Implement DNS pod placement API

This change implements [NE-547](https://issues.redhat.com/browse/NE-547).

* `pkg/operator/controller/controller_dns_daemonset.go` (`desiredDNSDaemonSet`): Use the node selector and tolerations, if specified, from the dns to configure the DNS daemonset.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`): Verify that the default node selector and tolerations are set.
(`TestDesiredDNSDaemonsetNodePlacement`): New test.  Verify that `desiredDNSDaemonSet` copies the node selector and tolerations from the dns to the daemonset.

#### Guard against overly restrictive node placement

When handling an update for the dns daemonset, check whether the daemonset currently has any pods and whether the update would prevent any pods from being scheduled, and ignore the update if both of those conditions are met to reduce the risk that the administrator will break cluster DNS with poorly chosen node-placement parameters.

* `manifests/0000_70_dns-operator_00-cluster-role.yaml`: Grant the operator access to nodes.
* `pkg/operator/controller/controller_dns_daemonset.go` (`updateDNSDaemonSet`): Check if the update is safe, using the new `daemonsetUpdateIsSafe` method, and omit the updated node selector and tolerations from the update if they are not safe.
(`daemonsetUpdateIsSafe`): New method.  Given the current and updated dns daemonsets, check whether the current daemonset has pods and whether the updated daemonset's node selector and tolerations would allow any pods to be scheduled, and return a value indicating whether the update is safe.
(`tolerationsTolerateTaints`): New helper function for `daemonsetUpdateIsSafe`.
* `test/e2e/operator_test.go` (`TestDNSNodePlacement`): New test.  Verify that the node-placement API behaves correctly when a node selector for master nodes is configured and when a node selector that selects no nodes is configured.

#### computeDNSProgressingCondition: Check node selector

Compare the node selector and tolerations that are specified in the dns with those specified in the daemonset, and report that the operator is progressing if the values do not match.

* `pkg/operator/controller/dns_status.go` (`syncDNSStatus`): Pass the dns to `computeDNSStatusConditions.`
(`computeDNSStatusConditions`): Replace the conditions parameter with a parameter for the dns.  Pass the dns to `computeDNSProgressingCondition`.
(`computeDNSProgressingCondition`): Add a parameter for the dns.  Compare the dns's node selector and tolerations with the daemonset's, and report if they do not match.
* `pkg/operator/controller/dns_status_test.go` (`TestDNSStatusConditions`): Update to call `computeDNSStatusConditions` with an empty dns.
(`TestComputeDNSProgressingCondition`): New test.  Verify that `computeDNSProgressingCondition` behave as expected.

----

- [x] Revendor once https://github.com/openshift/api/pull/870 is merged.
- [ ] ~Depends on #208, which should be backported to release-4.6 before this change is shipped in 4.7.~  Having the node-resolver container running in two daemonsets is redundant but should be harmless.  I kept the first commit, which adds the logic to delete the node-resolver daemonset, as a distinct commit in case we decide to backport it, but it may not be necessary to do so.


assets/dns/update-node-resolver.sh
```diff
@@ -0,0 +1,57 @@
...
+  declare -A svc_ips
+  for svc in "${services[@]}"; do
+    # Fetch service IP from cluster dns if present. We make several tries
+    # to do it: IPv4, IPv6, IPv4 over TCP and IPv6 over TCP. The two last ones
+    # are for deployments with Kuryr on older OpenStack (OSP13) - those do not
+    # support UDP loadbalancers and require reaching DNS through TCP.
+    cmds=('dig -t A @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+          'dig -t AAAA @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+          'dig -t A +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"'
+          'dig -t AAAA +tcp @"${NAMESERVER}" +short "${svc}.${CLUSTER_DOMAIN}"')
```
rfredette, line 25: `cmds` here will need to updated to keep the bug fix from #223

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
```diff
@@ -0,0 +1,247 @@
...
+			Namespace: name.Namespace,
+		},
+		Spec: appsv1.DaemonSetSpec{
+			Selector: NodeResolverDaemonSetPodSelector(),
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: NodeResolverDaemonSetPodSelector().MatchLabels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
```
zshi-redhat, line 104: Would it be possible to use `hostNetwork: true` for the node resolver daemonset?

This is related to the Smart-NIC use case we are trying to address (WIP) on OpenShift baremetal, where for a particular node, it is not possible to create a regular veth pod with ovn-k8s network type. so every infra pod like dns resolver that needs to run on every node (including the Smart-NIC node) may not be able to create if not running with `hostNetwork:true`.

Miciah: I don't see why we couldn't use `hostNetwork: true` for the node resolver daemonset.  If I understand correctly, the use-case is that the daemonset's pods may need to run on nodes where container networking cannot be configured for the pod; in this scenario, can the pod still use the service network to preform DNS queries?

zshi-redhat: > I don't see why we couldn't use `hostNetwork: true` for the node resolver daemonset. If I understand correctly, the use-case is that the daemonset's pods may need to run on nodes where container networking cannot be configured for the pod; 

Container networking can only be used together with a SR-IOV (hardware device) interface in Smart-NIC case.
Since SR-IOV devices are limited resource on a k8s node, resources.limits & resources.requests are required in pod spec in order for dns or other infra pods to request such network (other pod spec changes like pod annotation might also be needed). So an easy fix in my opinion is to run dns pod in hostNetwork - not using container networking, Or be able to update the dns pod spec with a user-defined resource which I think is not trivial.

> in this scenario, can the pod still use the service network to perform DNS queries?

Yes, service network will still be available.

Miciah: Yeah, that makes sense.  I don't see any problem with using `hostNetwork: true`, but I'll raise the question with my team for feedback.

Miciah: Latest push specifies `hostNetwork: false` on the node-resolver daemonset.

zshi-redhat: > Latest push specifies `hostNetwork: false` on the node-resolver daemonset.

Thanks @Miciah ! 
I think you mean `hostNetwork: true` as written in the code :-)

Miciah: > I think you mean `hostNetwork: true` as written in the code :-)

Indeed!  The code was right, but the GitHub comment was wrong.  Sorry for the error!

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if len(clusterDomain) > 0 {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "CLUSTER_DOMAIN",
+			Value: clusterDomain,
+		})
+	}
+	trueVal := true
+	name := NodeResolverDaemonSetName()
+	daemonset := appsv1.DaemonSet{
```
sgreene570, line 94: I am not opposed to defining this new daemonset completely in code, but out of curiosity, any particular reason not to make the daemonset skeleton available via `assets/`?

test/e2e/operator_test.go
```diff
@@ -442,3 +444,114 @@ func TestDNSForwarding(t *testing.T) {
...
+			"foo": "bar",
+		},
+	}
+	if err := cl.Update(context.TODO(), defaultDNS); err != nil {
+		t.Fatalf("failed to update dns %s: %v", defaultDNS.Name, err)
+	}
+
+	// Verify that DNS pods continue to run (meaning the label selector that
+	// matched no nodes is being ignored).
+	err = wait.PollImmediate(1*time.Second, 60*time.Second, func() (bool, error) {
```
sgreene570, line 547: Might make sense to run this polling loop for more than one minute, since if the DNS daemonset were to scale down to zero replicas, that would presumably take some time.

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -158,13 +153,88 @@ func (r *reconciler) updateDNSDaemonSet(current, desired *appsv1.DaemonSet) (boo
...
 }
 
+// daemonsetUpdateIsSafe takes current and updated daemonsets, checks if the
+// update is safe, and returns a Boolean value and a string value indicating
+// whether the update is safe or the reason why it is not.
+func (r *reconciler) daemonsetUpdateIsSafe(current, updated *appsv1.DaemonSet) (bool, string, error) {
+	// Allow the update if the current daemonset doesn't even have a
+	// valid selector.
+	selector, err := metav1.LabelSelectorAsSelector(current.Spec.Selector)
+	if err != nil {
```
sgreene570, line 177: Should we log a message here?

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -158,13 +153,88 @@ func (r *reconciler) updateDNSDaemonSet(current, desired *appsv1.DaemonSet) (boo
...
+	// cannot reduce the number of pods below 0.
+	if len(podList.Items) == 0 {
+		return true, "", nil
+	}
+	nodeLabelSelector := &metav1.LabelSelector{
+		MatchLabels: updated.Spec.Template.Spec.NodeSelector,
+	}
+	nodeSelector, err := metav1.LabelSelectorAsSelector(nodeLabelSelector)
+	if err != nil {
+		return true, "", nil
```
sgreene570, line 198: Do we want to allow the update if `updated.Spec.Template.spec.NodeSelector` is invalid?

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -158,13 +153,88 @@ func (r *reconciler) updateDNSDaemonSet(current, desired *appsv1.DaemonSet) (boo
...
+		tolerations := updated.Spec.Template.Spec.Tolerations
+		if !tolerationsTolerateTaints(tolerations, taints) {
+			continue
+		}
+		// We have found a node to which it should be possible to
+		// schedule pods for the updated daemonset, so the update
+		// can be considered safe.
+		return true, "", nil
+	}
+	return false, "updated daemonset's node placement parameters match 0 nodes", nil
```
sgreene570, line 221: Maybe we should clarify this error message a bit more. How about the following?
```suggestion
	return false, "updated daemonset's node placement parameters match 0 schedulable nodes", nil
```

assets/dns/daemonset.yaml
```diff
@@ -158,15 +77,12 @@ spec:
...
-      - name: hosts-file
-        hostPath:
-          path: /etc/hosts
-          type: File
       - name: metrics-tls
         # secretName is set at runtime
       tolerations:
       # DNS needs to run everywhere. Tolerate all taints
-      - operator: Exists
+      - key: "node-role.kubernetes.io/master"
```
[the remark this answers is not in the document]
Miciah: I'll delete the comment.  As for the toleration, I'm ambivalent.  First, `node-role.kubernetes.io/infra` isn't exactly codified as a standard anywhere as far as I know, but then it does feature in a lot of examples in the documentation.  Second, seeing a toleration for `node-role.kubernetes.io/infra` could be confusing to users who don't have infra nodes, but then has no effect if no infra nodes exist, and moreover it does seem likely that administrators who have configured infra nodes want DNS pods to run on them.  Finally, the ingress operator doesn't configure router pods with a toleration for infra nodes, and I don't have a counterargument to that point—having DNS pods tolerate infra by default and not having router pods tolerate infra by default seems arbitrary and inconsistent.  So I'm leaning slightly towards not having a toleration for infra nodes by default.

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
Miciah: I'm kind of fed up with yaml:

* Running `make generate` after updating yaml is an extra step for developers.  
* Loading the yaml so that we can mutate the result is an extra step in the code.  
* Modifying `bindata.go` leads to more merge conflicts, especially during backports, and thus more work for developers.  
* Using yaml instead of Go leads to uncaught syntax errors, such as the one that https://github.com/openshift/cluster-dns-operator/pull/221 fixes.  (Maybe we just need better unit tests.)  

Thus I'm disinclined to add new yaml assets for values that we manipulate in Go code.  

As far as I'm concerned, the only situations in which using assets is making life easier for us are simple yaml that we don't manipulate in Go code, such as CVO manifests and RBAC resources; and assets that are useful independently, such as the `update-node-resolver.sh` shell script that you might want to run directly or assets that you are likely to want to pass into `oc apply` or the like.

test/e2e/operator_test.go
Miciah: Good point.  I don't want  to slow down CI's happy path too much, but it would make sense to wait a few minutes.  I've increase the polling loop's timeout to 3 minutes.  I've also increased the timeout of the other polling loop (that verifies that pods are only running on master nodes when a selector for master nodes has been set) to 3 minutes to reduce the risk that that loop could cause flakiness.

pkg/operator/controller/controller_dns_daemonset.go
Miciah: Yup!  Thanks!

pkg/operator/controller/controller_dns_daemonset.go
Miciah: Done.  Thanks!

pkg/operator/controller/controller_dns_daemonset.go
Miciah: Whoops!  Probably not.  Fixed.  Thanks!

assets/dns/daemonset.yaml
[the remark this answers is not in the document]
sgreene570: Fair enough.

My only concern with not tolerating infra node taints is that when a cluster that uses infra node taints is upgraded, DNS pods will _not_ be scheduled to the the infra nodes (even though they probably previously were). Granted, its trivial for a cluster administrator to adjust the DNS spec to reschedule DNS pods to infra nodes. Maybe we should be sure to include a release note about the change in the DNS daemonset's default tolerations, or something along those lines?

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
sgreene570: > As far as I'm concerned, the only situations in which using assets is making life easier for us are simple yaml that we don't manipulate in Go code, such as CVO manifests and RBAC resources; and assets that are useful independently, such as the update-node-resolver.sh shell script that you might want to run directly or assets that you are likely to want to pass into oc apply or the like.

Sounds good to me.

pkg/operator/controller/dns_status.go
```diff
@@ -49,112 +51,192 @@ func computeDNSStatusConditions(oldConditions []operatorv1.OperatorCondition, cl
...
-func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clusterIP string,
-	ds *appsv1.DaemonSet) operatorv1.OperatorCondition {
+// based on the status of clusterIP and the DNS and node-resolver daemonsets.
+func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clusterIP string, haveDNSDaemonset bool, dnsDaemonset *appsv1.DaemonSet, haveNodeResolverDaemonset bool, nodeResolverDaemonset *appsv1.DaemonSet) operatorv1.OperatorCondition {
 	degradedCondition := &operatorv1.OperatorCondition{
 		Type: operatorv1.OperatorStatusTypeDegraded,
 	}
-	numberUnavailable := ds.Status.DesiredNumberScheduled - ds.Status.NumberAvailable
-	// TODO: Replace GetValueFromIntOrPercent with
-	// GetScaledValueFromIntOrPercent after rebasing on Kubernetes 1.20.
```
sgreene570, line 69: Convenient timing that we get to pick this new function up now :grin:

Miciah:
Rebased.

PR #213, opened 2020-11-13
Revert "Configure CoreDNS to shut down gracefully"

cgwalters:
This reverts commit f094ddf7edc95dad8398179482687bc2a7a0c15b.
It didn't actually help, and causes system shutdown to take
noticeably longer which makes the MCO tests time out.

The real fix will involve backporting
https://github.com/kubernetes/kubernetes/pull/96129


kikisdeliveryservice:
We really need this revert bc our ci is currently blocked.

Miciah:
> It didn't actually help, and causes system shutdown to take
> noticeably longer which makes the MCO tests time out.

It seems like #205 is doing the right thing by making the readiness probe actually report readiness and adding a grace period before the pod is terminated.  Is #205 technically wrong, or is the case that #205 is technically correct and the broken behavior that https://github.com/kubernetes/kubernetes/pull/96129 fixes means that doing the correct thing in the operator makes the overall upgrade behavior worse?

> The real fix will involve backporting
> [kubernetes/kubernetes#96129](https://github.com/kubernetes/kubernetes/pull/96129)

If we revert #205 and subsequently backport https://github.com/kubernetes/kubernetes/pull/96129, do we then need to restore #205 (i.e., revert the reversion)?  

Note that reverting #205 might improve the situation for node reboots, but I think it will make the situation worse for rolling updates of the pod where node reboots are not involved.  Maybe the tradeoff is appropriate though.

Miciah:
> We really need this revert bc our ci is currently blocked.

All right, if #205 broke CI, we need to revert it and figure out a way to re-introduce it (if appropriate) in a way that doesn't break CI.  

/approve
/lgtm

cgwalters:
To clarify, it's the MCO's CI which is hit the most by this.  The openshift e2e tests don't apply changes to nodes, and the e2e-upgrade only does it once.

> It seems like #205 is doing the right thing by making the readiness probe actually report readiness and adding a grace period before the pod is terminated.

Having a readiness probe makes total sense but...the problem is the grace period.   If we're going to reboot the node, then what we want to do is:

- Drop it out of the list of endpoints
- reboot and kill the pod without any "grace"

Miciah:
> Having a readiness probe makes total sense but...the problem is the grace period. 

Would a shorter grace period (say 20 seconds) be an acceptable compromise?  I used 2 minutes in #205 to give kubelet's readiness probes time to fail several times and thus mark the pod as unready.  Really though, the endpoints controller _should_ immediately take note that the pod's deletion timestamp is set.  So if the endpoints controller and kube-proxy are operating properly, they _should_ stop using a given DNS pod's endpoint pretty quickly (I'd guess <10s, possibly a bit longer on a stressed cluster?) once deletion is requested, so the grace period probably can be pretty short to prevent any blips.

kikisdeliveryservice:
This seems like it's getting hit by : https://bugzilla.redhat.com/show_bug.cgi?id=1897604

see: incident-kcm-auth channel for current status

cgwalters:
Whee, timed out on lease.
/test e2e-aws

cgwalters:
Right hmm DNS is degraded...I think I may see the problem.  We still need to carry the diff logic to roll out a new daemonset if the readiness probe changes across upgrades, since that's what we're doing now.

cgwalters:
And yep, upgrade test is green now.  This case was clearly our CI tests doing their job correctly.  Events like CI going red across the board are bad because they teach us not to look in depth at failures and hope for things to go through in retries.

yuqi-zhang:
/retest

The e2e-aws errors in that failed test seem to happen occasionally on PRs from this repo (see 212 and 210) so maybe its a flake?

cgwalters:
Hmm in that e2e-aws run looks like one node failed to fully join the cluster.  We didn't get journal logs.  Nothing obvious on the console (though man we really need to teach the MCD to dump status to the console).

PR #216, opened 2020-11-26
Bug 1915473: Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-dns-operator CRD
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.


cfergeau:
Could this get a review?

cfergeau:
@danehans @candita could you take a look?

cfergeau:
How can I get a review on this? @Miciah @sgreene570 ?

praveenkumar:
Can some look into this with higher priority because it is breaking the CI in cluster is provision with `single-node-developer` profile as this operator is excluded and there is no dns for cluster.

sgreene570:
Sorry for the delay!
/lgtm

PR #217, opened 2020-11-26
Bug 1880148: Set DNS DaemonSet's maxUnavailable value to 10%

Miciah:
Enable rolling updates to scale linearly with the cluster number of nodes by setting the `maxUnavailable` value to a percentage instead of the default value of 1.

* `assets/dns/daemonset.yaml`: Set `maxUnavailable` to 10%.
* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check if the update strategy changed and update it if it did.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Add test case to verify that `daemonsetConfigChanged` detects updates to `spec.updateStrategy`.
* `pkg/operator/controller/dns_status.go` (`computeDNSDegradedCondition`): Handle percentages values for `maxUnavailable`.
* `pkg/operator/controller/dns_status_test.go` (`TestComputeDNSDegradedCondition`): New test.


assets/dns/daemonset.yaml
```diff
@@ -150,6 +150,10 @@ spec:
       dnsPolicy: Default
       nodeSelector:
         kubernetes.io/os: linux      
+      updateStrategy:
+        type: RollingUpdate
+        rollingUpdate:
+          maxUnavailable: 10%
```
sgreene570, line 156: TIL that `maxUnavailable` values specified by percentage are rounded _up_ [for Daemonsets](https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L488-L491) and _down_ [for Deployments](https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L347-L349). In other words, setting `maxUnavailable: 10%` here is effectively setting `maxUnavailable: 1` for any cluster with <= 10 nodes (such as our CI clusters) as opposed to `maxUnavailable: 0`.

Would it be worth adding in a comment here to clarify that the calculated percentage is rounded up?

pkg/operator/controller/dns_status_test.go
```diff
@@ -168,6 +168,108 @@ func TestDNSStatusConditions(t *testing.T) {
...
+			clusterIP: "172.30.0.10",
+			ds:        makeDaemonSet(6, 5, intstr.FromInt(1)),
+			expected:  operatorv1.ConditionFalse,
+		},
+		{
+			name:      "all available",
+			clusterIP: "172.30.0.10",
+			ds:        makeDaemonSet(6, 6, intstr.FromString("10%")),
+			expected:  operatorv1.ConditionFalse,
+		},
```
sgreene570, line 258: Could you add a test case with an invalid `maxUnavailable` `intstr` such as `intsr.FromString("TEST")`?

sgreene570 (commented):
Some quick comments.

pkg/operator/controller/dns_status_test.go
Miciah: Added.  Thanks!

assets/dns/daemonset.yaml
Miciah: Comment added.

PR #218, opened 2020-11-26
Keep reporting old versions while progressing

Miciah:
* `pkg/operator/controller/status/controller.go` (`Reconcile`): Use the new `checkDNSStatusCondition` function to check whether the DNS is progressing. Pass this information to `computeOperatorStatusVersions`.
(`computeOperatorStatusVersions`): Replace `allDNSESAvailable` parameter with `dnsProgressing` parameter.  Continue reporting the old versions as long as the DNS is progressing.
(`checkDNSAvailable`): Rename...
(`checkDNSStatusCondition`): ...to this.  Parameterize the status condition type.
(`computeOperatorProgressingCondition`): Delete bogus to-do comment.
* `pkg/operator/controller/status/controller_test.go` (`TestComputeOperatorStatusVersions`): Replace `dnsAvailable` test input with `dnsProgressing` and adjust test cases accordingly.

----

@wking

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -120,9 +120,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	co.Status.RelatedObjects = related
 
-	dnsAvailable := checkDNSAvailable(&state.DNS)
+	dnsAvailable := checkDNSStatusCondition(&state.DNS, operatorv1.OperatorStatusTypeAvailable)
+	dnsProgressing := checkDNSStatusCondition(&state.DNS, operatorv1.OperatorStatusTypeProgressing)
 
-	co.Status.Versions = r.computeOperatorStatusVersions(oldStatus.Versions, dnsAvailable)
+	co.Status.Versions = r.computeOperatorStatusVersions(oldStatus.Versions, dnsProgressing)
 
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -208,10 +209,10 @@ func (r *reconciler) getOperatorState(nsName string) (operatorState, error) {
 }
 
 // computeOperatorStatusVersions computes the operator's current versions.
-func (r *reconciler) computeOperatorStatusVersions(oldVersions []configv1.OperandVersion, allDNSESAvailable bool) []configv1.OperandVersion {
+func (r *reconciler) computeOperatorStatusVersions(oldVersions []configv1.OperandVersion, dnsProgressing bool) []configv1.OperandVersion {
 	// We need to report old version until the operator fully transitions to the new version.
 	// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version-reporting-during-an-upgrade
-	if !allDNSESAvailable {
+	if dnsProgressing {
```
wking, line 215: Hmm.  If, say, one compute node is down for maintenance, might that block the DaemonSet going Progressing=False?  I'd rather not have the update overall block on that sort of peripheral issue.  On the other hand, we don't want to leave dangling pods with old logic for too long after we've claimed a successful update.  I'm not sure what an elegant situation is for that; might need to happen at the DaemonSet controller level?  Seems like overkill to split into control-plane and compute DaemonSets...

PR #219, opened 2020-12-03
[release-4.6] Bug 1903887: Set DNS DaemonSet's maxUnavailable value to 10%

Miciah:
Enable rolling updates to scale linearly with the number of nodes by setting the maxUnavailable value to a percentage instead of the default value of 1.

* `assets/dns/daemonset.yaml`: Set `maxUnavailable` to 10%.
* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check if the update strategy changed and update it if it did.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Add test case to verify that `daemonsetConfigChanged` detects updates to `spec.updateStrategy`.
* `pkg/operator/controller/dns_status.go` (`getIntOrPercentValueSafely`, `getScaledValueFromIntOrPercent`): New functions based on definitions in Kubernetes 1.20.
(`computeDNSDegradedCondition`): Handle percentages values for `maxUnavailable`.
* `pkg/operator/controller/dns_status_test.go` (`TestComputeDNSDegradedCondition`): New test.

----

Manual cherry-pick of #217 and #221.


pkg/operator/controller/dns_status.go
```diff
@@ -64,7 +65,14 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 		Type: operatorv1.OperatorStatusTypeDegraded,
 	}
 	numberUnavailable := ds.Status.DesiredNumberScheduled - ds.Status.NumberAvailable
+	// TODO: Replace GetValueFromIntOrPercent with
+	// GetScaledValueFromIntOrPercent after rebasing on Kubernetes 1.20.
```
sgreene570, line 69: These TODO's won't apply to OCP 4.6. Should they be removed, or am I being too picky?

Miciah: Interesting question.  ¯\\_(ツ)\_\/¯

We wouldn't worry about that sort of thing in an automatic cherry-pick, but since this is a manual cherry-pick, we can tailor the change to the release branch.  I'll keep the test as a comment but replace the to-do with an explanation of why it is not enabled on 4.6.  Does that seem reasonable?

Miciah: Maybe I should just copy `GetScaledValueFromIntOrPercent` into `dns_status.go` for the backport.  I'm wishing now I had done that in #217, and then to-do for 4.7 would be just to delete the copy and use the original after the rebase.

sgreene570: No worries. Copying over the necessary functions into 4.6 sounds good to me. We can just worry about the rebase in 4.7 now. :grin:

sgreene570:
/retitle [release-4.6] Bug 1903887: Set DNS DaemonSet's maxUnavailable value to 10%

lihongan:
/lgtm
(see also https://bugzilla.redhat.com/show_bug.cgi?id=1903887#c2)

PR #223, opened 2020-12-11
Bug 1882485: Prevent dig errors from corrupting host's /etc/hosts

rfredette:
It's possible for `dig` to print error messages and still succeed, such as
if retries are attempted. If that happens, each word of the error
message is treated as a separate IP for the service that's being
resolved, causing /etc/hosts to be updated with garbage data.

This PR does 2 things:
- filters out error messages from `dig`, which always begin the line
  with ';' (usually with ';;', but some messages only have one
  semicolon)
- sets maximum retries to 0 when using tcp for dns, so that the behavior
  matches that of udp


Miciah:
Thanks!
/lgtm

Miciah:
Looks like we're seeing [BZ#1886922](https://bugzilla.redhat.com/show_bug.cgi?id=1886922).
/retest

PR #226, opened 2020-12-18
Bug 1908891: test/e2e: Block on TestCoreDNSImageUpgrade image revert

sgreene570:
**Block on TestCoreDNSImageUpgrade image revert**

TestCoreDNSImageUpgrade: Ensure that the CoreDNS image change for default DNS is completely reverted before moving onto the next test.

TestDNSForwarding: Ensure that DNS pods are all available before verifying Corefile contents from each DNS pod. Also, log pod status if a given pod's Corefile doesn't meet the test's expectations. 

This commit enhances the DNS operator tests to resolve BZ#1908891, in which TestDNSForwarding is noted as very flakey due to TestDNSForwarding's image revert rollout not blocking the premature execution of TestDNSFowarding.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -177,6 +177,9 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
 }
 
 func TestCoreDNSImageUpgrade(t *testing.T) {
+	// TODO: Revise this test and remove skip.
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1912522
+	t.Skip("skipping TestCoreDNSImageUpgrade as this test is fundamentally broken")
 	cl, err := getClient()
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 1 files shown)


test/e2e/operator_test.go
```diff
@@ -263,6 +252,30 @@ func setImage(deployment *appsv1.Deployment, image string) {
...
+			for _, container := range pod.Spec.Containers {
+				if container.Name == "dns" {
+					if container.Image == expectedImage {
+						return true, nil
+					}
+					break
+				}
+			}
+		}
+		return false, nil
```
Miciah, line 273: We want to verify that every pod has the expected image, but it looks like this logic only makes sure that some pod does.  We need to invert the logic as follows:  
```suggestion
		for _, pod := range podList.Items {
			for _, container := range pod.Spec.Containers {
				if container.Name == "dns" {
					if container.Image != expectedImage {
						return false, nil
					}
					break
				}
			}
		}
		return true, nil
```

sgreene570: huh, nice catch!

sgreene570: fixed.

Miciah:
It worked!  It's failing on `TestCoreDNSImageUpgrade` instead of `TestDNSForwarding` now.  * grin *.

Miciah:
Maybe need to increase the timeout in `TestCoreDNSImageUpgrade`.

sgreene570:
bumped the timeout in `checkCurrentDNSImage` to 7 minutes (~1 min per node in a 6 node cluster). Hopefully that works out.

sgreene570:
@Miciah For the operator e2e test, could it be the case that the cluster version operator is reverting changes made to the CoreDNS Image env variable by the operator test's `setImage` function, thus never allowing all pods in the daemonset to rollout with the desired image? Do we need to rethink the `TestCoreDNSImageUpgrade` test (Im assuming we dont want to broadly disable the CVO during our operator tests)?

Miciah:
> @Miciah For the operator e2e test, could it be the case that the cluster version operator is reverting changes made to the CoreDNS Image env variable by the operator test's `setImage` function, thus never allowing all pods in the daemonset to rollout with the desired image? Do we need to rethink the `TestCoreDNSImageUpgrade` test (Im assuming we dont want to broadly disable the CVO during our operator tests)?

I think you're right, and the test needs fundamental changes.  For now, I'd suggest adding `t.Skip` to the test.  Long-term, we can either add an override on the clusterversion or refactor the test to limit the scope of what is tested or use mocks.

sgreene570:
> For now, I'd suggest adding `t.Skip` to the test.

Done. I also created a bugzilla to track this fix (and mentioned it in code comments above the skip statement).

PR #230, opened 2021-01-15
Bug 1912522: test/e2e: Re-enable TestCoreDNSImageUprade

sgreene570:
**test/e2e/operator_test.go:** 

Re-enable TestCoreDNSImageUpgrade by adding a
cluster version override for the DNS operator deployment. This override
allows TestCoreDNSImageUpgrade to safely modify the DNS deployment
(specifically the operator's container image field) without contention
from the CVO. The cluster version override is removed at the end of the
test via a `defer` block.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -177,13 +177,44 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
 }
 
 func TestCoreDNSImageUpgrade(t *testing.T) {
-	// TODO: Revise this test and remove skip.
-	// See https://bugzilla.redhat.com/show_bug.cgi?id=1912522
-	t.Skip("skipping TestCoreDNSImageUpgrade as this test is fundamentally broken")
 	cl, err := getClient()
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Override the cluster version to temporarily disable CVO updates to the DNS operator deployment.
+	deplOverride := configv1.ComponentOverride{
+		Kind:      "Deployment",
+		Group:     "apps/v1",
+		Namespace: "openshift-dns-operator",
+		Name:      "dns-operator",
+		Unmanaged: true,
+	}
+
+	clusterVersion := &configv1.ClusterVersion{}
+	cvName := types.NamespacedName{Namespace: "default", Name: "version"}
+
+	if err := cl.Get(context.TODO(), cvName, clusterVersion); err != nil {
+		t.Fatalf("failed to get cluster version %s/%s: %v", cvName.Namespace, cvName.Name, err)
+	}
+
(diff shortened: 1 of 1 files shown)


sgreene570:
Let's make sure the first pass wasn't a fluke.
/test e2e-aws-operator
/test e2e-aws

test/e2e/operator_test.go
```diff
@@ -177,17 +177,48 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
...
+	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{deplOverride}
+
+	if err := cl.Update(context.TODO(), clusterVersion); err != nil {
+		t.Logf("failed to update cluster version %s/%s: %v", cvName.Namespace, cvName.Name, err)
+	}
+
+	// Remove cluster version override once the test returns.
+	defer func() {
+		if err := cl.Get(context.TODO(), cvName, clusterVersion); err != nil {
+			t.Logf("failed to get cluster version %s/%s: %v", cvName.Namespace, cvName.Name, err)
```
frobware, line 209: I really think we should fail harder.  Can the next test continue if this fails?

sgreene570: Arguably yes, the following tests _could_ run, but it's probably best that we ensure cluster version is in a clean state first.

sgreene570: fixed.

sgreene570: Also made the prior cluster version override API calls hard fail the test if they fail. Thanks!

test/e2e/operator_test.go
```diff
@@ -255,7 +286,7 @@ func setImage(deployment *appsv1.Deployment, image string) {
 }
 
 func checkCurrentDNSImage(t *testing.T, cl client.Client, expectedImage string) error {
-	err := wait.PollImmediate(1*time.Second, 7*time.Minute, func() (bool, error) {
+	err := wait.PollImmediate(1*time.Second, 10*time.Minute, func() (bool, error) {
```
frobware, line 289: What was the motivation for both the timeout changes? CI runs? other?

sgreene570: I observed `checkCurrentDNSImage` taking close to the 7 minute timeout when I ran the test on my own cluster. Just added some cushion. Dropped the 5 minute timeout for fetching the deployment since that just seems excessive. 

I can revert these changes if you'd prefer.

frobware: > I can revert these changes if you'd prefer.

If they're currently OK and passing in CI I'd be inclined to revert.

sgreene570: reverted.

PR #240, opened 2021-03-02
Bug 1933761: Set CoreDNS's cache's maximum TTL to 900 seconds

Miciah:
Before this change, the operator configured CoreDNS's cache plugin with a maximum TTL of 30 seconds, meaning cache entries expired after 30 seconds (or sooner, depending on the TTL from the upstream resolver).  This change increases the maximum TTL to 900 seconds to reduce load on upstream resolvers.

* `pkg/operator/controller/controller_dns_configmap.go` (`corefileTemplate`): Configure the cache plugin with a 900-second maximum TTL.
* `pkg/operator/controller/controller_dns_configmap_test.go` (`TestDesiredDNSConfigmap`): Update to expect the new value.

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -41,7 +41,7 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
     forward . /etc/resolv.conf {
         policy sequential
     }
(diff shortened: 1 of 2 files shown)


Miciah:
Thanks, openshift-cherrypick-robot!  I guess you just needed a reminder.  * grin *.  I appreciate your efforts!

PR #241, opened 2021-03-04
Bug 1934905: Enable errors plugin for custom upstream resolvers

Miciah:
Before this change, the operator did not enable the "errors" plugin in server blocks for custom upstream resolvers.  As a consequence, failures from upstream resolvers were not logged.  This change enables the errors plugin for all server blocks.

* `pkg/operator/controller/controller_dns_configmap.go` (`corefileTemplate`): Add the errors plugin to server blocks for custom upstream resolvers.
* `pkg/operator/controller/controller_dns_configmap_test.go` (`TestDesiredDNSConfigmap`): Update to expect the errors plugin in all server blocks.


sgreene570:
Nice find!
/lgtm

PR #234, opened 2021-02-11
Bug 1919737: Set CoreDNS readiness probe period and timeout each to 3 seconds

Miciah:
#### Set CoreDNS readiness probe period and timeout to 3 seconds

Change the readiness probe for CoreDNS from using a 10-second period and timeout to using a 3-second period and timeout in order to reduce the time to remove the endpoint if CoreDNS becomes unresponsive.

* `assets/dns/daemonset.yaml`: Set readiness probe period and timeout to 3 seconds.
* `pkg/manifests/bindata.go`: Regenerate.

#### Reconcile all readiness probe parameters

Check all parameters of the CoreDNS daemonset's readiness probe, and update the daemonset if any of the probe's parameters differ from the expected values.

Before this commit, the DNS controller only reconciled the readiness probe's handler and ignored changes to other parameters of the probe.

* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check if any of the readiness probe's parameters changed and update them if they did.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Add test case to verify that `daemonsetConfigChanged` detects changes to the readiness probe's `periodSeconds` parameter.

PR #242, opened 2021-03-04
Bug 1935297: Set CoreDNS readiness probe period and timeout each to 3 seconds

Miciah:
This is a manual cherry-pick of #234.  This cherry-pick also includes parts of #205 that #213 did not revert and that #234 builds upon.

#### Set CoreDNS readiness probe period and timeout to 3

Change the readiness probe for CoreDNS from using a 10-second period and timeout to using a 3-second period and timeout in order to reduce the time to remove the endpoint if CoreDNS becomes unresponsive.

* `assets/dns/daemonset.yaml`: Set readiness probe period and timeout to 3 seconds.

#### Reconcile all readiness probe parameters

Check all parameters of the CoreDNS daemonset's readiness probe, and update the daemonset if any of the probe's parameters differ from the expected values.

Before this commit, the DNS controller ignored all changes to the readiness probe's parameters.

* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check if any of the readiness probe's parameters changed and update them if they did.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Add test case to verify that `daemonsetConfigChanged` detects changes to the readiness probe's `path` or `periodSeconds` parameters.


knobunc (approved):
/lgtm

PR #243, opened 2021-03-04
Bug 1936022: Fix spurious reconciliation of DNS daemonset and service

Miciah:
#### Log diffs when reconciling resources

This PR improves log output when reconciling managed resources to log the diff between the current and updated resources when performing an update.

* `pkg/operator/controller/controller_cluster_role.go` (`updateDNSClusterRole`):
* `pkg/operator/controller/controller_dns_daemonset.go` (`updateDNSDaemonSet`):
* `pkg/operator/controller/controller_dns_service.go` (`updateDNSService`):
* `pkg/operator/controller/controller_service_monitor.go` (`updateDNSServiceMonitor`): Log the diff between the current and updated resources when performing an update.

#### `daemonsetConfigChanged`: Fix for `terminationGracePeriodSeconds` defaulting

When comparing daemonsets to determine whether an update is required, treat implicit and explicit default values for `terminationGracePeriodSeconds` as equal.

Before this PR, the update logic would keep trying to revert the default value that the API set for the daemonset's `terminationGracePeriodSeconds` field.

Follow-up to #205, which added the logic which reconciles `terminationGracePeriodSeconds`.

* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Use the new `cmpTerminationGracePeriodSeconds` function to compare `terminationGracePeriodSeconds` values so that two daemonsets are considered equal if the only difference between them is that one does not specify `terminationGracePeriodSeconds` and the other daemonset has the default value.
(`cmpTerminationGracePeriodSeconds`): New function.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDaemonsetConfigChanged`): Verify that `daemonsetConfigChanged` returns false if the only mutation is that `terminationGracePeriodSeconds` has been updated from empty to the default value.

#### `serviceChanged`: Fix for `clusterIPs` defaulting

Ignore the `clusterIPs` field when comparing services to determine whether an update is required.

Before this PR, the update logic would keep trying to revert the value that the API set for the service's `clusterIPs` field.

The `clusterIPs` field is new in Kubernetes 1.20 (OpenShift 4.7).

* `pkg/operator/controller/controller_dns_service.go` (`serviceChanged`): Ignore the services' `clusterIPs` field.
* `pkg/operator/controller/controller_dns_service_test.go` (`TestDNSServiceChanged`): Verify that `serviceChanged` returns false if the only mutation is that the `clusterIPs` field's value has changed.

### DIFF
--- a/pkg/operator/controller/controller_dns_daemonset_test.go
+++ b/pkg/operator/controller/controller_dns_daemonset_test.go
@@ -171,7 +171,7 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 		{
 			description: "if the config-volume default mode value is defaulted",
 			mutate: func(daemonset *appsv1.DaemonSet) {
-				newVal := volumeDefaultMode
+				newVal := corev1.ConfigMapVolumeSourceDefaultMode
 				daemonset.Spec.Template.Spec.Volumes[0].ConfigMap.DefaultMode = &newVal
 			},
 			expect: false,
@@ -187,7 +187,7 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 		{
 			description: "if the metrics-tls default mode value is defaulted",
 			mutate: func(daemonset *appsv1.DaemonSet) {
-				newVal := volumeDefaultMode
+				newVal := corev1.SecretVolumeSourceDefaultMode
 				daemonset.Spec.Template.Spec.Volumes[2].Secret.DefaultMode = &newVal
 			},
 			expect: false,
--- a/pkg/operator/controller/controller_dns_daemonset_test.go
+++ b/pkg/operator/controller/controller_dns_daemonset_test.go
@@ -215,6 +215,14 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 			expect: true,
 		},
 		{
+			description: "if the termination grace period is defaulted",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				newVal := int64(corev1.DefaultTerminationGracePeriodSeconds)
+				daemonset.Spec.Template.Spec.TerminationGracePeriodSeconds = &newVal
+			},
+			expect: false,
+		},
+		{
 			description: "if the termination grace period changes",
 			mutate: func(daemonset *appsv1.DaemonSet) {
 				sixty := int64(60)
--- a/pkg/operator/controller/controller_cluster_role.go
+++ b/pkg/operator/controller/controller_cluster_role.go
@@ -62,10 +62,12 @@ func (r *reconciler) updateDNSClusterRole(current, desired *rbacv1.ClusterRole) 
 		return false, nil
 	}
 
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
 	if err := r.client.Update(context.TODO(), updated); err != nil {
 		return false, fmt.Errorf("failed to update dns cluster role %s/%s: %v", updated.Namespace, updated.Name, err)
 	}
-	logrus.Infof("updated dns cluster role: %s/%s", updated.Namespace, updated.Name)
+	logrus.Infof("updated dns cluster role %s/%s: %v", updated.Namespace, updated.Name, diff)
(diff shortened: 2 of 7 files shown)


sgreene570:
This is fantastic! Should we look into doing something similar for the ingress operator?

sgreene570:
Looking at DNS operator logs with these changes 
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/243/pull-ci-openshift-cluster-dns-operator-master-e2e-aws/1367511193526210560/artifacts/e2e-aws/gather-extra/artifacts/pods/openshift-dns-operator_dns-operator-5cb7fcdbbc-rz42g_dns-operator.log

There's an abundance of white-space in the diff outputs... anything we can do to help with that? (Or do we not really care?)

Miciah:
> This is fantastic! Should we look into doing something similar for the ingress operator?

Thanks!  Yeah, the ingress operator has similar diff output for deployments but not for other resources, and it would be useful to have it for all resources.  I will probably make a similar change in the ingress operator for the resources other than deployments.  

> Looking at DNS operator logs with these changes
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/243/pull-ci-openshift-cluster-dns-operator-master-e2e-aws/1367511193526210560/artifacts/e2e-aws/gather-extra/artifacts/pods/openshift-dns-operator_dns-operator-5cb7fcdbbc-rz42g_dns-operator.log
> 
> There's an abundance of white-space in the diff outputs... anything we can do to help with that? (Or do we not really care?)

I don't know a good way to improve the whitespace with our logging solution.  I used `sed -ne 's/.*msg=\(".*"\)$/\1/gp' | jq -r .` to extract and format the relevant information.

Miciah:
We should backport this to OpenShift 4.7, which added the `clusterIPs` field in the API and added the logic to reconcile `terminationGracePeriodSeconds` in the ingress operator.  
/cherry-pick release-4.7

sgreene570:
> I don't know a good way to improve the whitespace with our logging solution. I used `sed -ne 's/.*msg=\(".*"\)$/\1/gp' | jq -r .` to extract and format the relevant information.

Good enough for me :grin:

pkg/operator/controller/controller_dns_daemonset_test.go
```diff
@@ -214,6 +214,14 @@ func TestDaemonsetConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if the termination grace period is defaulted",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				thirty := int64(30)
```
sgreene570, line 220: Can we get the API defaulted value right from the `corev1` API? 
(Not that I anticipate the value changing anytime soon, but IMO using the API constant would make the test cleaner)

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2601

sgreene570:
> This PR improves log output when reconciling managed resources to log the diff between the current and updated resources when performing an update.
> 
> * `pkg/operator/controller/controller_cluster_role.go` (`updateDNSClusterRole`):
> * `pkg/operator/controller/controller_dns_daemonset.go` (`updateDNSDaemonSet`):
> * `pkg/operator/controller/controller_dns_service.go` (`updateDNSService`):
> * `pkg/operator/controller/controller_service_monitor.go` (`updateDNSServiceMonitor`): Log the diff between the current and updated resources when performing an update.

@Miciah any reason not to also print diffs for corefile config map updates?

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: Nifty!  Thanks for the pointer!  Done.

Miciah:
> @Miciah any reason to not also print diffs for corefile config map updates?

Nope!  Done.  Thanks!

PR #244, opened 2021-03-06
Give CoreDNS access to endpointslices

Miciah:
Allow CoreDNS to read endpointslices.  CoreDNS 1.8.1 switched from reading endpoints to reading endpointslices.

* `assets/dns/cluster-role.yaml`:
* `manifests/0000_70_dns-operator_00-cluster-role.yaml`: Allow access to endpointslices.

---

This change is required because https://github.com/openshift/coredns/pull/52 bumped OpenShift from CoreDNS 1.6.6 to CoreDNS 1.8.1.


smarterclayton:
Force merging, these errors are unrelated.

PR #257, opened 2021-04-06
Bump openshift/api, controller runtime, and kube component versions

sgreene570:
**go.mod: Bump controller-runtime to v0.8.3**

**Support controller runtime v0.8.3**

pkg/operator/controller/controller.go:
Use new context parameter in `r.Reconcile(...)`.

pkg/operator/controller/status/controller.go:
Use new context parameter in `r.Reconcile(...)`.

pkg/operator/operator.go:
Add custom manager.ClientBuilder type definition with custom methods to
specifically side-step the default client cache. Use a context in
`o.Start`, and replace the existing stop channel with `ctx.Done()`.

cmd/dns-operator/main.go:
Update controller-runtime signals package import.

**go.mod: Bump for kube 1.21 & controller-runtime v0.9.0-alpha** 

This commit consists of the following changes:

* Bump openshift/api to commit `47be537` (current latest commit).
  openshift/api commit `2447417` moved openshift/api to kube 1.21,
  which inspired this commit. This commit is not strictly necessary
  at the moment, but will save us time when we need to pull in a new
  relevant API change in the future.

* Bump controller runtime to v0.9.0-aplha.1. Controller runtime v0.8.x
  is currently incompatible with kubernetes 1.21.

* Fix a breaking change in `pkg/operator/operator.go` introduced in
  controller runtime v0.9.0. Luckily this breaking change simplifies the
  manager workaround used for the version v0.8.3 bump.

---

This PR is similar to some of the commits in https://github.com/openshift/cluster-ingress-operator/pull/589


pkg/operator/controller/status/controller.go
```diff
@@ -79,14 +79,14 @@ func New(mgr manager.Manager, config operatorconfig.Config) (controller.Controll
 
 // Reconcile computes the operator's current status and therefrom creates or
 // updates the ClusterOperator resource for the operator.
-func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	nsManifest := manifests.DNSNamespace()
 
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: operatorcontroller.DNSClusterOperatorName().Name}}
 	if err := r.client.Get(context.TODO(), operatorcontroller.DNSClusterOperatorName(), co); err != nil {
```
Miciah, line 86: Missed one!  We can get this in a follow-up.

sgreene570: Darn!

sgreene570: This didnt make FF so might as well get it now.

PR #258, opened 2021-04-07
Bug 1946893: update inconsistent error message for AvailableCondition

lihongan:
use same error message for Degraded/Progressing/Available conditions if len(clusterIP) == 0


Miciah:
Thanks!
/lgtm

PR #260, opened 2021-04-08
Allow dns operator to be disabled with managementState field

rfredette:
this change allows resources normally managed by the dns operator to be left unmanaged when `spec.managementState` is set to `Unmanaged`. This will make debugging issues with dns resources easier

Also dependent on openshift/api PR [#888](https://github.com/openshift/api/pull/888) for API change


Miciah:
This is an excellent start!  Two things:

* You are adding logic to report the `Upgradeable` status condition on the dns object (`dnses.operator.openshift.io/default`), and we need the clusteroperator object (`clusteroperators.config.openshift.io/dns`) to report the `Upgradeable` status condition.  The clusteroperator's status is computed by the status controller under `pkg/operator/controller/status/controller.go`.  It can use status condition that you added to the dns object to compute the status condition on the clusteroperator object (basically just copy it from the dns to the clusteroperator).
* We need to make sure that we handle an empty value for `managementState`.  Please add a unit test `TestComputeDNSProgressingCondition` in `pkg/operator/controller/dns_status_test.go` to verify that your new `computeDNSProgressingCondition` function properly handles this case.

Miciah:
Sorry, I meant `computeDNSUpgradeableCondition`.  Replace "Progressing" with "Upgradeable" in my previous comment.

rfredette:
Updating `github.com/openshift/build-machinery-go` to `v0.0.0-20210712174854-1bb7fd1518d3` is not a manual update; it is required in the `github.com/openshift/api` version update.

pkg/operator/controller/status/controller.go
```diff
@@ -341,6 +342,32 @@ func checkDNSAvailable(dns *operatorv1.DNS) bool {
...
 }
 
+// computeOperatorUpgradeableCondition computes the operator's current
+// Upgradeable status state by examining the Upgradeable status of the DNS
+// operator
+func computeOperatorUpgradeableCondition(dns *operatorv1.DNS) configv1.ClusterOperatorStatusCondition {
+	var upgradeable bool
+	for _, cond := range dns.Status.Conditions {
+		if cond.Type == operatorv1.OperatorStatusTypeUpgradeable && cond.Status == operatorv1.ConditionTrue {
+			upgradeable = true
```
Miciah, line 352: Probably better to set `Upgradeable=True` if the DNS doesn't have an `Upgradeable` status condition:
```suggestion
	upgradeable := true
	for _, cond := range dns.Status.Conditions {
		if cond.Type == operatorv1.OperatorStatusTypeUpgradeable {
			upgradeable = cond.Status == operatorv1.ConditionTrue
```

pkg/operator/controller/status/controller.go
```diff
@@ -341,6 +342,32 @@ func checkDNSAvailable(dns *operatorv1.DNS) bool {
...
+		}
+	}
+	if !upgradeable {
+		return configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.OperatorUpgradeable,
+			Status:  configv1.ConditionFalse,
+			Reason:  "DNSNotUpgradeable",
+			Message: fmt.Sprintf("DNS %s is not upgradeable", dns.Name),
```
Miciah, line 360: It might be useful to include `cond.Message` from the DNS's status condition in the clusteroperator's status condition.

pkg/operator/controller/dns_status_test.go
```diff
@@ -19,46 +19,72 @@ func TestDNSStatusConditions(t *testing.T) {
 		availDNS, desireDNS int32
 		haveNR              bool
 		availNR, desireNR   int32
+		isManaged           bool
```
Miciah, line 22: Besides `Managed` or `Unmanaged`, `ManagementState` also could be `Force`, `Removed`, or unset.

pkg/operator/controller/dns_status_test.go
```diff
@@ -19,46 +19,72 @@ func TestDNSStatusConditions(t *testing.T) {
...
+		{testIn{true, true, 0, 2, true, 1, 2, false}, testOut{true, true, false, false}},
+		{testIn{true, true, 1, 2, true, 0, 2, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 1, 2, true, 1, 2, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 1, 2, true, 2, 2, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 2, 2, true, 0, 2, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 2, 2, true, 2, 2, false}, testOut{false, false, true, false}},
+		{testIn{true, true, 1, 3, true, 3, 3, false}, testOut{true, true, true, false}},
+		{testIn{true, true, 3, 3, true, 0, 3, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 2, 3, true, 3, 3, false}, testOut{false, true, true, false}},
+		{testIn{true, true, 0, 1, true, 0, 1, false}, testOut{true, true, false, false}},
```
Miciah, line 80: We don't necessarily need to test every combination; a few test cases where `ManagementState` is not `Managed` should be fine.

pkg/operator/controller/controller.go
```diff
@@ -111,38 +111,57 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-		}
-
-		if dns.DeletionTimestamp != nil {
-			// Handle deletion.
-			if err := r.ensureOpenshiftExternalNameServiceDeleted(); err != nil {
-				errs = append(errs, fmt.Errorf("failed to delete external name for openshift service: %v", err))
-			}
-			if err := r.ensureDNSDeleted(dns); err != nil {
-				errs = append(errs, fmt.Errorf("failed to ensure deletion for dns %s: %v", dns.Name, err))
+		if dns.Spec.ManagementState != operatorv1.Unmanaged {
```
Miciah, line 114: What do you think about switching the `if` and `else` so the simpler case (`dns.Spec.ManagementState == operatorv1.Unmanaged`) comes first?  It might also make sense to use a `switch` statement (even if we only have two cases for now, `case operatorv1.Unmanaged:` and `default:`).

Miciah (commented):
Overall looks good, just a few suggestions.

Miciah:
BTW, I usually squash commits and have just one commit with the vendor bump and `make generate` and a second commit with the implementation.

rfredette:
Thanks @Miciah. I agree with all your suggestions; the latest changes should have them all implemented.

pkg/operator/controller/dns_status_test.go
```diff
@@ -156,7 +176,18 @@ func TestDNSStatusConditions(t *testing.T) {
...
+		case operatorv1.Managed:
+			managementState = "Managed"
+		case operatorv1.Force:
+			managementState = "Force"
+		case operatorv1.Removed:
+			managementState = "Removed"
+		case operatorv1.Unmanaged:
+			managementState = "Unmanaged"
+		}
+		description := fmt.Sprintf("%s, %d/%d DNS pods available, %d/%d node-resolver pods available, managementState is %s", haveClusterIP, tc.inputs.availDNS, tc.inputs.desireDNS, tc.inputs.availNR, tc.inputs.desireNR, managementState)
```
Miciah, line 190: Can you use `tc.inputs.managementState` directly in the print statement?
```suggestion
		description := fmt.Sprintf("%s, %d/%d DNS pods available, %d/%d node-resolver pods available, managementState is %s", haveClusterIP, tc.inputs.availDNS, tc.inputs.desireDNS, tc.inputs.availNR, tc.inputs.desireNR, string(tc.inputs.managementState))
```
We can merge as-is leave that for a follow-up though.

Miciah:
Thanks!
/lgtm

PR #261, opened 2021-04-08
status: Watch clusteroperators

Miciah:
Add a watch on clusteroperators in the status controller so that it updates or recreates the dns clusteroperator as necessary should something else update or delete it.

* `pkg/operator/controller/status/controller.go` (`New`): Watch clusteroperators with a map function and predicate to reconcile `dnses.operator.openshift.io/default` if `clusteroperators.config.openshift.io/dns` is updated.
* test/e2e/operator_test.go (TestOperatorRecreatesItsClusterOperator): New test.  Verify that the operator recreates the "dns" clusteroperator if the clusteroperator is deleted.

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -60,7 +62,9 @@ type reconciler struct {
 // the logic for creating the ClusterOperator operator and updating its status.
 //
 // The controller watches DNS resources in the manager namespace and uses them
-// to compute the operator status.
+// to compute the operator status.  It also watches the clusteroperators
```
sgreene570, line 65:
```suggestion
// to compute the operator status.  It also watches the clusteroperator
```

Miciah:
Rebased.

PR #263, opened 2021-04-10
Bug 1948232: Fix for daemonset's maxSurge and service's ipFamilies and ipFamilyPolicy defaulting

Miciah:
#### `serviceChanged`: Fix for `ipFamilies` defaulting

Ignore the `ipFamilies` and `ipFamilyPolicy` fields when comparing services to determine whether an update is required.

Before this change, the update logic would keep trying to revert the value that the API set for the service's `ipFamilies` and `ipFamilyPolicy` fields.

These fields are enabled by the `IPv6DualStack` feature gate, which is enabled by default in Kubernetes 1.21 and OpenShift 4.8.

* `pkg/operator/controller/controller_dns_service.go` (`serviceChanged`): Ignore the `IPFamilies` and `IPFamilyPolicy` fields.
* `pkg/operator/controller/controller_dns_service_test.go` (`TestDNSServiceChanged`): Add test case to verify that updates to `.spec.ipFamilies` and `.spec.ipFamilyPolicy` are ignored.

#### Fix for daemonset `maxSurge` defaulting

Specify `spec.updateStrategy.rollingUpdate.maxSurge: 0` in the dns and node-resolver daemonsets.

Before this change, the operator left the field unspecified, and as a consequence, the update logic would keep trying to revert the default value that the API set for this field.  The default value for this field is 0, so specifying the value 0 has the result that the API does not need to update the daemonsets to fill in the default value, and so the operator's update logic is no longer triggered by the API's defaulting of this field.

In the future, we may want to set `maxSurge` to a positive value so as to ensure that DNS and node-resolver pods remain available locally on each node during rolling updates of the daemonsets.

* `assets/dns/daemonset.yaml`: Set `spec.updateStrategy.rollingUpdate.maxSurge` to 0.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go` (`ensureNodeResolverDaemonSet`): Set `spec.updateStrategy.rollingUpdate.maxSurge` to 0.

#### `updateDNSDaemonSet`: Skip update if no safe changes

If the dns daemonset update logic detects an unsafe update to the daemonset's node-placement parameters and no changes remain after removing the unsafe changes, do not attempt an update.

Before this change, the update logic could remove the unsafe part of an update and then perform an update with no changes, logging an empty diff.

* `pkg/operator/controller/controller_dns_daemonset.go` (`updateDNSDaemonSet`): Return without performing an update if no changes remain after ignoring unsafe changes.

#### `updateDNSDaemonSet`: Fix logging of ignored changes

When skipping an unsafe update to the node-placement parameters for the dns daemonset and logging a diff of the changes that are being ignored, show the ignored changes as additions to the update rather than as removals.

* `pkg/operator/controller/controller_dns_daemonset.go` (`updateDNSDaemonSet`): Swap `updated` and `ignored` when generating the diff of ignored changes.

#### node-resolver: Reconcile update strategy

Check if the node-resolver daemonset's `spec.updateStrategy` is changed and revert any detected change.

* `pkg/operator/controller/controller_dns_node_resolver_daemonset.go` (`nodeResolverDaemonSetConfigChanged`): Check if the current and expected daemonsets' `spec.updateStrategy` fields differ.  Set the updated daemonset's `spec.updateStrategy` field if a change is detected.
* `pkg/operator/controller/controller_dns_node_resolver_daemonset_test.go` (`TestNodeResolverDaemonsetConfigChanged`): Add a test case to verify that `nodeResolverDaemonSetConfigChanged` detects changes to the update strategy.


pkg/operator/controller/controller_dns_node_resolver_daemonset.go
```diff
@@ -157,6 +159,7 @@ func desiredNodeResolverDaemonSet(clusterIP, clusterDomain, openshiftCLIImage st
 			UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
 				Type: appsv1.RollingUpdateDaemonSetStrategyType,
 				RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+					MaxSurge:       &maxSurge,
```
sgreene570, line 162: Should `nodeResolverDaemonSetConfigChanged` check for changes to `Spec.UpdateStrategy`?

sgreene570:
Overall LGTM, modulo my one question.

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
Miciah: It should, but that may be outside the scope for this fix, and we don't have any need for it right now (except maybe to override users' changes in case users try to modify it).  Should fixing it be part of this PR?

sgreene570: Hm, good points.

One counter-point, however, is that the DNS daemonset is reconciled on updates to `Spec.UpdateStrategy`.
I think I am would be in favor of fixing this in this PR, or at the very least, adding a TODO comment. Thoughts?

Miciah: Fixed.

Miciah:
Rebased.

sgreene570:
Thanks for fixing this!
/lgtm

PR #269, opened 2021-04-26
Bug 1952174: status: Report old versions while progressing

Miciah:
Report the current `status.versions` value in the clusteroperator while the clusteroperator is reporting `Progressing=True` so that the clusteroperator doesn't report versions before it has finished upgrading to them.

* `pkg/operator/controller/status/controller.go` (`Reconcile`): Compute `newVersions` and pass it to `computeOperatorProgressingCondition` and `computeOperatorStatusVersions`.  Also pass the clusteroperator's newly computed `Progressing` status condition to `computeOperatorStatusVersions`.
(`computeOperatorStatusVersions`): Delete the `haveDNS` and `dns` parameters and the logic that checked the dns's `Progressing` status condition.  In their place, add a parameter for the clusteroperator's `Progressing` status condition, and use this status condition to decide whether to return the old versions or the updated versions.  Add a parameter for the updated versions.
* `pkg/operator/controller/status/controller_test.go` (`TestComputeOperatorStatusVersions`): Update unit tests.

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -115,21 +115,44 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 	co.Status.RelatedObjects = related
 
-	co.Status.Versions = r.computeOperatorStatusVersions(state.haveDNS, &state.dns, oldStatus.Versions)
+	newVersions := []configv1.OperandVersion{
+		{
+			Name:    OperatorVersionName,
+			Version: r.OperatorReleaseVersion,
+		},
+		{
+			Name:    CoreDNSVersionName,
+			Version: r.CoreDNSImage,
+		},
+		{
+			Name:    OpenshiftCLIVersionName,
+			Version: r.OpenshiftCLIImage,
+		},
+		{
+			Name:    KubeRBACProxyName,
+			Version: r.KubeRBACProxyImage,
+		},
+	}
 
+	operatorProgressingCondition := computeOperatorProgressingCondition(
+		state.haveDNS,
+		&state.dns,
+		oldStatus.Versions,
+		newVersions,
+		r.OperatorReleaseVersion,
+		r.CoreDNSImage,
+		r.OpenshiftCLIImage,
+		r.KubeRBACProxyImage,
+	)
 	co.Status.Conditions = mergeConditions(co.Status.Conditions,
 		computeOperatorAvailableCondition(state.haveDNS, &state.dns),
-		computeOperatorProgressingCondition(
-			state.haveDNS,
-			&state.dns,
-			oldStatus.Versions,
-			co.Status.Versions,
-			r.OperatorReleaseVersion,
-			r.CoreDNSImage,
-			r.OpenshiftCLIImage,
-			r.KubeRBACProxyImage,
-		),
+		operatorProgressingCondition,
 		computeOperatorDegradedCondition(state.haveDNS, &state.dns),
+	)
+	co.Status.Versions = r.computeOperatorStatusVersions(
+		&operatorProgressingCondition,
+		oldStatus.Versions,
+		newVersions,
 	)
 
 	if !operatorStatusesEqual(*oldStatus, co.Status) {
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -217,41 +240,15 @@ func (r *reconciler) getOperatorState() (operatorState, error) {
 }
 
 // computeOperatorStatusVersions computes the operator's current versions.
-func (r *reconciler) computeOperatorStatusVersions(haveDNS bool, dns *operatorv1.DNS, oldVersions []configv1.OperandVersion) []configv1.OperandVersion {
+func (r *reconciler) computeOperatorStatusVersions(operatorProgressingCondition *configv1.ClusterOperatorStatusCondition, oldVersions, newVersions []configv1.OperandVersion) []configv1.OperandVersion {
 	// We need to report old version until the operator fully transitions to the new version.
 	// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version-reporting-during-an-upgrade
-	if !haveDNS {
+	switch operatorProgressingCondition.Status {
+	case configv1.ConditionTrue, configv1.ConditionUnknown:
```
wking, line 247: I dunno what feeds you `Progressing` condition, but [the godocs say][1]:

> Progressing indicates that the operator is actively rolling out new code, propagating config changes, or otherwise moving from one steady state to another.  Operators should not report progressing when they are reconciling a previously known state.

So if the DNS operator has any config knobs, it would be possible to be `Progressing=True` after a config bump that was not an "update to new version" request.  This might delay the operator from leveling slightly longer than it needed to if someone bumped a config during an update.

Also, ideally the reported versions for coredns, openshift-cli, and kube-rbac-proxy would be bumped when they each completed leveling, instead of being held until the entire update had been completed and then bumped all at once.  

But I have no problem if any of that gets bumped to follow-up work, because delaying the version bumps slightly too long is much less of a problem than bumping versions too early.

[1]: https://github.com/openshift/api/blob/156ca80f8d830c353639afbf7a87f6546f2c6cff/config/v1/types_cluster_operator.go#L147-L151

Miciah: > I dunno what feeds you `Progressing` condition, but [the godocs say](https://github.com/openshift/api/blob/156ca80f8d830c353639afbf7a87f6546f2c6cff/config/v1/types_cluster_operator.go#L147-L151):
> 
> > Progressing indicates that the operator is actively rolling out new code, propagating config changes, or otherwise moving from one steady state to another.  Operators should not report progressing when they are reconciling a previously known state.
> 
> So if the DNS operator has any config knobs, it would be possible to be `Progressing=True` after a config bump that was not an "update to new version" request. This might delay the operator from leveling slightly longer than it needed to if someone bumped a config during an update.

The quoted text says that the operator should set `Progressing=True` if the operator is propagating config changes, and then you say the operator would be delaying "leveling slightly longer than it needed to" if it set `Progressing=True` after a config bump, right?  How is "bumping a config" for "config knobs" different from "propagating config changes"?  

To be concrete, the operator sets `Progressing=True` if the dns-default service does not have an address or if the dns-default daemonset (which runs coredns and kube-rbac-proxy) or the node-resolver daemonset (which uses openshift-cli) is not at level (e.g., the user configured a node selector or toleration, and the change has not yet been propagated to the daemonset, or the daemonset's `status.numberAvailable` is not equal to the daemonset's `status.desiredNumberScheduled`).  Is it wrong for the clusteroperator to report `Progressing=True` in some of these cases?  
 
Are you saying the user may have updated the release and then updated some config, and if the operator finishes rolling out the new images but is still updating the config, then the operator should report `Progressing=True` but still bump `status.versions`?  Hm, I guess the `status.versions` logic would instead of looking at the daemonsets need to look at each individual pod to make sure the desired number of pods both have the desired image and are available.  Is that what you had in mind?  

> Also, ideally the reported versions for coredns, openshift-cli, and kube-rbac-proxy would be bumped when they each completed leveling, instead of being held until the entire update had been completed and then bumped all at once.

Ah, that is not how I interpreted https://bugzilla.redhat.com/show_bug.cgi?id=1952174#c0:

    The DNS operator needs to delay the versions bump until the operands have leveled.  For example, from [2]:
    
      An operator reports a new "operator" version when it has rolled out the new version to all of its operands.

Is "operator" a special-case then?  That is, we bump each component _within_ `status.versions` _that corresponds to an operand_ as the operand is updated, and then bump the component for the _operator_ once all the operands are updated?

For what it's worth, we used to have one daemonset that deployed pods where each pod had a container for each operand image, so we would have bumped all versions together anyway.  Now we have two daemonests, where one uses the coredns and kube-rbac-proxy image and the other uses the openshift-cli image, so we could bump openshift-cli independently of the other two.  

> But I have no problem if any of that gets bumped to follow-up work, because delaying the version bumps slightly too long is much less of a problem than bumping versions too early.

Ack.  Let's do that, and iterate as we refine the definition of "progressing".

wking: > ... or the daemonset's status.numberAvailable is not equal to the daemonset's status.desiredNumberScheduled...

When that happens as a result of a `spec` bump, that's fine.  But if it is the result of disruption like "a node rebooted", I would expect it to fall into the `reconciling a previously known state` case which should not trigger `Progressing=True`.

> Hm, I guess the status.versions logic would instead of looking at the daemonsets need to look at each individual pod to make sure the desired number of pods both have the desired image and are available.

Yeah, that sounds right.  But again, minor nit.

> That is, we bump each component within status.versions that corresponds to an operand as the operand is updated, and then bump the component for the operator once all the operands are updated?

[`versions` godocs][1] only talk about `operator`, so forget I said anything about operand versions until I get some clarity landed in the godocs ;).

[1]: https://github.com/openshift/api/blob/156ca80f8d830c353639afbf7a87f6546f2c6cff/config/v1/types_cluster_operator.go#L43-L46

sgreene570:
Looks good! Comments will be addressed in a follow up, from my understanding. 
/lgtm

sgreene570:
Retest bot is on PTO I guess
/retest

PR #273, opened 2021-05-12
Bug 1939723: Don't check node-resolver status for DNS Degraded condition

candita:
DNS operator goes degraded when a machine is added and removed (in serial tests).  In 4.8 the node-resolver daemonset has been split out from the default dns daemonset.  Node-resolver lands on every node and tolerates taints but dns no longer does.  Both daemonsets still contribute to the Degraded status. Therefore node-resolver can land on nodes that aren't ready, and this ends up marking the whole operator as degraded.  The fix will calculate the Degraded status without checking the status of the node-resolver daemonset.

Also update the node resolver daemonset's MaxUnavailable to 33%, as the 10% metric is intended for traffic serving applications on all nodes.


candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 3 pods failed before test on SCC errors
Error creating: pods "cloud-credential-operator-5945794f94-" is forbidden: unable to validate against any security context constraint: provider "privileged": Forbidden: not usable by user or serviceaccount for ReplicaSet.apps/v1/cloud-credential-operator-5945794f94 -n openshift-cloud-credential-operator happened 6 times
Error creating: pods "ingress-operator-59fdbcf9bc-" is forbidden: unable to validate against any security context constraint: provider "privileged": Forbidden: not usable by user or serviceaccount for ReplicaSet.apps/v1/ingress-operator-59fdbcf9bc -n openshift-ingress-operator happened 7 times
Error creating: pods "marketplace-operator-59bfb69f7f-" is forbidden: unable to validate against any security context constraint: provider "privileged": Forbidden: not usable by user or serviceaccount for ReplicaSet.apps/v1/marketplace-operator-59bfb69f7f -n openshift-marketplace happened 8 times

/test e2e-upgrade

candita:
May 12 21:20:53.574: INFO: Running 'oc --namespace=e2e-test-verify-run-fs-cwtft --kubeconfig=/tmp/configfile453051164 start-build verify-run-fs -o=name'
May 12 21:20:54.083: INFO: 
start-build output with args [verify-run-fs -o=name]:
Error><nil>
StdOut>
build.build.openshift.io/verify-run-fs-1
StdErr>

/test e2e-aws

candita:
https://bugzilla.redhat.com/show_bug.cgi?id=1959564 was opened for the [sig-builds] verify filesystem failure

sgreene570:
Could you drop the 2 merge commits from this PR?

> Also update the MaxUnavailable to 33%, as the 10% metric is intended for traffic serving applications on all nodes.
Could you add a comment along these lines to the Daemonset yaml, if you don't mind?

sgreene570:
> Could you drop the 2 merge commits from this PR?

By the way, I think the easiest way to do this would be to run the following commands on the branch you are trying to merge from:
```
git reset HEAD~3 --hard
git pull upstream master (where `upstream` is the openshift/cluster-dns-operator remote)
git cherry-pick 65f3485a16d4c5e141f0736dca9be590f175128b
```

candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 24 pods failed before test on SCC errors

/test e2e-upgrade

candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 14 pods failed before test on SCC errors
Error creating: pods "cloud-credential-operator-75d8cdd488-" is forbidden: unable to validate against any security context constraint: provider "privileged": Forbidden: not usable by user or serviceaccount for ReplicaSet.apps/v1/cloud-credential-operator-75d8cdd488 -n openshift-cloud-credential-operator happened 6 times

/test e2e-upgrade

assets/dns/daemonset.yaml
```diff
@@ -89,4 +89,6 @@ spec:
       maxSurge: 0
       # Note: The daemon controller rounds the percentage up
       # (unlike the deployment controller, which rounds down).
-      maxUnavailable: 10%
+      # 10% metric is intended for traffic serving applications
+      # on all nodes.  33% is reasonable for DNS.
+      maxUnavailable: 33%
```
Miciah, line 94: The dns daemonset should be 10%, and the node-resolver daemonset (defined in `desiredNodeResolverDaemonSet`) should be 33%, right?

Miciah:
After this change, what is the signal to the cluster admin if there is a problem with the node-resolver pods?

assets/dns/daemonset.yaml
candita: I see I forgot it in the config for node resolver, and I'll add it there.  Should it be for both DNS and nodeResolver daemonsets?

candita:
> After this change, what is the signal to the cluster admin if there is a problem with the node-resolver pods?
IIrc, we discussed the fact that we don't have a signal until we have an operator that monitors the noderesolver daemonset.  Should there be something here in the interim?

sgreene570:
> > After this change, what is the signal to the cluster admin if there is a problem with the node-resolver pods?
> 
> IIrc, we discussed the fact that we don't have a signal until we have an operator that monitors the noderesolver daemonset. Should there be something here in the interim?

Are there generic alerts for daemonsets not reaching full availability? If not (or if existing alerts are not sufficient), could we construct an alert around the node resolver daemonset to give the cluster administrator some signal?

assets/dns/daemonset.yaml
Miciah: DNS should be 10% to minimize disruption (especially in case the cluster admin has configured DNS pods not to run on every node).

Miciah:
> Are there generic alerts for daemonsets not reaching full availability? If not (or if existing alerts are not sufficient), could we construct an alert around the node resolver daemonset to give cluster administrator some signal here?

This makes sense to me: a general alert, or if there isn't one, an alert specifically for node-resolver.

We can do this as a follow-up.

candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 21 pods failed before test on SCC errors 
/retest

candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 8 pods failed before test on SCC errors
/retest

candita:
fail [github.com/openshift/origin/test/extended/authorization/scc.go:73]: 28 pods failed before test on SCC errors 
/retest

candita:
https://github.com/openshift/cluster-dns-operator/pull/273#issuecomment-847990805 to be addressed in https://bugzilla.redhat.com/show_bug.cgi?id=1967963

candita:
This was not verified, and original bug went stale.

PR #290, opened 2021-08-26
Bug 1939723: use a grace period for requeuable degraded conditions

candita:
DNS goes degraded when the number of pods in the DNS daemonset reaches maxUnavailable, but we want to be a little more forgiving and allow a grace period to reach the recommended number of available pods.
- assets/dns/daemonset.yaml: change maxUnavailable to 33% here
- pkg/operator/controller/controller.go: observe retryable error from syncDNSStatus and ensureDNS in the Reconcile loop
- pkg/operator/controller/dns_status.go: produce retryable error from computeDNSStatusConditions, via changes to computeDNSStatusConditions, computeDNSDegradedCondition; in computeDNSDegradedCondition add gracePeriod logic to condition MaxUnavailableDNSPodsExceeded (but no other conditions)
- pkg/operator/controller/dns_status_test.go: clean up use of noderesolver in computerDNSDegradedCondition (no longer used to compute DNS Degraded); observe expectRequeue and a retry duration where needed


go.mod
```diff
@@ -7,8 +7,10 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/kevinburke/go-bindata v3.11.0+incompatible
-	github.com/openshift/api v0.0.0-20210719174558-55d730c80aa9
+	github.com/openshift/api v0.0.0-20210722202039-91e26cfb36fc
 	github.com/openshift/build-machinery-go v0.0.0-20210712174854-1bb7fd1518d3
+	github.com/openshift/cluster-ingress-operator v0.0.0-20210824175145-977aa63c247d
```
Miciah, line 12: Better to copy in the "github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror" package than import it from the operator.  (Best would be to put it in a separate library, but we can consider doing that as a follow-up.)

pkg/operator/controller/controller.go
```diff
@@ -165,7 +168,13 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 				// Handle everything else.
 				if err := r.ensureDNS(dns); err != nil {
-					errs = append(errs, fmt.Errorf("failed to ensure dns %s: %v", dns.Name, err))
+					switch e := err.(type) {
+					case retry.Error:
+						// Got retryable error; requeue
+						return reconcile.Result{RequeueAfter: e.After()}, nil
+					default:
+						errs = append(errs, fmt.Errorf("failed to ensure dns %s: %v", dns.Name, err))
+					}
```
Miciah, line 177: Would it be better to leave this logic unchanged and add the retry logic at the end of the `Reconcile` method?  Something like this:
```go
				if err := r.ensureDNS(dns); err != nil {
￼					errs = append(errs, fmt.Errorf("failed to ensure dns %s: %v", dns.Name, err))
				} else if err := r.ensureExternalNameForOpenshiftService(); err != nil {
					errs = append(errs, fmt.Errorf("failed to ensure external name for openshift service: %v", err))
				}
			}
		}

	if err := retry.NewMaybeRetryableAggregate(errs); err != nil {
		switch e := err.(type) {
		case retryable.Error:
			log.Error(e, "got retryable error; requeueing", "after", e.After())
			return reconcile.Result{RequeueAfter: e.After()}, nil
		default:
			// Log in case of errors as the controller's logs get eaten.
			logrus.Errorf("failed to reconcile request %s: %v", request, err)
			return reconcile.Result{}, err
		}
	}
	return reconcile.Result{}, nil
```￼

File: pkg/operator/controller/controller.go [hunk 3]
```diff
@@ -412,7 +421,13 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
...
+		switch err.(type) {
+		case retry.Error:
+			// Got retryable error; requeue
+			return retry.NewMaybeRetryableAggregate([]error{err})
+		default:
+			errs = append(errs, fmt.Errorf("failed to sync status of dns %q: %w", dns.Name, err))
+		}
 	}
 
 	return utilerrors.NewAggregate(errs)
```
Miciah, line 433: Would the following suffice?
```go
	if err := r.syncDNSStatus(dns, clusterIP, clusterDomain, haveDNSDaemonset, dnsDaemonset, haveNodeResolverDaemonset, nodeResolverDaemonset); err != nil {
		errs = append(errs, fmt.Errorf("failed to sync status of dns %q: %w", dns.Name, err))
	}

	return retry.NewMaybeRetryableAggregate(errs)
```
That is, keep the logic in `ensureDNS` the same as it was before this PR except for replacing the `NewAggregate` with `NewMaybeRetryableAggregate`.  This way, `ensureDNS` would behave the same as before if there were a non-retryable error but would return a retryable error if all errors were retryable.

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +170,110 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
 
-	if len(degradedReasons) != 0 {
-		degradedCondition.Status = status
-		degradedCondition.Reason = strings.Join(degradedReasons, "")
-		degradedCondition.Message = strings.Join(messages, "\n")
-	} else {
-		degradedCondition.Status = operatorv1.ConditionFalse
-		degradedCondition.Reason = "AsExpected"
-		degradedCondition.Message = "Enough DNS pods are available, and the DNS service has a cluster IP address."
+	graceConditions, checkedConditions, requeueAfter := checkConditions(expectedConditions, degradedConditions)
```
Miciah, line 188: Do I understand correctly that `degradedConditions` and `checkedConditions` are only used to compute the final `Degraded` status condition, and are never published in any API object?  If so, that's a clever approach to enable us to re-use the existing logic.  As a follow-up, we might want to consider actually publishing these status conditions, if that would be useful.

go.mod
candita: Done

pkg/operator/controller/controller.go
candita: Thanks, that's better!  Though I'll still need to `utilerrors.NewAggregate(errs)` in the default case instead of just `err`.  Done.

pkg/operator/controller/controller.go
candita: Thanks, done.

pkg/operator/controller/dns_status.go
candita: Before, we joined all the `degradedReasons` and put that in the `Condition.Reason`, so maybe I should keep that instead of setting the `Condition.Reason` to `DNSControllerDegradedConditionType`

candita:
`E0902 19:31:09.108818     123 reflector.go:138] k8s.io/client-go/tools/watch/informerwatcher.go:146: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: Get "https://api.ci-op-fc6zhy1k-230ac.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/kube-system/configmaps?fieldSelector=metadata.name%3Dbootstrap&resourceVersion=286": dial tcp 72.44.51.243:6443: connect: connection refused
`
/test e2e-upgrade

go.mod
```diff
@@ -7,8 +7,9 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/kevinburke/go-bindata v3.11.0+incompatible
-	github.com/openshift/api v0.0.0-20210719174558-55d730c80aa9
+	github.com/openshift/api v0.0.0-20210722202039-91e26cfb36fc
```
Miciah, line 10: The version bump presumably came from importing github.com/openshift/cluster-ingress-operator, and now that that import is gone, the version bump should be unnecessary.  We can keep it if you prefer, but it is a little odd and adds some unnecessary changes to the PR.

pkg/operator/controller/dns_status.go
```diff
@@ -52,34 +68,96 @@ func computeDNSStatusConditions(dns *operatorv1.DNS, clusterIP string, haveDNSDa
...
+		},
+		{
+			condition: DNSNoDNSPodsDesired,
+			status:    operatorv1.ConditionFalse,
+		},
+		{
+			condition:   DNSNoDNSPodsAvailable,
+			status:      operatorv1.ConditionFalse,
+			gracePeriod: time.Minute * 5,
+		},
```
Miciah, line 142: Does having a grace period make sense for `DNSNoDNSPodsAvailable`?  If no pods are available, that means DNS service is completely unavailable.

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +170,109 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+			degradedConditions = append(degradedConditions, operatorv1.OperatorCondition{Type: DNSNoDNSPodsDesired, Status: operatorv1.ConditionTrue, LastTransitionTime: now})
 			messages = append(messages, "No DNS pods are desired; this could mean all nodes are tainted or unschedulable.")
 		case have == 0:
-			status = operatorv1.ConditionTrue
-			degradedReasons = append(degradedReasons, "NoDNSPodsAvailable")
+			degradedConditions = append(degradedConditions, operatorv1.OperatorCondition{Type: DNSNoDNSPodsAvailable, Status: operatorv1.ConditionTrue, LastTransitionTime: now})
 			messages = append(messages, "No DNS pods are available.")
 		case intstrErr != nil:
-			degradedReasons = append(degradedReasons, "InvalidDNSMaxUnavailable")
+			keepConditionUnknown = true
```
Miciah, line 179: This logic has a subtle problem.  With the old logic, when we reach this case, `status` could be "True" if the `if len(clusterIP) == 0 {` check above were true, and otherwise `status` would have its initial value of `Unknown`; either way, this `case intstrErr != nil:` check would leave `status` set to the one of those two values (true or unknown) that it had before this check.  With the new logic, this `case intstrErr != nil:` check always sets `keepConditionUnknown = true` and thus causes the `if keepConditionUnknown {` check below to set `Status` to "Unknown", even if the `if len(clusterIP) == 0 {` check above was true.  

The idea in the original logic is that starting out, we don't know the status, so we set it to "unknown"; then we check for various possible problems; each check can determine that something is wrong and set the degraded status to "true"; and if a previous check has determined that something is wrong, a subsequent check cannot not set the state _back_ to "unknown" because we _do_ know (from the previous check) that _something_ is wrong.  Once we know something is wrong, there is no circumstance in which we can go back to saying it's unknown whether anything is wrong.

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +170,109 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
-		degradedCondition.Message = strings.Join(messages, "\n")
-	} else {
-		degradedCondition.Status = operatorv1.ConditionFalse
-		degradedCondition.Reason = "AsExpected"
-		degradedCondition.Message = "Enough DNS pods are available, and the DNS service has a cluster IP address."
+	graceConditions, checkedConditions, requeueAfter := checkConditions(expectedConditions, degradedConditions)
+
+	// graceConditions have grace periods configured in the expectedConditions, so if present, this should be requeued
+	if len(graceConditions) != 0 {
+		finalDegradedCondition.Status = operatorv1.ConditionTrue
```
Miciah, line 192: The point of the grace period is that we report `Degraded=False` during the grace period, so shouldn't this be `finalDegradedCondition.Status = operatorv1.ConditionFalse`?  Also, as `checkConditions` has any conditions without a  grace period, shouldn't we check it first so that we do report `Degraded=True` immediately if there are any problems that don't have a grace period?

pkg/operator/controller/dns_status.go
Miciah: The fact that this subtle problem wasn't caught by the unit tests suggests we need a test case for "no clusterIP, DNS invalid MaxUnavailable"; such a test case should expect `Degraded=True`, and if I'm reading the new logic correctly, it would fail.

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +170,109 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+		grace := formatConditions(graceConditions)
+
+		// If the oldCondition changed more recently than requeueAfter period of time, reduce the requeueAfter period of time
+		if oldCondition != nil && (!oldCondition.LastTransitionTime.IsZero()) {
+			t1 := now.Add(-requeueAfter)
+			t2 := oldCondition.LastTransitionTime
+			if t2.After(t1) {
+				requeueAfter = t2.Sub(t1)
+			}
+		}
```
Miciah, line 204: The purpose of this logic isn't clear to me.  `checkConditions` already checks each condition's last transition time...  But now I realize that the status conditions computed above all have "now" as their last transition time; is the logic here intended to adjust for that?  

I'm starting to think that it really would be simpler to put all the status conditions in the dns's actual status field so that we would have accurate last transition times.

pkg/operator/controller/dns_status.go
candita: This presumes that enough will come up eventually.  Maybe a much longer grace period?  I'm okay with not using a gracePeriod on this condition, but this was the only other condition I chose because it seemed like it could easily change from no DNS pods available to less than the number of DNS pods available, which is the only other grace period condition.

go.mod
candita: Ok, removed it.

candita:
Sep 21 21:16:18.541: FAIL: invalid configuration: [context was not found for specified context: default/api-ci-op-zxcrpvxx-bcd5b-origin-ci-int-aws-dev-rhcloud-com:6443/system:admin, cluster has no server defined]
/test e2e-aws

pkg/operator/controller/dns_status.go
candita: Me too - done

pkg/operator/controller/dns_status.go
candita: You're right,  I prioritized the grace period evaluation and should not have, and should not mark it as Degraded during the grace period.  Fixed.

pkg/operator/controller/dns_status.go
candita: I removed the grace period for this.  We can add it later if it makes sense.

pkg/operator/controller/dns_status.go
candita: Done

pkg/operator/controller/dns_status.go
candita: Done - fixed other test cases too, which used to look for ConditionUnknown

candita:
> level=info msg=Creating infrastructure resources...
> level=error
> level=error msg=Error: Provider produced inconsistent result after apply
> level=error
> level=error msg=When applying changes to module.vpc.aws_route_table.default[0], provider
> level=error msg="registry.terraform.io/-/aws" produced an unexpected new value for was
> level=error msg=present, but now absent.
> level=error
> level=error msg=This is a bug in the provider, which should be reported in the provider's own
> level=error msg=issue tracker.
> level=error
> level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change

/test e2e-aws-operator

pkg/operator/controller/dns_status.go
```diff
@@ -94,32 +176,116 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+			grace := formatConditions(graceConditions)
+
+			// If the oldCondition changed more recently than requeueAfter period of time, reduce the requeueAfter period of time
+			if oldCondition != nil && (!oldCondition.LastTransitionTime.IsZero()) {
+				t1 := now.Add(-requeueAfter)
+				t2 := oldCondition.LastTransitionTime
+				if t2.After(t1) {
+					requeueAfter = t2.Sub(t1)
+				}
+			}
```
Miciah, line 228: I still don't understand the purpose of this code, and I thought you were going to delete it and instead persist all the status conditions so that they would have meaningful, non-zero last transition times (see https://github.com/openshift/cluster-dns-operator/pull/290#discussion_r701569067).

pkg/operator/controller/dns_status_test.go
```diff
@@ -212,140 +223,184 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
 		},
+		{
+			name:          "not enough DNS pods available, check requeue",
+			clusterIP:     "172.30.0.10",
+			dnsDaemonset:  makeDaemonSet(10, 8, intstr.FromString("10%")),
+			oldCondition:  defaultOldCondition,
+			expected:      operatorv1.ConditionFalse,
+			expectAfter:   time.Minute * 5,
+			expectRequeue: true,
+		},
```
Miciah, line 349: This is covered by the "too few DNS pods available (percentage)" test case.

pkg/operator/controller/dns_status_test.go
```diff
@@ -212,140 +223,184 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
+			oldCondition:  cond(DNSControllerDegradedConditionType, operatorv1.ConditionTrue, "", clock.Now().Add(time.Minute*-6)),
+			expected:      operatorv1.ConditionTrue,
+			expectRequeue: false,
+		},
+		{
+			name:          "not enough DNS pods available for >5m",
+			clusterIP:     "172.30.0.10",
+			dnsDaemonset:  makeDaemonSet(10, 8, intstr.FromString("10%")),
+			oldCondition:  cond(DNSControllerDegradedConditionType, operatorv1.ConditionTrue, "", clock.Now().Add(time.Minute*-6)),
+			expected:      operatorv1.ConditionFalse,
```
Miciah, line 380: I thought the point of the original issue reported in Bugzilla and the point of this PR was that in the "not enough DNS pods available for <5m" case, the operator should report `Degraded=False`, and in the "not enough DNS pods available for >5m" case, the operator should report `Degraded=True`.  Why does this test case expect `Degraded=False`?
```suggestion
			expected:      operatorv1.ConditionTrue,
```

pkg/operator/controller/dns_status.go
candita: I removed the rest of it, but this is needed for when the request has already been requeued, and the condition changed less than 5m ago. In the unit test which has an oldCondition that last transitioned 4 minutes ago, I thought we would want to requeue it in 1 minute rather than wait another 5.  If that's not correct, and we want to requeue in 5 minutes no matter when the last transition occurred, I can change it.

pkg/operator/controller/dns_status_test.go
candita: In this case I"m explicitly checking for a requeue value.

pkg/operator/controller/dns_status_test.go
candita: Whenever it is going to requeue, it will set Degraded=False and set the requeue wait to at most 5 minutes.  It will always requeue if not enough DNS pods are available, no matter how long they've been unavailable.  All I'm checking in this test is that the requeue wait time is adjusted.  The point is to requeue and try to  recalculate the degraded condition while there aren't enough pods, reporting degraded=false in the meantime.  In the case that there are NO pods available, it's always degraded=true.

pkg/operator/controller/dns_status.go
```diff
@@ -52,34 +68,100 @@ func computeDNSStatusConditions(dns *operatorv1.DNS, clusterIP string, haveDNSDa
...
+	// These conditions are expected to be false
+	expectedConditions := []dnsCondition{
+		{
+			condition: DNSNoService,
+			status:    operatorv1.ConditionFalse,
+		},
+		{
+			condition: DNSControllerDegradedConditionType,
+			status:    operatorv1.ConditionFalse,
+		},
```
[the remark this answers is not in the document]
candita: Removed.

pkg/operator/controller/dns_status_test.go
Miciah: What do you mean?  Both "too few DNS pods available (percentage)" and "not enough DNS pods available, check requeue" have the following:
```go
			expectAfter:   time.Minute * 5,
			expectRequeue: true,
```
The two test cases only differ in the name and daemonset, namely one has the following:
```go
			dnsDaemonset:  makeDaemonSet(100, 89, intstr.FromString("10%")),
```
and the other has the following:
```go
			dnsDaemonset:  makeDaemonSet(10, 8, intstr.FromString("10%")),
```
which ultimately should both lead to the same degraded-but-with-grace outcome.

pkg/operator/controller/dns_status_test.go
Miciah: In a scenario in which some pods are available but not enough pods are available, the operator should allow for the grace period and then set `Degraded=True` when the grace period expires.  This test case looks like exactly that scenario and condition (not enough pods, grace period has expired).  Expecting a requeue with the appropriate interval is fine, but why do we expect `Degraded=False`?  Do we have a test case that verifies that `computeDNSDegradedCondition` sets `Degraded=True` when the grace period expires?

pkg/operator/controller/dns_status.go
Miciah: > I removed the rest of it, but this is needed for when the request has already been requeued, and the condition changed less than 5m ago. 

We really want to requeue after the remaining portion of the `MaxUnavailableDNSPodsExceeded` condition's grace period.  However, we won't have this condition's last transition time unless we persist it to the API, which is why I thought you were going to change this logic to publish all the status conditions.  

> In the unit test which has an oldCondition that last transitioned 4 minutes ago, I thought we would want to requeue it in 1 minute rather than wait another 5. If that's not correct, and we want to requeue in 5 minutes no matter when the last transition occurred, I can change it.

This part is fine; it's just a question of how we get the last transition time (specifically which condition's last transition time we use).

candita:
> Error creating: pods "console-operator-56965db7b8-" is forbidden: unable to validate against any security context constraint: [provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount] for ReplicaSet.apps/v1/console-operator-56965db7b8 -n openshift-console-operator happened 12 times


/test e2e-aws-serial

pkg/operator/controller/dns_status.go
candita: Removed all of it, PTAL when you get a chance @Miciah

pkg/operator/controller/dns_status_test.go
candita: Removed

pkg/operator/controller/dns_status_test.go
candita: Fixed, it now sets `Degraded=True` when not enough pods are available for > 5 min

pkg/operator/controller/dns_status_test.go
```diff
@@ -212,142 +225,204 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
+
+	defaultOldCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeDegraded,
+		Status: operatorv1.ConditionUnknown,
+	}
+
+	defaultOldGraceCondition := &operatorv1.OperatorCondition{
+		Type:   DNSMaxUnavailableDNSPodsExceeded,
+		Status: operatorv1.ConditionUnknown,
+	}
```
Miciah, line 237: The omission of `LastTransitionTime` is conspicuous, but the test works because both the fake clock as well as the last transition time of these status conditions are initialized with time 0.  It would be safer though to set the last transition times using the fake clock:
```suggestion
	defaultOldCondition := &operatorv1.OperatorCondition{
		Type:               operatorv1.OperatorStatusTypeDegraded,
		Status:             operatorv1.ConditionUnknown,
		LastTransitionTime: metav1.NewTime(clock.Now()),
	}

	defaultOldGraceCondition := &operatorv1.OperatorCondition{
		Type:               DNSMaxUnavailableDNSPodsExceeded,
		Status:             operatorv1.ConditionUnknown,
		LastTransitionTime: metav1.NewTime(clock.Now()),
	}
```

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +166,114 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedConditions = append(degradedConditions, setDNSLastTransitionTime(degradedCondition, oldDegradedCondition))
+
+		return degradedConditions, nil
+	}
+
+	// If the graceCondition occurs, this should be requeued.
+	// Set the final degraded condition as false, add the grace condition, and set it up to requeue the reconcile.
+	// We check this last, because the other conditions are non-recoverable as degraded conditions.
+	if graceCondition.Status == operatorv1.ConditionTrue {
```
Miciah, line 191: Using `checkConditions` neatly abstracted the grace logic, so the caller didn't need to special-case individual grace conditions, compute the requeue time, or have strange edge cases like `requeueTime == 0`.  Also, using `checkConditions` made the DNS operator more consistent with the ingress operator.  Any reason to rework the logic not to use `checkConditions`?

pkg/operator/controller/dns_status_test.go
```diff
@@ -212,142 +225,204 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
+			expectRequeue:     false,
+			expected:          operatorv1.ConditionTrue,
+		},
+		{
+			name:              "no clusterIP, DNS invalid MaxUnavailable",
+			clusterIP:         "",
+			dnsDaemonset:      makeDaemonSet(6, 6, intstr.FromString("TEST")),
+			oldCondition:      defaultOldCondition,
+			oldGraceCondition: defaultOldGraceCondition,
+			expected:          operatorv1.ConditionTrue,
```
Miciah, line 291: Was this an intentional change (from reporting "Unknown")?  I suppose this is all right—it simplifies the logic, and you could make the case that the operator is degraded when the user provides invalid input that prevents accurate status reporting.

Miciah (commented):
Closing in on perfection.  I'm curious why you replaced `checkConditions` with `requeueAfter` in https://github.com/openshift/cluster-dns-operator/compare/d15e37e8b9e579eae4f557ee2cf8719283ecc17c..0af37f549ce48a602ca83cc47b81df963e749425 though.  

Also, I was wondering whether you could use `setDNSLastTransitionTime` for the `MaxUnavailableDNSPodsExceeded` status condition, and I realized that using `setDNSLastTransitionTime` would cause problems for unit tests.  Instead of the following:
```go
		condition.LastTransitionTime = metav1.Now()
```
`setDNSLastTransitionTime` should have the following:
```go
		condition.LastTransitionTime = metav1.NewTime(clock.Now())
```
so that it would use the fake clock set by unit tests.  

On a related note, `setDNSLastTransitionTime` probably should be ignoring whether the reason or message changed and should rather be updating the last transition time only if the status changed, per https://github.com/openshift/cluster-ingress-operator/pull/465#discussion_r494715168, but don't worry about that now unless it would simplify this PR.

pkg/operator/controller/dns_status_test.go
candita: IIrc, you asked me to set the DNSInvalidDNSMaxUnavailable as ConditionTrue in https://github.com/openshift/cluster-dns-operator/pull/290#discussion_r715758605

pkg/operator/controller/dns_status_test.go
```diff
@@ -212,142 +225,204 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
+			}
+		default:
+			t.Errorf("%q: unexpected error: %v", tc.name, retryErr)
+			continue
 		}
-		actual := computeDNSDegradedCondition(oldCondition, tc.clusterIP, true, tc.dnsDaemonset)
-		if actual.Status != tc.expected {
-			t.Errorf("%q: expected status to be %s, got %s: %#v", tc.name, tc.expected, actual.Status, actual)
+		for _, cond := range actual {
+			grace := cond.Type == DNSMaxUnavailableDNSPodsExceeded && cond.Status == operatorv1.ConditionTrue && tc.expectAfter == 0
```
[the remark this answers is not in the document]
candita: Not really.  I wanted to check what was returned in the conditions, but it's not necessary. Removed.

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +166,114 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
-			degradedReasons = append(degradedReasons, "MaxUnavailableDNSPodsExceeded")
-			messages = append(messages, fmt.Sprintf("Too many DNS pods are unavailable (%d > %d max unavailable).", numberUnavailable, maxUnavailable))
+			graceCondition = &operatorv1.OperatorCondition{Type: DNSMaxUnavailableDNSPodsExceeded, Reason: DNSMaxUnavailableDNSPodsExceeded, Status: operatorv1.ConditionTrue, Message: fmt.Sprintf("Too many DNS pods are unavailable (%d > %d max unavailable).", numberUnavailable, maxUnavailable)}
+		}
+	}
+
+	// If there are degraded conditions, there is no grace period, and this should not be requeued.
+	// Set the final degraded condition as true, and concatenate the reasons and messages
+	if len(degradedConditions) != 0 {
+		degradedCondition.Status = operatorv1.ConditionTrue
```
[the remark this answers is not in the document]
candita: ok

pkg/operator/controller/dns_status_test.go
Miciah: I was saying you should set `InvalidDNSMaxUnavailable=True`, but not necessarily `Degraded=True`.  However, if you want to say `Degraded=True` when the user provides an invalid value for `MaxUnavailable`, that doesn't seem too unreasonable.

pkg/operator/controller/dns_status.go
```diff
@@ -52,35 +72,96 @@ func computeDNSStatusConditions(dns *operatorv1.DNS, clusterIP string, haveDNSDa
 		}
 	}
 
-	conditions := []operatorv1.OperatorCondition{
-		computeDNSDegradedCondition(oldDegradedCondition, clusterIP, haveDNSDaemonset, dnsDaemonset),
-		computeDNSProgressingCondition(oldProgressingCondition, dns, clusterIP, haveDNSDaemonset, dnsDaemonset, haveNodeResolverDaemonset, nodeResolverDaemonset),
-		computeDNSAvailableCondition(oldAvailableCondition, clusterIP, haveDNSDaemonset, dnsDaemonset),
-		computeDNSUpgradeableCondition(oldUpgradeableCondition, dns),
-	}
+	conditions := []operatorv1.OperatorCondition{}
```
Miciah, line 75: This line is superfluous because the next line redeclares `conditions`.

pkg/operator/controller/dns_status.go
```diff
@@ -52,35 +72,96 @@ func computeDNSStatusConditions(dns *operatorv1.DNS, clusterIP string, haveDNSDa
...
+	// DNSMaxUnavailableDNSPodsExceeded indicates that the number of unavailable DNS
+	// pods is greater than the configured MaxUnavailable
+	DNSMaxUnavailableDNSPodsExceeded = "MaxUnavailableDNSPodsExceeded"
+)
+
 // computeDNSDegradedCondition computes the dns Degraded status condition
 // based on the status of clusterIP and the DNS daemonset.  The node-resolver
 // daemonset is not a part of the calculation of degraded condition.
-func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clusterIP string, haveDNSDaemonset bool, dnsDaemonset *appsv1.DaemonSet) operatorv1.OperatorCondition {
+func computeDNSDegradedCondition(oldDegradedCondition *operatorv1.OperatorCondition, oldGraceCondition *operatorv1.OperatorCondition, clusterIP string, haveDNSDaemonset bool, dnsDaemonset *appsv1.DaemonSet) ([]operatorv1.OperatorCondition, error) {
```
Miciah, line 111: Small
```suggestion
func computeDNSDegradedCondition(oldDegradedCondition, oldGraceCondition *operatorv1.OperatorCondition, clusterIP string, haveDNSDaemonset bool, dnsDaemonset *appsv1.DaemonSet) ([]operatorv1.OperatorCondition, error) {
```

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +173,137 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+	// If there are degraded conditions, there is no grace period, and this should not be requeued.
+	// Set the final degraded condition as true, and concatenate the reasons and messages
+	if len(degradedConditions) != 0 {
+		reasons, messages := formatConditions(degradedConditions)
+		degradedCondition.Status = operatorv1.ConditionTrue
+		degradedCondition.Reason = reasons
+		degradedCondition.Message = messages
+		degradedConditions = append(degradedConditions, setDNSLastTransitionTime(degradedCondition, oldDegradedCondition))
+
+		return degradedConditions, nil
```
Miciah, line 215: Does persist all degraded status conditions to the API?  Could the following happen?

1. User sets an invalid value for `MaxUnavailable`.
2. Operator sets the `InvalidDNSMaxUnavailable=True` and `Degraded=True` status conditions.
3. User sets a valid value for `MaxUnavailable`.  
4. Operator updates the `Degraded` status condition to `False` and leaves `InvalidDNSMaxUnavailable=True`.  

`computeDNSDegradedCondition` isn't setting `LastTransitionTime` on status conditions other than `Degraded`, so that's also odd if we are in fact persisting them.

pkg/operator/controller/dns_status_test.go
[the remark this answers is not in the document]
Miciah: Sorry if I wasn't clear; I wasn't expecting you to remove the status-condition check entirely, I was just asking about the grace condition/period.  I suppose `TestDNSStatusConditions` is sufficient to verify that `Degraded` is set to the correct status though.  If `TestComputeDNSDegradedCondition` isn't going to check the status condition, we can delete `expected` from `testCases` (the type definition and the test cases).

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +173,137 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+
+	// If the graceCondition occurs, this should be requeued.
+	// Set the final degraded condition as false, add the grace condition, and set it up to requeue the reconcile.
+	// We check this last, because the other conditions are non-recoverable as degraded conditions.
+	if graceCondition.Status == operatorv1.ConditionTrue {
+		requeueTime := requeueAfter(expectedConditions, graceCondition, oldGraceCondition)
+		if requeueTime == 0 {
+			graceCondition.Status = operatorv1.ConditionUnknown
+		}
+		degradedConditions = append(degradedConditions, setDNSLastTransitionTime(graceCondition, oldGraceCondition))
```
Miciah, line 226: Similar question to https://github.com/openshift/cluster-dns-operator/pull/290#discussion_r742436545:  Do we only persist the `MaxUnavailableDNSPodsExceeded` status condition if its status is `True`?  Once it's been set to `True`, does anything ever delete it or update it to `False`?

pkg/operator/controller/dns_status.go
```diff
@@ -92,34 +173,137 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+func formatConditions(conditions []operatorv1.OperatorCondition) (string, string) {
+	var reasons, messages string
+	if len(conditions) == 0 {
+		return "", ""
+	}
+	for _, cond := range conditions {
+		reasons = reasons + fmt.Sprintf(", %s=%s (%s)", cond.Type, cond.Status, cond.Reason)
+		messages = messages + fmt.Sprintf(", %s=%s (%s)", cond.Type, cond.Status, cond.Message)
+	}
+	return reasons[2:], messages[2:]
```
Miciah, line 302: The status condition's reason should be a CamelCase string.  You could do something like this:
```suggestion
		reasons = reasons + "And" + cond.Reason
		messages = messages + fmt.Sprintf(", %s=%s (%s)", cond.Type, cond.Status, cond.Message)
	}
	return reasons[3:], messages[2:]
```
Or just use something like "DegradedConditions" as the reason.

Miciah:
> > Closing in on perfection. I'm curious why you replaced checkConditions with requeueAfter in https://github.com/openshift/cluster-dns-operator/compare/d15e37e8b9e579eae4f557ee2cf8719283ecc17c..0af37f549ce48a602ca83cc47b81df963e749425 though.
> 
> Since there was only one graceCondtion possible, and I already had the degraded conditions, it didn't make sense to me to do that again. All I'm doing in requeueAfter is getting the requeue time.

It is a little confusing because `requeueAfter` retains a lot of the logic from `checkConditions` that made sense for checking a list of grace conditions, but that logic now it has no effect (e.g., you have a map with 1 entry, you check `condition.Status == expected.status` and `expected.gracePeriod != 0` even though those conditions will always be true, and so on).  The same is true of `computeDNSDegradedCondition` (e.g., you list multiple status conditions in `expectedConditions`, but `requeueAfter` ignores all but one of them), so it almost seems like `requeueAfter` is supposed to be re-usable as `checkConditions` was, but `requeueAfter` also has logic that is very specific to `computeDNSDegradedCondition`.  I might be biased because I'm more familiar with the `checkConditions` logic, but to me, the abstraction that `checkConditions` provided made the logic easier to reason about.  (Having real persisted status conditions in the ingress operator also helps keep the logic relatively simple there.)  If it's just me though, we can stick with `requeueAfter`.  

> > On a related note, setDNSLastTransitionTime probably should be ignoring whether the reason or message changed and should rather be updating the last transition time only if the status changed, per [openshift/cluster-ingress-operator#465](https://github.com/openshift/cluster-ingress-operator/pull/465) (comment), but don't worry about that now unless it would simplify this PR.
> 
> It could introduce some regressions, since it is used in other code. If you feel strongly, I can do this in a separate PR.

We should fix it at some point, but not in this PR.

candita:
e2e-upgrade cluster creation error:

>  msg=failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed to apply Terraform: exit status 1
> level=error
> level=error msg=Error: error creating EC2 NAT Gateway: InvalidSubnet: The subnet ID 'subnet-0a219e1d950f6f2f5' does not exist
> level=error msg=	status code: 400, request id: fae11075-9034-43eb-aabe-c521b0e60e2

pkg/operator/controller/dns_status.go
candita: removed

pkg/operator/controller/dns_status.go
candita: If there's any change to conditions, it does try to persist that in `syncDNSStatus`.  But `computeDNSDegradedCondition` freshly sets the conditions before the attempt to persist.

candita:
> level=error msg=failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed to apply Terraform: exit status 1
> level=error
> level=error msg=Error: error creating EC2 NAT Gateway: InvalidSubnet: The subnet ID 'subnet-0a219e1d950f6f2f5' does not exist
> level=error msg=	status code: 400, request id: fae11075-9034-43eb-aabe-c521b0e60e27
> 

/test e2e-upgrade

candita:
> === RUN   TestDNSOverTLSForwarding
> I0614 22:13:20.248424    8309 request.go:665] Waited for 1.023841583s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-op-k0ik9q46-599c0.origin-ci-int-aws.dev.rhcloud.com:6443/apis/authorization.openshift.io/v1?timeout=32s
>     operator_test.go:751: upstream-tls/upstream-tls-1: failed to find "NOERROR"
>     operator_test.go:779: upstream-tls/upstream-tls-1: failed to find "NOERROR"
>     operator_test.go:785: upstream-tls/upstream-tls-2: failed to find "NOERROR"
>     operator_test.go:792: failed to parse "NOERROR" from upstream resolver pods: failed to find "NOERROR"
> --- FAIL: TestDNSOverTLSForwarding (154.65s)
/test e2e-aws-operator

pkg/operator/controller/controller.go
```diff
@@ -199,11 +198,20 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
-	// Log in case of errors as the controller's logs get eaten.
-	if len(errs) > 0 {
-		logrus.Errorf("failed to reconcile request %s: %v", request, utilerrors.NewAggregate(errs))
+	if err := retry.NewMaybeRetryableAggregate(errs); err != nil {
+		switch e := err.(type) {
+		case retry.Error:
+			// Got retryable error; requeue
+			logrus.Infof("got retryable error; requeueing after %s", e.After().String())
```
Miciah, line 205:
```suggestion
			logrus.Errorf("got retryable error; requeueing after %s: %v", e.After().String(), e)
```

pkg/operator/controller/dns_status.go
Miciah: I see, `computeDNSDegradedCondition` just doesn't return `MaxUnavailable` etc. if they are false, so `syncDNSStatus` removes them, right?  So that hypothetical scenario would go like the following:

1. User sets an invalid value for `MaxUnavailable`.  
2. Operator sets the `InvalidDNSMaxUnavailable=True` and `Degraded=True` status conditions.  
3. User sets a valid value for `MaxUnavailable`.  
4. Operator updates the `Degraded` status condition to `False` and deletes the `InvalidDNSMaxUnavailable` status condition.  

So the degraded status conditions (other than `Degraded` itself) will come and go, rather than their status being updated to true or false.

Miciah:
> I might be biased because I'm more familiar with the `checkConditions` logic, but to me, the abstraction that `checkConditions` provided made the logic easier to reason about. (Having real persisted status conditions in the ingress operator also helps keep the logic relatively simple there.) If it's just me though, we can stick with `requeueAfter`.

If I moved cluster-ingress-operator's `expectedCondition` type and `checkConditions` function to a utility package and improved its documentation (maybe with some examples of how an operator is expected to use them), would you be more amenable to using those in cluster-dns-operator?

PR #296, opened 2021-09-15
Bug 2006803: Set CoreDNS cache entries for forwarded zones

mheler:
Forward entries are missing a cache TTL on them, for larger environments with forwarded zones this can lead to heavier then expected load on DNS servers.

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -29,6 +29,9 @@ var corefileTemplate = template.Must(template.New("Corefile").Parse(`{{range .Se
     {{- end}}
     errors
     bufsize 512
+    cache 900 {
+        denial 9984 30
+    }
 }
 {{end -}}
 .:5353 {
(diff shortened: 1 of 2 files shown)


Miciah:
Please update `TestDesiredDNSConfigmap` and ensure make sure unit tests pass locally: https://github.com/openshift/cluster-dns-operator/blob/a00bd3ea305708f69326b824290566f1ff858fe3/pkg/operator/controller/controller_dns_configmap_test.go#L37-L47

mheler:
@Miciah All set, please let me know if anything else is needed.

Miciah:
Thanks!  Do you mind filing a Bugzilla report for this?  If it's any trouble, I can take care of it, but having a BZ number will be useful for tracking the issue and facilitating backports if these are needed.  

/ok-to-test

roldancer:
Hi, I would like to know if it possible to activate the cache's serve_stale feature for the forwareds zone, serve_stale could help to improve application resilience in case of DNS resolver or DNS authoritative outage.

https://coredns.io/plugins/cache/

serve_stale, when serve_stale is set, cache always will serve an expired entry to a client if there is one available. When this happens, cache will attempt to refresh the cache entry after sending the expired cache entry to the client. The responses have a TTL of 0. DURATION is how far back to consider stale responses as fresh. The default duration is 1h.

Thanks.

PR #298, opened 2021-10-01
podsecurity: enforce privileged for openshift-dns namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-dns` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz


candita:
You will need to add the generated files to satisfy the `verify` test.

assets/dns/namespace.yaml
```diff
@@ -10,3 +10,6 @@ metadata:
     openshift.io/run-level: "0"
     # allow openshift-monitoring to look for ServiceMonitor objects in this namespace
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
```
candita, line 13: Could you add a comment for these new labels?

s-urbaniak: @candita fixed, ptal :-)

candita:
> numerous ContainerExit code/137

/test e2e-upgrade

candita:
> Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-dns` namespace to run privileged pods.

Just checking, lest I miss an opportunity to learn new jargon, is "admirably restricted" supposed to be "at most restricted"?

candita:
Just a grammar nit, and need to wait for successful CI test, but otherwise
/lgtm

Miciah:
@s-urbaniak,
> Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-dns` namespace to run privileged pods.

The `openshift-dns` namespace specifies `openshift.io/run-level: "0"` to skip admission entirely so as to avoid a cyclic dependency between DNS and admission; see https://github.com/openshift/cluster-dns-operator/pull/34/commits/27a4f94a9168c1a0be5447d0fb9120022cc45558 by @deads2k.  Is the intent eventually to remove `openshift.io/run-level: "0"` from the namespace?  If so, how will we prevent a cyclic dependency without specifying run-level?  I don't see this addressed in https://github.com/openshift/enhancements/blob/master/enhancements/authentication/pod-security-admission.md.

PR #300, opened 2021-10-12
Bug 1967963: Add the node-resolver daemonset alert

brandisher:


brandisher:
/hold

Looking at some additional options after further review from the monitoring team.

PR #306, opened 2021-11-12
Update Corefile in configmap based on upstreamResolvers in DNS spec

sherine-k:
Update Corefile in configmap based on upstreamResolvers in DNS spec

[CFE-158](https://issues.redhat.com/browse/CFE-158)
[RFE-1797](https://issues.redhat.com/browse/RFE-1797)


go.mod
```diff
@@ -16,3 +16,5 @@ require (
 	k8s.io/kubectl v0.21.0
 	sigs.k8s.io/controller-runtime v0.9.0
 )
+
+replace github.com/openshift/api => github.com/sherine-k/api v0.0.0-20211112165433-12e91273cfc1
```
arjunrn, line 20: Because the `replace` directive is still in here this PR should be marked as WIP.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -106,14 +108,32 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string) (*corev1.Con
...
 	}
 
+	upstreamResolvers := operatorv1.UpstreamResolvers{
+		Upstreams: []string{"/etc/resolv.conf"},
+		Policy:    operatorv1.SequentialForwardingPolicy,
+	}
+
+	if dns.Spec.UpstreamResolvers.Upstreams != nil && len(dns.Spec.UpstreamResolvers.Upstreams) > 0 {
+		upstreamResolvers.Upstreams = dns.Spec.UpstreamResolvers.Upstreams
+		if len(dns.Spec.UpstreamResolvers.Upstreams) == 1 && dns.Spec.UpstreamResolvers.Upstreams[0] == "" {
```
arjunrn, line 118: Why would this happen? Why would the user specify an array with the first item being an empty string.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -51,9 +51,11 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
         fallthrough in-addr.arpa ip6.arpa
     }
     prometheus 127.0.0.1:9153
-    forward . /etc/resolv.conf {
-        policy sequential
+	{{- with .UpstreamResolvers }}
```
arjunrn, line 54: Can you add unit tests for this?

go.mod
sherine-k: Sure :)

pkg/operator/controller/controller_dns_configmap.go
sherine-k: A scenario might be a misinterpretation from a user who wants to simply modify the  policy, so  instead of omitting the upstreams completely, would set it empty. 
Maybe far fetched? 
If you all think that, let's remove it

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -106,14 +108,32 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string) (*corev1.Con
...
+		Upstreams: []string{"/etc/resolv.conf"},
+		Policy:    operatorv1.SequentialForwardingPolicy,
+	}
+
+	if dns.Spec.UpstreamResolvers.Upstreams != nil && len(dns.Spec.UpstreamResolvers.Upstreams) > 0 {
+		upstreamResolvers.Upstreams = dns.Spec.UpstreamResolvers.Upstreams
+		if len(dns.Spec.UpstreamResolvers.Upstreams) == 1 && dns.Spec.UpstreamResolvers.Upstreams[0] == "" {
+			upstreamResolvers.Upstreams[0] = "/etc/resolv.conf"
+		}
+	}
```
arjunrn, line 121: You don't need to check for nil before calling `len()` on an slice type. If it's `nil` then it will [return 0](https://pkg.go.dev/builtin#len).
```suggestion
	if len(dns.Spec.UpstreamResolvers.Upstreams) > 0 {
		upstreamResolvers.Upstreams = dns.Spec.UpstreamResolvers.Upstreams
		// if user sets the upstream resolver list to an empty string default to the host resolve config file.
		if len(dns.Spec.UpstreamResolvers.Upstreams) == 1 && dns.Spec.UpstreamResolvers.Upstreams[0] == "" {
			upstreamResolvers.Upstreams[0] = "/etc/resolv.conf"
		}
	}
```

sherine-k:
/assign @Miciah 
Hi Miciah, do you think we can merge this PR?

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -336,3 +337,1449 @@ foo.com:5353 {
...
+    forward . 1.2.3.4 9.8.7.6 6.4.3.2:53 2.3.4.5:5353 127.0.0.53 {
+        policy round_robin
+    }
+    cache 900 {
+        denial 9984 30
+    }
+    reload
+}
+`,
+			expectedError: nil,
```
alebedev87, line 502: Just to save some lines in this already long file: this field can be omitted if no error is expected, it'll be `nil` by default.

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -336,3 +337,1449 @@ foo.com:5353 {
...
+								Port:    5353,
+							},
+							{
+								Type:    operatorv1.NetworkResolverType,
+								Address: "127.0.0.53",
+							},
+						},
+						Policy: operatorv1.RoundRobinForwardingPolicy,
+					},
+					Servers: []operatorv1.Server{
```
alebedev87, line 383: By the way, if the goal of this test is the upstream resolvers, maybe it's not that necessary to add 4 servers?
Would only one be enough? That can save a lot of space.

alebedev87 (commented):
Thanks for the PR @sherine-k ! Some remakes left though, mostly cosmetic.

manifests/0000_70_dns-operator_00.crd.yaml
```diff
@@ -0,0 +1,351 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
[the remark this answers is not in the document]
sherine-k: This is because on the api side, the post-generation make tasks (crd patching) only kick in when the files are named *.crd.yaml

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -170,6 +199,23 @@ func corefileChanged(current, expected *corev1.ConfigMap) (bool, *corev1.ConfigM
...
+		return RESOLV_CONF, nil
+	case operatorv1.NetworkResolverType:
+		if upstream.Address == "" {
+			return "", ErrInvalidNetworkUpstream
+		}
+		if upstream.Port > 0 {
+			return fmt.Sprintf("%s:%d", upstream.Address, upstream.Port), nil
+		} else {
+			return upstream.Address, nil
+		}
```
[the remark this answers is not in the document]
sherine-k: If the port is not set, then coreDNS applies a 53 port by default: setting an upstream with just the ip is  allowed.

The api should have us covered though: the default port value is set to 53. 
In a same manner, the api should never allow Address to be empty. so technically speaking we should never enter this code L204

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -170,6 +199,20 @@ func corefileChanged(current, expected *corev1.ConfigMap) (bool, *corev1.ConfigM
 	return true, updated
 }
 
+func coreDNSResolver(upstream operatorv1.Upstream) (string, error) {
+	if upstream.Type == operatorv1.NetworkResolverType {
+		if upstream.Address == "" {
+			return "", errInvalidNetworkUpstream
+		}
+		if upstream.Port > 0 {
+			return fmt.Sprintf("%s:%d", upstream.Address, upstream.Port), nil
```
Miciah, line 208: The address could be IPv6, right?  
```suggestion
			return net.JoinHostPort(upstream.Address, fmt.Sprintf("%d", upstream.Port)), nil
```

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -336,3 +338,739 @@ foo.com:5353 {
...
+    }
+    cache 900 {
+        denial 9984 30
+    }
+    reload
+}
+`,
+		},
+		{
+			name: "Ce with multiple upstreamResolvers, of which /etc/resolv.conf should return coreFile with all the upstreams",
```
Miciah, line 853:
```suggestion
			name: "CR with multiple upstreamResolvers, of which /etc/resolv.conf is one, should return Corefile with all the upstreams",
```

pkg/operator/controller/controller_dns_configmap.go
sherine-k: great catch! thx

Miciah:
Thanks!  
/lgtm
/hold cancel

PR #308, opened 2021-12-13
Bug 2031685: Duplicated forward upstreams should be removed from the dns operator

sherine-k:
Fixes 

- https://bugzilla.redhat.com/show_bug.cgi?id=2031685

Found by QE (cc @ShudiLi ) for https://issues.redhat.com/browse/NE-684

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"text/template"
 
 	"github.com/openshift/cluster-dns-operator/pkg/manifests"
@@ -23,6 +24,7 @@ import (
 )
 
 const resolvConf = "/etc/resolv.conf"
+const defaultDNSPort = 53
 
 var errInvalidNetworkUpstream = fmt.Errorf("The address field is mandatory for upstream of type Network, but was not provided")
 var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.FuncMap{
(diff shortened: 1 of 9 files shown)


pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -237,3 +248,24 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
...
+				if strings.EqualFold(anUpstream.Address,upstream.Address){
+					if anUpstream.Port == upstream.Port {
+						return true
+					}
+					continue
+				}
+				continue
+			}
+		}
+		continue
```
Miciah, line 268: These continue statements are unnecessary.

Miciah: Also, you could reduce indentation by inverting the if conditions; i.e., change this:
```go
for ... {
	if a {
		if b {
			// etc.
		}
	}
}
```
to this:
```go
for ... {
	if !a {
		continue
	}
	if !b {
		continue
	}
	// etc.
}
```

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -237,3 +248,21 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
...
+
+func contains(upstreams []operatorv1.Upstream, upstream operatorv1.Upstream) bool {
+	for _, anUpstream := range upstreams {
+		if upstream.Type == operatorv1.SystemResolveConfType && anUpstream.Type == upstream.Type {
+			return true
+		}
+		if upstream.Type == operatorv1.NetworkResolverType {
+			if anUpstream.Type == upstream.Type {
+				if strings.EqualFold(anUpstream.Address, upstream.Address) {
+					if anUpstream.Port == upstream.Port {
```
Miciah, line 260: Should empty and "53" be treated as equal?  That is, should the following:
```yaml
    - address: 100.1.1.1
      port: 53
      type: Network
```
be considered a duplicate of the following?
```yaml
    - address: 100.1.1.1
      type: Network
```
It might be worth using go-cmp to simplify this logic (e.g., something like `daemonsetConfigChanged` and `cmpTerminationGracePeriodSeconds` in `controller_dns_daemonset.go`).

sherine-k: Nice catch.
Change made

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -237,3 +249,40 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
...
+		//}
+		//if upstream.Type == operatorv1.NetworkResolverType {
+		//	if anUpstream.Type == upstream.Type {
+		//		if strings.EqualFold(anUpstream.Address, upstream.Address) {
+		//			if anUpstream.Port == upstream.Port {
+		//				return true
+		//			}
+		//		}
+		//	}
+		//}
```
Miciah, line 266: No need to keep the old code in a comment.

sherine-k: oops sorry about that :/

Miciah:
Thanks!
/approve
/lgtm

PR #309, opened 2021-12-17
Bug 2031699: Fix CoreDNS config ipv6 addresses should be always upper…

sherine-k:
…case

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap_test.go
+++ b/pkg/operator/controller/controller_dns_configmap_test.go
@@ -366,7 +366,7 @@ func TestDesiredDNSConfigmapUpstreamResolvers(t *testing.T) {
 							},
 							{
 								Type:    operatorv1.NetworkResolverType,
-								Address: "6.4.3.2",
+								Address: "1001:AAAA:bbbb:cCcC::2222",
 								Port:    53,
 							},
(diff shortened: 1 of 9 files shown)


pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -1054,6 +1054,75 @@ foo.com:5353 {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: DefaultDNSController,
+				},
+				Spec: operatorv1.DNSSpec{
+					UpstreamResolvers: operatorv1.UpstreamResolvers{
+						Upstreams: []operatorv1.Upstream{
+							{
+								Type:    operatorv1.NetworkResolverType,
+								Address: "1001:AAAA:bbbb:cCcC::2222",
+								Port:    5353,
```
Miciah, line 1071: I believe if you omitted the port here, then this unit test would provide coverage for the `else` branch here: https://github.com/openshift/cluster-dns-operator/blob/b402c778bb7b3deda5200b259dce34e61aa96399/pkg/operator/controller/controller_dns_configmap.go#L212.

Miciah:
Thanks!
/approve
/lgtm

PR #318, opened 2022-04-05
Check for latest condition update time to prevent status flaps

suleymanakbas91:


pkg/operator/controller/dns_status.go
```diff
@@ -17,6 +18,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// conditionUpdateInterval is the update interval for progressing and degraded conditions.
+const conditionUpdateInterval = 20 * time.Second
```
brandisher, line 22: Is there a "why" for this interval? I see in the BZ that the lameduck mode for CoreDNS lasts 20s; is that what's represented by this value?

frobware: > Exactly, that's the reason.

I think some commentary would help then. Even if it's just a reference to the BZ.

pkg/operator/controller/dns_status.go
```diff
@@ -25,6 +29,11 @@ func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain
 	updated.Status.ClusterDomain = clusterDomain
 	updated.Status.Conditions = computeDNSStatusConditions(dns, clusterIP, haveDNSDaemonset, dnsDaemonset, haveNodeResolverDaemonset, nodeResolverDaemonset)
 	if !dnsStatusesEqual(updated.Status, dns.Status) {
+		// skip frequent updates to progressing and degraded conditions to reduce status flapping.
+		if conditionsUpdatedWithinLast(dns.Status.Conditions, conditionUpdateInterval) {
+			logrus.Infof("last condition update was less than %s ago, skipping this update.", conditionUpdateInterval)
```
frobware, line 34: I'm wondering whether we want to spam the logs here.

pkg/operator/controller/dns_status.go
frobware, line 22: Should this be slightly longer than the lame duck interval? +2s perhaps?

frobware (changes requested):
Overall LGTM but can add a unit test at all?

pkg/operator/controller/dns_status.go
suleymanakbas91: I've added an explanation and a link to the BZ.

pkg/operator/controller/dns_status.go
suleymanakbas91: I changed it to debug.

pkg/operator/controller/dns_status.go
suleymanakbas91: Makes sense. I've made it 22s.

pkg/operator/controller/dns_status.go
```diff
@@ -34,8 +44,27 @@ func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain
...
+				degradedUpdated = true
+			}
+		}
+	}
+	return progressingUpdated && degradedUpdated
+}
+
 // computeDNSStatusConditions computes dns status conditions based on
-// the status of ds and clusterIP.
+// the status of dns and clusterIP.
```
Miciah, line 67: `ds` refers to the DNS daemonset.
```suggestion
// the status of the dns and node-resolver daemonsets and clusterIP.
```

pkg/operator/controller/dns_status.go
```diff
@@ -34,8 +44,27 @@ func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain
 	return nil
 }
 
+// conditionsUpdatedWithinLast computes if the last update for progressing and degraded conditions
```
Miciah, line 47: Is it necessary to check the degraded status condition?  If a Corefile update causes enough pods to go unavailable at once that `computeDNSDegradedCondition` reports degraded, then it seems like that really is a problem.

pkg/operator/controller/dns_status.go
```diff
@@ -25,6 +30,11 @@ func (r *reconciler) syncDNSStatus(dns *operatorv1.DNS, clusterIP, clusterDomain
 	updated.Status.ClusterDomain = clusterDomain
 	updated.Status.Conditions = computeDNSStatusConditions(dns, clusterIP, haveDNSDaemonset, dnsDaemonset, haveNodeResolverDaemonset, nodeResolverDaemonset)
 	if !dnsStatusesEqual(updated.Status, dns.Status) {
+		// skip frequent updates to progressing and degraded conditions to reduce status flapping.
+		if conditionsUpdatedWithinLast(dns.Status.Conditions, conditionUpdateInterval) {
```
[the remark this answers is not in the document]
suleymanakbas91: It cannot leave the condition at `Progressing=True` because it prevents updates only when `Progressing=False`. The only time we will miss updates is when `Progressing` becomes `True` temporarily, which should be ok.

pkg/operator/controller/dns_status.go
suleymanakbas91: If there is a problem that lasts longer than 22s, then there will be reconciliations coming anyway as cluster keeps trying to fix the problem. In this case, we will set it to `Degraded=True` in one of the reconciliations after 22s. But, if a problem lasts less than 22s, then would it still make sense to set `Degraded=True` for a short time?

pkg/operator/controller/dns_status.go
```diff
@@ -17,6 +18,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// statusUpdateRecency is the time duration to decide if we can skip a status update when the last statuses of progressing and degraded conditions were set to false,
```
frobware, line 21: Typo? `statusUpdatedRecently`?

frobware: As discussed in slack you're going to reword.

brandisher (commented):
These changes look good to me. Appropriate test coverage for the change and the comment explaining the "why" behind the change is helpful!

pkg/operator/controller/dns_status.go
```diff
@@ -110,6 +123,11 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 	}
 
 	if len(degradedReasons) != 0 {
+		// if the last status was set to false within the last statusTransitionAllowedDeltaTimeInSeconds, skip the new update
```
frobware, line 126: `statusTransitionAllowedDeltaTimeInSeconds` should be `transitionUnchangedToleration`.

pkg/operator/controller/dns_status.go
```diff
@@ -169,6 +191,11 @@ func computeDNSProgressingCondition(oldCondition *operatorv1.OperatorCondition,
 		}
 	}
 	if len(messages) != 0 {
+		// if the last status was set to false within the last statusTransitionAllowedDeltaTimeInSeconds, skip the new update
```
frobware, line 194: `statusTransitionAllowedDeltaTimeInSeconds` should be `transitionUnchangedToleration`.

pkg/operator/controller/dns_status_test.go
```diff
@@ -689,3 +828,86 @@ func TestDNSStatusesEqual(t *testing.T) {
...
+		},
+		{
+			description: "recent: elapsed time 9s and will tolerate 9s",
+			currTime:    time.Date(2022, time.Month(5), 19, 1, 10, 20, 0, time.UTC),
+			prevTime:    time.Date(2022, time.Month(5), 19, 1, 10, 11, 0, time.UTC),
+			toleration:  9 * time.Second,
+			isRecent:    true,
+		},
+		{
+			description: "recent: elapsed time is 0s 2ns and will tolerate 2ns", // XXX
```
frobware, line 869: Drop the `// XXX`

PR #321, opened 2022-06-02
Bug 2092509: Append an error when a non-existent configmap is configured for DNS-over-TLS

brandisher:
* Fixes a nil pointer error that was crashing the operator when configuring DNS-over-TLS with a configmap that does not exist.
* Simplifies some of the configmap management logic.
* Adds a corresponding e2e test.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -406,10 +406,13 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS, reconcileResult *reconcile.R
 		errs = append(errs, fmt.Errorf("failed to create ca bundle configmaps for dns %s: %w", dns.Name, err))
 	}
 
-	cmMap, err := r.caBundleRevisionMap(dns)
-	if err != nil {
-		return fmt.Errorf("failed to generate ca bundle revision map: %w", err)
-	}
+	// cmMap is a map of ca bundle configmaps' names and their resource versions used in
+	// the calculation of the path of ca bundles in Corefile and volume mount path in daemonset.
+	// When the map is empty or missing a configmap because it is not found on the cluster,
+	// the operator will print a warning, and it will not specify a path in Corefile for this configmap
+	// which will lead to the default behavior of using the system certificates.
+	// Also, the operator will not add a volume to daemonset for this configmap.
+	cmMap := r.caBundleRevisionMap(dns)
 
 	haveDNSDaemonset, dnsDaemonset, err := r.ensureDNSDaemonSet(dns, cmMap)
 	if err != nil {
@@ -461,17 +464,19 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS, reconcileResult *reconcile.R
 // in the source ca bundle configmap (e.g. cert rotation). Apart from resource versions,
 // server name is also used in the path of ca bundles in case the same ca bundle configmap
 // is specified for two different servers.
-func (r *reconciler) caBundleRevisionMap(dns *operatorv1.DNS) (map[string]string, error) {
+func (r *reconciler) caBundleRevisionMap(dns *operatorv1.DNS) map[string]string {
 	caBundleRevisions := map[string]string{}
 	transportConfig := dns.Spec.UpstreamResolvers.TransportConfig
 	if transportConfig.Transport == operatorv1.TLSTransport {
 		if transportConfig.TLS != nil && transportConfig.TLS.CABundle.Name != "" {
 			name := CABundleConfigMapName(transportConfig.TLS.CABundle.Name)
 			cm := &corev1.ConfigMap{}
-			if err := r.client.Get(context.TODO(), name, cm); err != nil {
-				return caBundleRevisions, err
+			err := r.client.Get(context.TODO(), name, cm)
+			if err != nil {
+				logrus.Warningf("failed to get destination ca bundle configmap %s: %v", name.Name, err)
+			} else {
+				caBundleRevisions[transportConfig.TLS.CABundle.Name] = fmt.Sprintf("%s-%s", cm.Name, cm.ResourceVersion)
 			}
-			caBundleRevisions[transportConfig.TLS.CABundle.Name] = fmt.Sprintf("%s-%s", cm.Name, cm.ResourceVersion)
 		}
 	}
 
@@ -481,15 +486,17 @@ func (r *reconciler) caBundleRevisionMap(dns *operatorv1.DNS) (map[string]string
 			if transportConfig.TLS != nil && transportConfig.TLS.CABundle.Name != "" {
 				name := CABundleConfigMapName(transportConfig.TLS.CABundle.Name)
 				cm := &corev1.ConfigMap{}
-				if err := r.client.Get(context.TODO(), name, cm); err != nil {
-					return caBundleRevisions, err
+				err := r.client.Get(context.TODO(), name, cm)
+				if err != nil {
+					logrus.Warningf("failed to get destination ca bundle configmap %s: %v", name.Name, err)
+				} else {
+					caBundleRevisions[transportConfig.TLS.CABundle.Name] = fmt.Sprintf("%s-%s", cm.Name, cm.ResourceVersion)
 				}
-				caBundleRevisions[transportConfig.TLS.CABundle.Name] = fmt.Sprintf("%s-%s", cm.Name, cm.ResourceVersion)
 			}
 		}
 	}
 
-	return caBundleRevisions, nil
+	return caBundleRevisions
 }
 
 // getClusterIPFromNetworkConfig will return 10th IP from the service CIDR range
--- a/pkg/operator/controller/controller_cabundle_configmap.go
+++ b/pkg/operator/controller/controller_cabundle_configmap.go
@@ -64,14 +64,18 @@ func (r *reconciler) ensureCABundleConfigMaps(dns *operatorv1.DNS) error {
 		}
 		haveSource, source, err := r.currentCABundleConfigMap(sourceName)
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to get source ca bundle config map %s: %w", sourceName.Name, err))
+			errs = append(errs, fmt.Errorf("failed to get source ca bundle configmap %s: %w", sourceName.Name, err))
+			continue
+		}
+		if !haveSource {
+			logrus.Warningf("source ca bundle configmap %s does not exist", sourceName.Name)
 			continue
 		}
 
 		destName := CABundleConfigMapName(source.Name)
 		have, current, err := r.currentCABundleConfigMap(destName)
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to get destination ca bundle config map %s: %w", destName.Name, err))
+			errs = append(errs, fmt.Errorf("failed to get destination ca bundle configmap %s: %w", destName.Name, err))
 			continue
 		}
 
--- a/pkg/operator/controller/controller_cabundle_configmap_test.go
+++ b/pkg/operator/controller/controller_cabundle_configmap_test.go
@@ -30,16 +30,16 @@ func TestDesiredCABundleConfigmap(t *testing.T) {
 
 	desired, cm, err := desiredCABundleConfigMap(dns, true, &sourceConfigmap, destName)
 	if err != nil || desired == false {
-		t.Errorf("Unexpected error : %v", err)
+		t.Errorf("unexpected error : %v", err)
 	} else if diff := cmp.Diff(cm.Data, sourceConfigmap.Data); diff != "" {
-		t.Errorf("Unexpected CA Bundle ConfigMap data;\n%s", diff)
+		t.Errorf("unexpected CA Bundle ConfigMap data;\n%s", diff)
 	} else if diff := cmp.Diff(cm.OwnerReferences, []metav1.OwnerReference{dnsOwnerRef(dns)}); diff != "" {
-		t.Errorf("Unexpected CA Bundle ConfigMap OwnerReference;\n%s", diff)
+		t.Errorf("unexpected CA Bundle ConfigMap OwnerReference;\n%s", diff)
 	}
 
 	desired, cm, err = desiredCABundleConfigMap(dns, false, &sourceConfigmap, destName)
 	if desired != false || cm != nil || err != nil {
-		t.Errorf("Unexpected error : %v", err)
+		t.Errorf("expected return values of false, nil, nil when haveSource is false: %v", err)
 	}
 
 	var delTimestamp *metav1.Time
@@ -50,6 +50,6 @@ func TestDesiredCABundleConfigmap(t *testing.T) {
 
 	desired, cm, err = desiredCABundleConfigMap(dns, true, &sourceConfigmap, destName)
 	if desired != false || cm != nil || err != nil {
-		t.Errorf("Unexpected error : %v", err)
+		t.Errorf("expected return values of false, nil, nil when dns.DeletionTimestamp is not nil: %v", err)
 	}
 }
--- a/pkg/operator/controller/controller_cabundle_configmap_test.go
+++ b/pkg/operator/controller/controller_cabundle_configmap_test.go
@@ -1,13 +1,12 @@
 package controller
 
 import (
-	"testing"
-
 	"github.com/google/go-cmp/cmp"
 	operatorv1 "github.com/openshift/api/operator/v1"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+	"time"
 )
 
 func TestDesiredCABundleConfigmap(t *testing.T) {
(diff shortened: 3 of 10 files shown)


brandisher:
Ran the e2e test for DNSOverTLS separately and it passes without issue repeatedly.

/retest

pkg/operator/controller/controller.go
```diff
@@ -461,17 +458,19 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
...
 	transportConfig := dns.Spec.UpstreamResolvers.TransportConfig
 	if transportConfig.Transport == operatorv1.TLSTransport {
 		if transportConfig.TLS != nil && transportConfig.TLS.CABundle.Name != "" {
 			name := CABundleConfigMapName(transportConfig.TLS.CABundle.Name)
 			cm := &corev1.ConfigMap{}
-			if err := r.client.Get(context.TODO(), name, cm); err != nil {
-				return caBundleRevisions, err
+			err := r.client.Get(context.TODO(), name, cm)
+			if err != nil {
+				logrus.Warningf("failed to get destination ca bundle configmap %s: %v", name.Name, err)
```
gcs278, line 470: Can you explain the difference between returning an `err` and this `logrus`? I understand the standard `err`, that just gets thrown in the cluster-dns-operator logs, but where does `logrus` go?

gcs278 (commented):
Didn't find any issues, just a had a question out of curiosity.

pkg/operator/controller/controller.go
brandisher: We don't return an error in this case because we don't want to stop execution if a configmap doesn't exist. `logrus` will send a record to the logs using whatever the operator has configured; as far as I can tell, the operator doesn't have any special logging configuration so it should be going to stdout.

gcs278: Ah got it. So returning an error actually can change code flow, but you just want to throw an warning and continue on if there is no valid configmap.

One thought: If I create an upstream configuration to use a configmap `foo`, but `foo` isn't created yet, but then I create `foo` after I create the configmap, what will happen? Will it ignore the newly create configmap `foo`? Or will it reconcile at some point and use `foo` accordingly?

brandisher: Good question! The reconcile loop will keep looking for the source CM and logging the warning until it's created. Once the source CM is created, it'll update the daemonset and corresponding Corefile. While the source configmap is missing, you'll see these in the operator logs:
```
time="2022-06-07T17:10:35Z" level=warning msg="source ca bundle configmap missing-cm does not exist"
time="2022-06-07T17:10:35Z" level=warning msg="failed to get destination ca bundle configmap ca-missing-cm: configmaps \"ca-missing-cm\" not found"
```
Once the source configmap is created you'll see it resolve right away:
```
time="2022-06-07T17:16:06Z" level=info msg="created configmap openshift-dns/ca-missing-cm"
time="2022-06-07T17:16:06Z" level=info msg="updated dns daemonset openshift-dns/dns-default:   &v1.DaemonSet
```
`ca-missing-cm` is the destination CM that the operator creates by making a copy of the source CM and then the next log is the daemonset/Corefile update.

gcs278: Got it. Thanks for the detailed answer!

pkg/operator/controller/controller.go
```diff
@@ -406,10 +406,7 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
 		errs = append(errs, fmt.Errorf("failed to create ca bundle configmaps for dns %s: %w", dns.Name, err))
 	}
 
-	cmMap, err := r.caBundleRevisionMap(dns)
-	if err != nil {
-		return fmt.Errorf("failed to generate ca bundle revision map: %w", err)
-	}
+	cmMap := r.caBundleRevisionMap(dns)
```
candita, line 409: Can you comment as to how the rest of the code would work with a potentially empty cmMap?

candita: e.g., in https://github.com/openshift/cluster-dns-operator/pull/321/files#diff-6712a6f071bb599a25fa8780722861fd1f5cd051b22abd505c6fa7b9f3c7cb53R154

pkg/operator/controller/controller_cabundle_configmap.go
```diff
@@ -64,14 +64,18 @@ func (r *reconciler) ensureCABundleConfigMaps(dns *operatorv1.DNS) error {
 		}
 		haveSource, source, err := r.currentCABundleConfigMap(sourceName)
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to get source ca bundle config map %s: %w", sourceName.Name, err))
+			errs = append(errs, fmt.Errorf("failed to get source ca bundle configmap %s: %w", sourceName.Name, err))
+			continue
+		}
+		if !haveSource {
```
candita, line 70: If we log it for the source, why don't we log it for the destination below?

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -76,7 +76,7 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
         {{- with $tls := .TransportConfig.TLS }}
         {{- with $serverName := $tls.ServerName }}
         tls_servername {{$serverName}}
-        tls {{- with $cm := $tls.CABundle.Name}} /etc/pki/{{$serverName}}-{{ index $.CABundleRevisionMap $cm }}/{{ $.CABundleFileName }}{{end}}
+        tls {{- with $cm := (index $.CABundleRevisionMap $tls.CABundle.Name) }} /etc/pki/{{$serverName}}-{{ $cm }}/{{ $.CABundleFileName }}{{end}}
```
candita, line 79: If it's possible for CABundleRevisionMap to be nil here, I think you have to also do a `{{- with .CABundleRevisionMap }}` check first.

pkg/operator/controller/controller_cabundle_configmap.go
suleymanakbas91: The source cm is provided by the user, whereas the destination is created by the operator based on the source cm. If the destination cm does not exist, that does not indicate a problem as it will be created by the operator later in line 99.

pkg/operator/controller/controller.go
suleymanakbas91: Sure, I've added a comment.

pkg/operator/controller/controller_dns_configmap.go
suleymanakbas91: Sure, I've applied that.

brandisher:
Infra failures

/retest

pkg/operator/controller/controller.go
```diff
@@ -406,10 +406,12 @@ func (r *reconciler) ensureDNS(dns *operatorv1.DNS) error {
...
 
-	cmMap, err := r.caBundleRevisionMap(dns)
-	if err != nil {
-		return fmt.Errorf("failed to generate ca bundle revision map: %w", err)
-	}
+	//cmMap is a map of ca bundle configmaps' names and their resource versions
+	//used in the calculation of the path of ca bundles in Corefile and volume mount path in daemonset.
+	//It can be empty or missing a configmap because it was not found on cluster. In this case,
+	//when a configmap is not found on cmMap, the operator will print a warning, and
+	//it will be ignored while calculating the mentioned paths.
```
candita, line 413:
```suggestion
	// cmMap is a map of ca bundle configmaps' names and their resource versions
	// used in the calculation of the path of ca bundles in Corefile and volume mount path in daemonset.
	// It can be empty or missing a configmap because it was not found on cluster. In this case,
	// when a configmap is not found on cmMap, the operator will print a warning, and
	// it will be ignored while calculating the mentioned paths.
```

candita: Can you please supply the default path names that are produced when the cmMap is missing?

suleymanakbas91: I've changed the comment a bit. Let me know if it is clearer now. When the cmMap is missing, there will be no path specified for caBundle that will default to using the system certificates. Also, the operator won't add the volume to daemonset. So, it will be safely ignored and the user will be notified through the warnings in the logs. If we decide not to proceed, this will block reconciliation of all the other resources.

pkg/operator/controller/controller_cabundle_configmap.go
candita: https://github.com/openshift/cluster-dns-operator/pull/321/files#diff-6321af060e023fc35bbfc959a1073934fbd15ff8778cc24126d9d4470e87ae77R152-R154 shows that if there is no source, then `want` is false.  If `want` and `have` are both false, we don't create anything, correct?

pkg/operator/controller/testdata/tls_with_non_existing_cabundle
```diff
@@ -0,0 +1,25 @@
...
+    }
+    ready
+    kubernetes cluster.local in-addr.arpa ip6.arpa {
+        pods insecure
+        fallthrough in-addr.arpa ip6.arpa
+    }
+    prometheus 127.0.0.1:9153
+    forward . tls://9.8.7.6 tls://[1001:AAAA:BBBB:CCCC::2222]:53 {
+        tls_servername example.com
+        tls
```
candita, line 18: I wonder if this entire line should be removed.  What happens when just `tls` is in the config?

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -40,7 +40,7 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
         {{- with $tls := .TransportConfig.TLS }}
         {{- with $serverName := $tls.ServerName }}
         tls_servername {{$serverName}}
-        tls {{- with $cm := $tls.CABundle.Name}} /etc/pki/{{$serverName}}-{{ index $.CABundleRevisionMap $cm }}/{{ $.CABundleFileName }}{{end}}
+        tls {{- with $.CABundleRevisionMap }}{{- with $cm := (index $.CABundleRevisionMap $tls.CABundle.Name) }} /etc/pki/{{$serverName}}-{{ $cm }}/{{ $.CABundleFileName }}{{end}}{{end}}
```
candita, line 43: Should the entire line be removed or does `tls` with nothing following it, work ok?

test/e2e/operator_test.go
```diff
@@ -793,6 +793,74 @@ func TestDNSOverTLSForwarding(t *testing.T) {
...
+	}
+
+	dnsOperatorPods, err := getClusterDNSOperatorPods(cl)
+	if err != nil {
+		t.Errorf("unable to get operator pods: #{err}")
+	}
+
+	for _, dnsOperatorPod := range dnsOperatorPods.Items {
+		if err := lookForStringInPodLog(operatorcontroller.DefaultOperatorNamespace, dnsOperatorPod.Name, operatorcontroller.ContainerNameOfDNSOperator, "level=info", 30*time.Second); err != nil {
+			t.Fatal("source ca bundle config map")
```
candita, line 858:
```suggestion
			t.Fatal("could not find pod logs %w", err)
```

test/e2e/operator_test.go
```diff
@@ -793,6 +793,74 @@ func TestDNSOverTLSForwarding(t *testing.T) {
...
+	})
+
+	// Verify that default DNS pods are all available before inspecting them.
+	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
+		t.Errorf("expected default DNS pods to be available: %v", err)
+	}
+
+	dnsOperatorPods, err := getClusterDNSOperatorPods(cl)
+	if err != nil {
+		t.Errorf("unable to get operator pods: #{err}")
```
candita, line 853: This `#{err}` looks like a goland artifact.

test/e2e/operator_test.go
candita: Would be nice if you could find the error log message that should have been posted for a missing CM.

test/e2e/operator_test.go
```diff
@@ -1192,3 +1260,21 @@ func upstreamTLSPod(name, ns, image string, configMap *corev1.ConfigMap) *corev1
 		},
 	}
 }
+
+func getClusterDNSOperatorPods(cl client.Client) (*corev1.PodList, error) {
```
candita, line 1264:
```suggestion
func getClusterDNSOperatorPods(cl client.Client) (*corev1.PodList, error) {
	t.Helper()
```

candita: Add t.Helper() here as well as other helper functions that should not be treated as standalone tests.

test/e2e/operator_test.go
brandisher: Fixed in https://github.com/openshift/cluster-dns-operator/pull/321/commits/9dd51de4c823be08b33d55b182652d9db3dfc6e8

test/e2e/operator_test.go
brandisher: Fixed in https://github.com/openshift/cluster-dns-operator/pull/321/commits/9dd51de4c823be08b33d55b182652d9db3dfc6e8

pkg/operator/controller/testdata/tls_with_non_existing_cabundle
brandisher: When only using `tls`, CoreDNS will use system certs to validate the server certificate of the upstream resolver. The docs are not terribly verbose on this but in manual and e2e testing, having `tls` alone on a line works: https://coredns.io/plugins/forward/

pkg/operator/controller/controller_dns_configmap.go
brandisher: Similarly here, `tls` alone on one line is a valid configuration. In our case a cluster admin can configure one or none configmaps; none indicating that they want to use the system certs for the upstream server validation.

test/e2e/operator_test.go
brandisher: I've addressed this by moving the helper functions out of the `operator_test.go` file.

brandisher:
tide/merge-method-squash

brandisher:
Infra issue

/test e2e-aws-single-node

pkg/operator/controller/controller_cabundle_configmap_test.go
```diff
@@ -1,13 +1,12 @@
...
 
 import (
-	"testing"
-
 	"github.com/google/go-cmp/cmp"
 	operatorv1 "github.com/openshift/api/operator/v1"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
```
candita, line 8: The go packages should be first in the list.

pkg/operator/controller/controller_cabundle_configmap_test.go
```diff
@@ -28,12 +27,28 @@ func TestDesiredCABundleConfigmap(t *testing.T) {
...
 		t.Errorf("Unexpected error : %v", err)
 	} else if diff := cmp.Diff(cm.Data, sourceConfigmap.Data); diff != "" {
 		t.Errorf("Unexpected CA Bundle ConfigMap data;\n%s", diff)
 	} else if diff := cmp.Diff(cm.OwnerReferences, []metav1.OwnerReference{dnsOwnerRef(dns)}); diff != "" {
 		t.Errorf("Unexpected CA Bundle ConfigMap OwnerReference;\n%s", diff)
 	}
+
+	desired, cm, err = desiredCABundleConfigMap(dns, false, &sourceConfigmap, destName)
+	if desired != false || cm != nil || err != nil {
+		t.Errorf("Unexpected error : %v", err)
```
candita, line 41: Either a comment or a more informative error message prefix would be helpful, here and in the next one.

nit: first word of error messages should be lower-case.

pkg/operator/controller/controller_cabundle_configmap_test.go
brandisher: Ah good catch; Goland thwarted me again :-) I've fixed this locally and will push up the fix shortly.

pkg/operator/controller/controller_cabundle_configmap_test.go
brandisher: Fixed this locally as well; I've downcased the first letter and made the messages more clear.

brandisher: Fixed in https://github.com/openshift/cluster-dns-operator/pull/321/commits/d52dedae2997a56997be10a907ea34cf679c7132

pkg/operator/controller/controller_cabundle_configmap_test.go
brandisher: Fixed in https://github.com/openshift/cluster-dns-operator/pull/321/commits/b34b2f79a523ef62abd01b3dc7c3f326f839d957

pkg/operator/controller/controller_cabundle_configmap_test.go
```diff
@@ -29,16 +30,16 @@ func TestDesiredCABundleConfigmap(t *testing.T) {
...
+		t.Errorf("unexpected CA Bundle ConfigMap data;\n%s", diff)
 	} else if diff := cmp.Diff(cm.OwnerReferences, []metav1.OwnerReference{dnsOwnerRef(dns)}); diff != "" {
-		t.Errorf("Unexpected CA Bundle ConfigMap OwnerReference;\n%s", diff)
+		t.Errorf("unexpected CA Bundle ConfigMap OwnerReference;\n%s", diff)
 	}
 
 	desired, cm, err = desiredCABundleConfigMap(dns, false, &sourceConfigmap, destName)
 	if desired != false || cm != nil || err != nil {
-		t.Errorf("Unexpected error : %v", err)
+		t.Errorf("expected return values of false, nil, nil when haveSource is false: %v", err)
```
candita, line 42: nit
```suggestion
		t.Errorf("unexpected results when source is missing: desired  is %t; %v", desired, err)
```

pkg/operator/controller/controller_cabundle_configmap_test.go
```diff
@@ -49,6 +50,6 @@ func TestDesiredCABundleConfigmap(t *testing.T) {
 
 	desired, cm, err = desiredCABundleConfigMap(dns, true, &sourceConfigmap, destName)
 	if desired != false || cm != nil || err != nil {
-		t.Errorf("Unexpected error : %v", err)
+		t.Errorf("expected return values of false, nil, nil when dns.DeletionTimestamp is not nil: %v", err)
```
candita, line 53: nit
```suggestion
		t.Errorf("expected results if config map was deleted: desired is %t: %v", desired, err)
```

PR #323, opened 2022-06-07
Bug 2093236: Revert "Merge pull request #318 from suleymanakbas91/status-flaps"

frobware:
This reverts commit d50df32df68f53c1d47db8f5e51a8b27c402f278, reversing
changes made to 5d78163210b5697442d3ad48a9ee1313a5010dce.

This breaks upgrades. I will follow up with a new fix.


frobware:
Rather than going ahead with the revert let's use the fix in PR #325.

PR #325, opened 2022-06-07
Bug 2093236: propagate retry request for progressing updates

frobware:
This is a follow up to:

    https://github.com/openshift/cluster-dns-operator/pull/318

When we prevent frequent status updates between True/False for
Progressing, ensure that reconcile will run again in the future for
the specific case where we return the existing condition.


frobware:
Jun  8 13:03:14.069: FAIL: disruption/openshift-api connection/new was unreachable during disruption:  for at least 10s of 1h11m5s (maxAllowed=9s):

/test e2e-upgrade

PR #332, opened 2022-07-06
add Ethany-RH to OWNERS

Ethany-RH:
add Ethan Yu to OWNERS


candita:
Access to cluster-dns-operator not needed by @Ethany-RH 
/close

PR #335, opened 2022-08-09
NE-929: CoreDNS Cache Tuning - Functional Implementation

brandisher:
EP: https://github.com/openshift/enhancements/pull/1139
API: https://github.com/openshift/api/pull/1214
Epic: https://issues.redhat.com/browse/NE-757

* Pulls in the updated `openshift/api` to make the APIs available.
* Adds `coreDNSCache()` which takes `spec.cache` and determines if the default caching values should be used. There's also a small adjustment to the Corefile template to ensure that the cache values are propagated in all the server blocks.


pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -344,6 +353,30 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
 	return "class error"
 }
 
+// coreDNSCache reads the TTL values set in spec.cache and returns integer representations of those values in seconds.
+// This returns uint32 values because CoreDNS won't allow setting the maxTTLs to 0 and there's no use case for setting a
+// negative maxTTL value.
+func coreDNSCache(dns *operatorv1.DNS) (positiveTTL, negativeTTL uint32) {
+	var p, n uint32
```
suleymanakbas91, line 360: As you already have named return parameters `positiveTTL` and `negativeTTL`, you can just use them instead of creating new variables.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -24,6 +24,8 @@ import (
 const resolvConf = "/etc/resolv.conf"
 const defaultDNSPort = 53
 const lameDuckDuration = 20 * time.Second
+const cacheDefaultMaxPositiveTTL = 900
```
suleymanakbas91, line 27: As we don't use `* time.Second` but `uint32`, it is not obvious if these are in seconds. You can use `InSeconds` in the constant name to make it more obvious.

suleymanakbas91 (changes requested):

pkg/operator/controller/controller_dns_configmap.go
brandisher: Fixed!

pkg/operator/controller/controller_dns_configmap.go
brandisher: Good call! I added a note about why they're integers instead of `time.Duration` and refactored the names to include `Seconds`.

brandisher:
/test e2e-upgrade

Infra failures

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -171,6 +176,9 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string, caBundleRevi
 		upstreamResolvers.Policy = dns.Spec.UpstreamResolvers.Policy
 	}
 
+	// Calculate the caching values (in seconds) for use in the Corefile
+	pTTL, nTTL := coreDNSCache(dns)
```
frobware, line 180: nit: can we add something to the function name? `Settings`?

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -344,6 +356,28 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
 	return "class error"
 }
 
+// coreDNSCache reads the TTL values set in spec.cache and returns integer representations of those values in seconds.
+// This returns uint32 values because CoreDNS won't allow setting the maxTTLs to 0 and there's no use case for setting a
+// negative maxTTL value.
+func coreDNSCache(dns *operatorv1.DNS) (positiveTTL, negativeTTL uint32) {
```
frobware, line 362: Would it be better to make this func more "functional"? So a function which takes in values and returns value and doesn't need to operate on an instance of `*operatorv1.DNS`. It's arguably easier to test free-form functions.

brandisher: If its a blocker I can change it. The intent with the function argument and name was to follow the same pattern as `coreDNSLogLevel()`, `coreDNSPolicy()`, etc. This one is nuanced comparitively since it returns two values but the scope is the same; taking the `DNSSpec`, finding the right configuration, then returning a Corefile acceptable value (or set of values).

pkg/operator/controller/controller_dns_configmap.go
brandisher: If it's not a blocker I'd like to leave it as-is for now and defer the renaming to a later date if more cache settings get surfaced in the API. Mostly because at that point we'll need to discuss whether we want to handle every cache setting in one function or break them out into discrete functions.

pkg/operator/controller/controller_dns_configmap.go
Miciah: I don't see the problem with taking a `*operatorv1.DNS`.  It's still a pure function, and having the function take `*operatorv1.DNS` as input means that the unit test for the function uses test input that is in the same format as what we get from the API.  

Using named return values is unusual in this codebase, but I won't hold that against you.  * grin *.

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -883,6 +914,84 @@ func TestDesiredDNSConfigmapUpstreamResolvers(t *testing.T) {
 	}
 }
 
+func Test_coreDNSCache(t *testing.T) {
+	testcases := []struct {
```
Miciah, line 918: Please add test cases for fractional values.  For example:
```go
		{
			name: "999ms rounds to 0, resulting in the default settings",
			dns: &operatorv1.DNS{
				Spec: operatorv1.DNSSpec{
					Cache: operatorv1.DNSCache{
						PositiveTTL: metav1.Duration{999 * time.Millisecond},
						NegativeTTL: metav1.Duration{999 * time.Millisecond},
					},
				},
			},
			expectedPTTL: cacheDefaultMaxPositiveTTLSeconds,
			expectedNTTL: cacheDefaultMaxNegativeTTLSeconds,
		},
		{
			name: "1999ms rounds to 1s",
			dns: &operatorv1.DNS{
				Spec: operatorv1.DNSSpec{
					Cache: operatorv1.DNSCache{
						PositiveTTL: metav1.Duration{1999 * time.Millisecond},
						NegativeTTL: metav1.Duration{1999 * time.Millisecond},
					},
				},
			},
			expectedPTTL: 1,
			expectedNTTL: 1,
		},
```

manifests/0000_70_dns-operator_00.crd.yaml
```diff
@@ -42,6 +42,47 @@ spec:
...
+                      it must be a value of 1s (1 second) or greater up to a theoretical
+                      maximum of several years. If not configured, the value will
+                      be 0 (zero) and OpenShift will use a default value of 30 seconds
+                      unless noted otherwise in the respective Corefile for your version
+                      of OpenShift. The default value of 30 seconds is subject to
+                      change. This field expects an unsigned duration string of decimal
+                      numbers, each with optional fraction and a unit suffix, e.g.
+                      \"100s\", \"1m30s\". Valid time units are \"s\", \"m\", and
+                      \"h\"."
+                    pattern: ^(0|([0-9]+(\.[0-9]+)?(s|m|h))+)$
```
Miciah, line 69: Are we going to see a repeat of [OCPBUGS-236](https://issues.redhat.com/browse/OCPBUGS-236) when the operator serializes "1.1s" to "1100ms" and tries to do an update?  @gcs278

pkg/operator/controller/controller_dns_configmap_test.go
brandisher: The minimum unit allowed by the API is seconds so are these tests a preventative measure? I tested some values [here](https://github.com/openshift/api/pull/1214#issuecomment-1215700550) per the API team's request.

manifests/0000_70_dns-operator_00.crd.yaml
gcs278: Good catch. I think that it will happen. Though, `1.1s` wouldn't actually be converted, but I know for sure `0.5s` will b  converted to `500ms`. Here's the [playground](https://go.dev/play/p/dgWS_-goMq8) you created that I tested with.

pkg/operator/controller/controller_dns_configmap_test.go
Miciah: Right, I'm looking for something along those lines, the goal being to codify the expected behavior around fractional values.  For example, a reasonable person might expect "0.9s" to be clamped or rounded up to 1s, but instead "0.9s" gets rounded down to 0 and thus defaulted to 30s, so the unit tests should codify that behavior.  

I see you also tested various ways of expressing zero values.  For the purposes of this unit test, we don't need to worry about the various yaml representations of zero values since the corresponding Go values are all equal—that is, the following test will pass (this is just to illustrate my point; I'm not asking you to add the following to your PR):
```go
	a := operatorv1.DNS{
		Spec: operatorv1.DNSSpec{
			Cache: operatorv1.DNSCache{
				PositiveTTL: metav1.Duration{},
				NegativeTTL: metav1.Duration{},
			},
		},
	}
	b := operatorv1.DNS{
		Spec: operatorv1.DNSSpec{
			Cache: operatorv1.DNSCache{
				PositiveTTL: metav1.Duration{0 * time.Second},
				NegativeTTL: metav1.Duration{0 * time.Second},
			},
		},
	}
	c := operatorv1.DNS{
		Spec: operatorv1.DNSSpec{
			Cache: operatorv1.DNSCache{},
		},
	}
	if !reflect.DeepEqual(a, b) {
		t.Errorf("a!=b\na: %v\nb: %v", a, b)
	}
	if !reflect.DeepEqual(a, c) {
		t.Errorf("a!=c\na: %v\nc: %v", a, c)
	}
```

manifests/0000_70_dns-operator_00.crd.yaml
brandisher: Fixed in the API here: https://github.com/openshift/api/pull/1315

I'll update the CRD and push it here.

Miciah:
BTW, please link to the relevant epic in the commit message.

pkg/operator/controller/controller_dns_configmap_test.go
brandisher: Added a few more test cases in the latest changes.

manifests/0000_70_dns-operator_00.crd.yaml
```diff
@@ -42,6 +42,47 @@ spec:
             description: spec is the specification of the desired behavior of the
```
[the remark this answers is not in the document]
brandisher: Fixed!

brandisher:
AWS networking issues.

/test e2e-aws-serial

brandisher:
`TestDNSLogging` flake -- still investigating this in #341 
/test e2e-aws-operator

manifests/0000_70_dns-operator_00.crd.yaml
```diff
@@ -42,14 +42,59 @@ spec:
...
+                      string of decimal numbers, each with optional fraction and a
+                      unit suffix, e.g. \"100s\", \"1m30s\", \"12h30m10s\". Values
+                      that are fractions of a second are rounded down to the nearest
+                      second. If the configured value is less than 1s, the default
+                      value will be used. If not configured, the value will be 0s
+                      and OpenShift will use a default value of 900 seconds unless
+                      noted otherwise in the respective Corefile for your version
+                      of OpenShift. The default value of 900 seconds is subject to
+                      change."
+                    pattern: ^(0|([0-9]+(\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+)$
```
candita, line 87: As I mentioned in chat, the pattern shouldn't allow dots if the end result has to be an integer.  The issue with 0.5s or 0.999s being converted to 0s, then replaced by 30s, seems like a bug to me.  I realize this is in the API, and you have an approval already, so I feel bad having to point it out.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -347,6 +365,28 @@ func coreDNSLogLevel(dns *operatorv1.DNS) string {
...
+	configuredPosTTL := uint32(dns.Spec.Cache.PositiveTTL.Seconds())
```
candita, line 372: uint32(math.Round(dns.Spec.Caceh.PositiveTTL.Seconds())) might give a better result.

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -51,8 +62,8 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
         {{$.LogLevel}}
     }
     bufsize 512
-    cache 900 {
-        denial 9984 30
+    cache {{ $.PositiveTTL }} {
+        denial 9984 {{ $.NegativeTTL }}
```
candita, line 66: Should there be a stanza for `success` as well as for `denial`?

candita, line 66:
```suggestion
    cache  {
        success 9984 {{ $.PositiveTTL }}
        denial 9984 {{ $.NegativeTTL }}
```

pkg/operator/controller/testdata/default_corefile_no_cache_configured
```diff
@@ -0,0 +1,26 @@
...
+        pods insecure
+        fallthrough in-addr.arpa ip6.arpa
+    }
+    prometheus 127.0.0.1:9153
+    forward . /etc/resolv.conf {
+        policy sequential
+    }
+    cache 900 {
+        denial 9984 30
+    }
```
candita, line 21: For a test of "no cache configured", could you remove this stanza entirely?

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -174,6 +175,36 @@ func TestDesiredDNSConfigmap(t *testing.T) {
...
+					Cache: operatorv1.DNSCache{
+						PositiveTTL: metav1.Duration{},
+						NegativeTTL: metav1.Duration{},
+					},
+				},
+			},
+			expectedCoreFile: mustLoadTestFile(t, "default_corefile_no_cache_configured"),
+		},
+		{
+			name: "Default Corefile with cache configured",
```
candita, line 194: Can you add one more test with something odd in the TTLs, like 0.5s.

manifests/0000_70_dns-operator_00.crd.yaml
brandisher: Does that imply that we wouldn't consider something like `0.5h` as a valid value?

brandisher: Maybe another way to put it...do we want to explicitly disallow fractional values?
Correct: `30m`
Incorrect: `0.5h`

pkg/operator/controller/controller_dns_configmap.go
brandisher: It's not necessary since we don't do any additional configuration of caching. We ship with this today:
```
cache 900 {
  denial 9984 30
}
```

pkg/operator/controller/controller_dns_configmap_test.go
brandisher: Added in my local changes.

pkg/operator/controller/testdata/default_corefile_no_cache_configured
brandisher: If I understand correctly, you're suggesting that "no cache configured" means "no cache stanza". In our case, there's always a cache stanza because if the API isn't configured then the default values are used as opposed to not having cache at all. The values we ship with are also different than the CoreDNS defaults so as far as I know, we wouldn't hit a case where there'd be a Corefile with no cache stanza.

manifests/0000_70_dns-operator_00.crd.yaml
candita: The biggest issue is that values between 0.5s and 0.999s end up as 30s.  Maybe if you use https://github.com/openshift/cluster-dns-operator/pull/335#discussion_r984777022 that issue would go away.  

However, a test of an update on a dns configuration that started with 0.5s is still warranted to make sure it doesn't cause a bug

pkg/operator/controller/controller_dns_configmap.go
candita: Imo it would be clearer to spell out each of the success and denial values separately, so we don't get bug reports or questions asking us why it doesn't show up, but in the end I guess it's a bit of a nit.
```suggestion
    cache {
      success 9984 900
      denial 9984 30
      }
```

pkg/operator/controller/controller_dns_configmap.go
brandisher: Fixed in the latest push

manifests/0000_70_dns-operator_00.crd.yaml
brandisher: I manually tested the following:
`900ms` rounds to `1s` resulting in `1` in the Corefile.
`0.4s` rounds to `0s` resulting in the default of `30` in the Corefile
`1.0004s` rounds to `1s` resulting in `1` in the Corefile.
`900000000ms` rounds to `900000s` resulting in `900000` in the Corefile.

I also set the following values and deleted a CoreDNS pod:
```
spec:
  cache:
    negativeTTL: 0.5s
    positiveTTL: 1.2h33.8m10.1s
```
The pod was able to restart without issue and the following config landed in the Corefile after pod restart:
```
    cache 6358 {
        denial 9984 1
    }
```
Are there any other values you want tested?

brandisher:
`TestDNSLogging` flake
/test e2e-aws-operator
/test e2e-aws-single-node

manifests/0000_70_dns-operator_00.crd.yaml
candita: Oh, you mean after the changes you did this?

brandisher: Correct. I implemented a version of what you suggested in this [comment](https://github.com/openshift/cluster-dns-operator/pull/335#discussion_r984777022). Instead of using `math` I used https://pkg.go.dev/time#Duration.Round

PR #336, opened 2022-08-17
OCPBUGS-360: Pod security policy change breaks DNS forwarding e2e tests

brandisher:
This fixes the e2e breakage introduced with the recent adjustment of the default security policy.

* Adds a `SecurityContext` configuration to `upstreamContainer()` as that's used by both `TestDNSForwarding` and `TestDNSOverTLSForwarding`. The configuration is standard with one exception; `RunAsNonRoot: false` is needed because we're using `cluster-dns-operator`'s CoreDNS image which runs as root in the container.
* Switches the dns-over-tls test namespace to `openshift-testdnsovertls` as a workaround to the pod security autolabeling.
* Adds a volume mode to `upstreamTLSPod()` to match `upstreamPod()` as the lack of the volume mode seemed to stop the test pods from starting.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -506,7 +506,7 @@ func TestDNSOverTLSForwarding(t *testing.T) {
 	// Create a separate namespace for the upstream resolver. Deleting this namespace will clean up the resources created.
 	tlsUpstreamNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: tlsUpstreamName,
+			Name: "openshift-testdnsovertls",
 		},
 	}
 	if err := cl.Create(context.TODO(), tlsUpstreamNamespace); err != nil {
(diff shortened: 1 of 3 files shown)


brandisher:
Infra failures

/test e2e-upgrade

gcs278:
/lgtm
I noticed this was breaking CI, but I don't see any issues with your updates.

test/e2e/operator_test.go
```diff
@@ -506,7 +506,7 @@ func TestDNSOverTLSForwarding(t *testing.T) {
 	// Create a separate namespace for the upstream resolver. Deleting this namespace will clean up the resources created.
 	tlsUpstreamNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: tlsUpstreamName,
+			Name: "openshift-testdnsovertls",
```
candita, line 509: What's the purpose of changing the namespace name here?  Does it still clean up the resources that were created?

candita:
@brandisher can you add a few details to the description and/or comment as to the change for volumeMount and the additional SecurityContext?  Is there a reference that explains why you chose each of these values, or was it trial and error?

test/e2e/operator_test.go
candita: Namespaces prefaced with "openshift" are less restricted.

PR #337, opened 2022-08-25
NE-1066: Enable the chaos plugin for CoreDNS in order to expose CoreDNS metadata

gcs278:
Turns on the `chaos` CoreDNS plugin (https://coredns.io/plugins/chaos/) that exposes some metadata that can help debugging in the future.

The regression https://issues.redhat.com/browse/OCPBUGS-488 found in 4.10 has exposed a CI test gap in local DNS endpoint preference and this plugin would allow a new CI tests to determine what CoreDNS pod is replying to a query so we can determine it is indeed the local DNS endpoint.

### DIFF
--- a/pkg/operator/controller/controller_dns_configmap.go
+++ b/pkg/operator/controller/controller_dns_configmap.go
@@ -87,6 +87,9 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
     }
     reload
 }
+hostname.bind:5353 {
+    chaos
+}
 `))
 
 // ensureDNSConfigMap ensures that a configmap exists for a given DNS.
--- a/pkg/operator/controller/testdata/1ipv6_and_policy
+++ b/pkg/operator/controller/testdata/1ipv6_and_policy
@@ -36,3 +36,6 @@ foo.com:5353 {
     }
     reload
 }
+hostname.bind:5353 {
+    chaos
+}
--- a/pkg/operator/controller/testdata/1ns_and_policy
+++ b/pkg/operator/controller/testdata/1ns_and_policy
@@ -36,3 +36,6 @@ foo.com:5353 {
     }
     reload
 }
+hostname.bind:5353 {
+    chaos
+}
--- a/pkg/operator/controller/testdata/1ns_no_policy
+++ b/pkg/operator/controller/testdata/1ns_no_policy
@@ -36,3 +36,6 @@ foo.com:5353 {
     }
     reload
 }
+hostname.bind:5353 {
+    chaos
+}
--- a/pkg/operator/controller/testdata/1sysresconf
+++ b/pkg/operator/controller/testdata/1sysresconf
@@ -36,3 +36,6 @@ foo.com:5353 {
     }
     reload
 }
+hostname.bind:5353 {
+    chaos
+}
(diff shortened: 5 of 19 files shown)


gcs278:
/hold
waiting on CI

gcs278:
ci/prow/e2e-aws-operator looks good, but failing on known pod security issue.

brandisher:
A couple of questions that popped up when looking at this:
1. What's the use case for having `chaos` for every default zone query?
2. Is there a reason for having it on every default zone query but not on other queries? (e.g. forwarded zone server blocks)

The other thing that comes to mind is, are we considering the version information non-privileged? That seems like platform specific information that pods wouldn't normally have access to, but will have access to which `chaos` enabled.

gcs278:
> What's the use case for having chaos for every default zone query?

Good question. No good reason tbh, I'm pretty new to DNS and zones, but I now that I read more on them, I updated the PR to add `hostname.bind:5353` as a zone, which restricts the chaos plugin to just `hostname.bind` and puts it in a new configurable zone. Let me know if you like the update.

> Is there a reason for having it on every default zone query but not on other queries? (e.g. forwarded zone server blocks)
I think there is no reason to enable chaos on other blocks such as forward because it the chaos metadata query is only a local query and shouldn't be involved with forwarding.

> The other thing that comes to mind is, are we considering the version information non-privileged? That seems like platform specific information that pods wouldn't normally have access to, but will have access to with chaos enabled.

Regarding my answer to your first question, the PR update to restrict to just `hostname.bind` will limit the exposure of any extra information that doesn't have a need to be exposed.

brandisher:
Awesome, thanks for the explanation @gcs278!

/lgtm

gcs278:
/retest
ImgPull Errors, Doesn't look related

candita:
@gcs278 needs BZ JIRA ticket.

gcs278:
/retitle NE-1066: Enable the chaos plugin for coredns in order to expose CoreDNS metadata

gcs278:
/retitle "NE-1066: Enable the chaos plugin for coredns in order to expose CoreDNS metadata"

gcs278:
/retitle NE-1066: Enable the chaos plugin for coredns in order to expose CoreDNS metadata

gcs278:
/retitle NE-1066: Enable the chaos plugin for coredns in order to expose CoreDNS metadata2

gcs278:
/retitle NE-1066: Enable the chaos plugin for CoreDNS in order to expose CoreDNS metadata

gcs278:
Stuck on DNS e2e test pod security issues. @brandisher and I are trying to figure this out.

Miciah:
#339 should have fixed the E2E failurs that were blocking this PR.
/test e2e-aws-operator

brandisher:
This might be a timing issue with the DNS logging test. All the forwarding tests happen before this one which trigger rollouts of the daemonset pods. With the DNS-over-TLS tests we found that in order for them to be successful, we needed to wait until the daemonset was stable before we started testing anything. Perhaps what's happening here is the rollout from the previous test(s) is still settling and then `TestDNSLogging` is updating the spec again, expanding the time it takes for the daemonset to be ready.

/retest

brandisher:
I opened #341 to test out a fix for the `TestDNSLogging` issue.

PR #339, opened 2022-09-02
OCPBUGS-873: Address e2e failures due to pod security

brandisher:
* Simplifies `TestDNSOverTLSForwarding`
* Adds the needed SecurityContext for `TestDNSForwarding` to succeed
* Moves the client pods for `TestDNSOverTLSForwarding` and `TestDNSForwarding` to privileged namespaces.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -579,27 +579,23 @@ func TestDNSOverTLSForwarding(t *testing.T) {
 	}
 
 	// Create the upstream resolver Pods and the client pod
-	upstreamResolver1 := upstreamTLSPod(tlsUpstreamName+"-1", tlsUpstreamNamespace.Name, coreImage, upstreamTLSConfigMap)
-	if err := cl.Create(context.TODO(), upstreamResolver1); err != nil {
-		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver1.Namespace, upstreamResolver1.Name, err)
-	}
-	upstreamResolver2 := upstreamTLSPod(tlsUpstreamName+"-2", tlsUpstreamNamespace.Name, coreImage, upstreamTLSConfigMap)
-	if err := cl.Create(context.TODO(), upstreamResolver2); err != nil {
-		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver2.Namespace, upstreamResolver2.Name, err)
+	upstreamResolver := upstreamTLSPod(tlsUpstreamName, tlsUpstreamNamespace.Name, coreImage, upstreamTLSConfigMap)
+	if err := cl.Create(context.TODO(), upstreamResolver); err != nil {
+		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver.Namespace, upstreamResolver.Name, err)
 	}
 	testClient := buildPod("test-client-tls", tlsUpstreamName, cliImage, []string{"sleep", "3600"})
 	if err := cl.Create(context.TODO(), testClient); err != nil {
 		t.Fatalf("failed to create pod %s/%s: %v", testClient.Namespace, testClient.Name, err)
 	}
 
-	// Wait for the first upstream resolver Pod to be ready.
-	name := types.NamespacedName{Namespace: upstreamResolver1.Namespace, Name: upstreamResolver1.Name}
+	// Wait for the upstream resolver Pod to be ready.
+	name := types.NamespacedName{Namespace: upstreamResolver.Namespace, Name: upstreamResolver.Name}
 	err = wait.PollImmediate(1*time.Second, 2*time.Minute, func() (bool, error) {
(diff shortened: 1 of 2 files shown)


test/e2e/utils.go
```diff
@@ -311,6 +311,16 @@ func buildContainer(name, image string, cmd []string) corev1.Container {
 		Name:    name,
 		Image:   image,
 		Command: cmd,
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: pointer.Bool(false),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{"ALL"},
+			},
+			RunAsNonRoot: pointer.Bool(false),
```
gcs278, line 319: This is a bit of a nit, but `buildContainer` appears to be intended to be for generic container object creation for all e2e tests, but we have `RunAsNonRoot: pointer.Bool(false),` which appears to be a bit specific to the container itself (specifically the DNS E2E tests that use this function).

Maybe all E2E tests need to run as root? Does it make sense to parameterize this `RunAsNonRoot` value?

test/e2e/operator_test.go
```diff
@@ -506,7 +506,13 @@ func TestDNSOverTLSForwarding(t *testing.T) {
 	// Create a separate namespace for the upstream resolver. Deleting this namespace will clean up the resources created.
 	tlsUpstreamNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "openshift-testdnsovertls",
+			Name: tlsUpstreamName,
+			Labels: map[string]string{
+				"pod-security.kubernetes.io/warn":                "privileged",
+				"pod-security.kubernetes.io/audit":               "privileged",
+				"pod-security.kubernetes.io/enforce":             "privileged",
```
gcs278, line 513: So basically all of these labels are creating a privileged namespace?

test/e2e/operator_test.go
```diff
@@ -572,50 +578,51 @@ func TestDNSOverTLSForwarding(t *testing.T) {
...
-		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver1.Namespace, upstreamResolver1.Name, err)
+	// Create the upstream resolver Pods and the client pod
+	upstreamResolver := upstreamTLSPod(tlsUpstreamName, tlsUpstreamNamespace.Name, coreImage, upstreamTLSConfigMap)
+	if err := cl.Create(context.TODO(), upstreamResolver); err != nil {
+		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver.Namespace, upstreamResolver.Name, err)
 	}
-	upstreamResolver2 := upstreamTLSPod(tlsUpstreamName+"-2", tlsUpstreamNamespace.Name, coreImage, upstreamTLSConfigMap)
-	if err := cl.Create(context.TODO(), upstreamResolver2); err != nil {
-		t.Fatalf("failed to create pod %s/%s: %v", upstreamResolver2.Namespace, upstreamResolver2.Name, err)
+	testClient := buildPod("test-client-tls", tlsUpstreamName, cliImage, []string{"sleep", "3600"})
```
gcs278, line 586: So I assume the TLDR is:
```
// This looks in both upstream resolvers because the forwarding policy could be random. This also serves the purpose
// of testing that one server certificate will work for multiple upstreams using the same ServerName.
```

We feel not necessary to test multiple upstream with one certificate? Is it a bit orthogonal? (Sorry I think you Miciah talked about this today, but I missed it...)

gcs278 (commented):
Looks good but mainly just questions.

test/e2e/operator_test.go
brandisher: Exactly. We're testing that we can use the CA to validate the server cert of the upstream and trusting that CoreDNS will move on to the next upstream if the first one doesn't work.

test/e2e/operator_test.go
brandisher: Correct.

test/e2e/utils.go
brandisher: That's a good catch and actually looks like the source of a failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/339/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-operator/1565792699041189888

I'll add the parameter in so it can be adjusted per use case.

brandisher: Oh actually, it looks like there are 3 occurrences of `buildPod()` which uses `buildContainer()` under the covers. DNS Forwarding, DNS Logging, and DNS-over-TLS tests use it but where they're all using the cli image they have to have root in the container. I think the short term fix is probably to run the test clients in a privileged namespace and mid to long term we should think about using a non-privileged container image. With that said, I'll move the client for DNS logging to `openshift-dns`

gcs278:
/retest
failure in e2e-aws-single-node appears like general cluster failure

PR #341, opened 2022-09-16
OCPBUGS-1439: Fix TestDNSLogging race condition

brandisher:
* Moves `TestDNSLogging` further up the chain so it isn't affected by the daemonset churn.
* Some minor test tweaks to improve the run time.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -779,6 +779,11 @@ func TestDNSLogging(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Ensure that DNS is stable before starting the test, otherwise we'd need to tweak individual test durations.
+	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
+		t.Errorf("expected default DNS pods to be available: %v", err)
+	}
+
 	// Get the CoreDNS image used by the test upstream resolver.
 	co := &configv1.ClusterOperator{}
 	if err := cl.Get(context.TODO(), opName, co); err != nil {
@@ -893,14 +898,8 @@ func TestDNSLogging(t *testing.T) {
 			t.Fatalf("failed to dig %v", err)
 		}
 
-		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 2*time.Minute); err != nil {
+		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 5*time.Minute); err != nil {
 			t.Fatalf("failed to set Debug logLevel for operator %s: %v", opName, err)
-		}
-
-		// Get the CoreDNS image used by the test upstream resolver.
-		co := &configv1.ClusterOperator{}
-		if err := cl.Get(context.TODO(), opName, co); err != nil {
-			t.Fatalf("failed to get clusteroperator %s: %v", opName, err)
 		}
 
 		if found == 0 {
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -180,6 +180,177 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
 	}
 }
 
+// TestDNSLogging verifies that the logLevel settings for CoreDNS and the cluster-dns-operator have propagated
+// appropriately. This was moved to this position in the code as a result of https://issues.redhat.com/browse/OCPBUGS-1439
+func TestDNSLogging(t *testing.T) {
+	cl, err := getClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Ensure that DNS is stable before starting the test, otherwise we'd need to tweak individual test durations.
+	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
+		t.Errorf("expected default DNS pods to be available: %v", err)
+	}
+
+	// Get the CoreDNS image used by the test upstream resolver.
+	co := &configv1.ClusterOperator{}
+	if err := cl.Get(context.TODO(), opName, co); err != nil {
+		t.Fatalf("failed to get clusteroperator %s: %v", opName, err)
+	}
+
+	// Update cluster DNS forwarding with the upstream resolver's Service IP address.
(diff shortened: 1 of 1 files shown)


test/e2e/operator_test.go
```diff
@@ -779,6 +779,11 @@ func TestDNSLogging(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Ensure that DNS is stable before starting the test.
+	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
+		t.Errorf("expected default DNS pods to be available: %v", err)
```
frobware, line 784: Can the test realistically make progress if this errors?

brandisher: I'd be inclined to say "no" mostly because if we change this to say, `t.Logf()`, it'll still have the opportunity to flake since we're not starting from a settled/known state. Judging from the test failure after making this change though it seems like there might be some other timing issue so I'll take another look.

brandisher: I've found that the wait for the operator to be steady seems to introduce too much time to the test and cause it to fail. I'm iterating a bit more on the individual values seen in the diff to see if I can at least make this flake less.

test/e2e/operator_test.go
```diff
@@ -180,6 +180,175 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
...
 }
 
+func TestDNSLogging(t *testing.T) {
+	cl, err := getClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Ensure that DNS is stable before starting the test, otherwise we'd need to tweak individual test durations.
+	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
```
brandisher, line 190: Added this to ensure that DNS is stable before starting the test. This cuts down on the number of individual waits that need to be tweaked.

test/e2e/operator_test.go
```diff
@@ -180,6 +180,175 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
...
+	found := 0
+	for _, corednspod := range coreDNSPods.Items {
+
+		// Dig the example dns forwarding host.
+		digCmd := []string{"dig", "test.svc.cluster.local"}
+		if err := lookForStringInPodExec(testClientForDNSLogging.Namespace, testClientForDNSLogging.Name, testClientForDNSLogging.Name, digCmd, "NXDOMAIN", 2*time.Minute); err != nil {
+			t.Fatalf("failed to dig %v", err)
+		}
+
+		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 5*time.Minute); err != nil {
```
brandisher, line 308: The timeout here was changed from `2m` to `5m` as this one consistently times out.

test/e2e/operator_test.go
```diff
@@ -773,176 +942,6 @@ func TestDNSOverTLSToleratesMissingSourceCM(t *testing.T) {
...
-		if err := lookForStringInPodExec(testClientForDNSLogging.Namespace, testClientForDNSLogging.Name, testClientForDNSLogging.Name, digCmd, "NXDOMAIN", 2*time.Minute); err != nil {
-			t.Fatalf("failed to dig %v", err)
-		}
-
-		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 2*time.Minute); err != nil {
-			t.Fatalf("failed to set Debug logLevel for operator %s: %v", opName, err)
-		}
-
-		// Get the CoreDNS image used by the test upstream resolver.
-		co := &configv1.ClusterOperator{}
```
brandisher, line 901: This is removed in the "moved version" of the test. It didn't seem like this was necessary and added a considerable amount of time to the test.

test/e2e/operator_test.go
```diff
@@ -180,6 +180,175 @@ func TestDefaultDNSSteadyConditions(t *testing.T) {
 	}
 }
 
+func TestDNSLogging(t *testing.T) {
```
Miciah, line 183: Is the move necessary in addition to the added `waitForDNSConditions` check?  Ideally, each test would ensure the cluster were in a good state before starting (thus the `waitForDNSConditions`) and leave the cluster in a good state for the next test (maybe we need another `waitForDNSConditions` in or at the end of this test?).  If re-ordering tests is still necessary for some reason, please add a comment here.
```suggestion
// TestDNSLogging verifies ...
// This test must run before other tests because ...
func TestDNSLogging(t *testing.T) {
```

brandisher: Added a comment and included reasoning in the commit message as well.

gcs278: Being devils advocate here, the fact we moved a test around and resolved a problem feels a bit fragile. If this is a requirement for this test to succeed, then I have hesitation that this test might break again.

If this is just 'helpful" to make things go smoother/quicker, then I can probably buy that.

Is it that the the other test causes so much churn, that even 5 minutes times out? These things aren't parallel, so how does it cause that much churn?

gcs278 (commented):
Looks good, just a question

test/e2e/operator_test.go
brandisher: >Is it that the the other test causes so much churn, that even 5 minutes times out?
That's correct. We have a handful of tests now that cause the daemonset to get re-rolled so the variability is much greater.

>If this is just 'helpful" to make things go smoother/quicker, then I can probably buy that.
Nail on the head. This PR is meant to unblock the existing PRs and only that. I have https://issues.redhat.com/browse/OCPBUGS-2002 open to officially deflake `TestDNSLogging` and I suspect that what we really need is a solution like `cluster-ingress-operator` where we separate serial and parallel tests.

gcs278:
Given more work will be done in https://issues.redhat.com/browse/OCPBUGS-2002, it looks good as stop-gap measure to unblock PRs
/lgtm

Miciah:
Thanks!  Let's merge this and iterate.  
/approve

Miciah:
[e2e-aws-serial failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/341/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-serial/1577336632565043200) because of an unhealthy etcd member.
/test e2e-aws-serial

PR #340, opened 2022-09-12
OCPBUGS-753: keep dns-default annotations intact

candita:
When adding the enable DaemonSet eviction annotation `cluster-autoscaler.kubernetes.io/enable-ds-eviction`, do not overwrite other annotations.  Fix missing managed annotation - `target.workload.openshift.io/management`. Fix unit tests. Add e2e test.

modified:   pkg/operator/controller/controller_dns_daemonset.go
modified:   pkg/operator/controller/controller_dns_daemonset_test.go
modified: test/e2e/operator_test.go

PR #343, opened 2022-09-22
OCPBUGS-1549: Reconcile the DNS namespace

candita:
The DNS namespace manifest has been updated, and changes need to be applied during the controller reconcile.  Add all the required labels to a set and compare the existing namespace labels to this set.  If they don't match, change them and update the namespace.  Add a unit test.

NamespaceRunLevelLabel           = "openshift.io/run-level"    
NamespaceClusterMonitoringLabel  = "openshift.io/cluster-monitoring" 
NamespacePodSecurityEnforceLabel = "pod-security.kubernetes.io/enforce"
NamespacePodSecurityAuditLabel   = "pod-security.kubernetes.io/audit" 
NamespacePodSecurityWarnLabel    = "pod-security.kubernetes.io/warn"

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"net"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -43,6 +44,22 @@ const (
 	DNSControllerFinalizer = "dns.operator.openshift.io/dns-controller"
 
 	controllerName = "dns_controller"
+
+	namespaceRunLevelLabel           = "openshift.io/run-level"             // "0"
+	namespaceClusterMonitoringLabel  = "openshift.io/cluster-monitoring"    // "true"
+	namespacePodSecurityEnforceLabel = "pod-security.kubernetes.io/enforce" // privileged
+	namespacePodSecurityAuditLabel   = "pod-security.kubernetes.io/audit"   // privileged
+	namespacePodSecurityWarnLabel    = "pod-security.kubernetes.io/warn"    // privileged
+)
+
+// managedDNSNamespaceLabels is a set of label keys for labels that the operator manages
+// for the operand namespace
+var managedDNSNamespaceLabels = sets.NewString(
+	namespaceClusterMonitoringLabel,
+	namespaceRunLevelLabel,
+	namespacePodSecurityAuditLabel,
+	namespacePodSecurityEnforceLabel,
+	namespacePodSecurityWarnLabel,
 )
 
 // New creates the operator controller from configuration. This is the
@@ -283,15 +300,25 @@ func (r *reconciler) ensureDNSDeleted(dns *operatorv1.DNS) error {
 // ensureDNSNamespace ensures all the necessary scaffolding exists for
 // dns generally, including a namespace and all RBAC setup.
 func (r *reconciler) ensureDNSNamespace() error {
-	ns := manifests.DNSNamespace()
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
+	existingNamespace := corev1.Namespace{}
+	desiredNamespace := manifests.DNSNamespace()
+	if err := r.client.Get(context.TODO(), DefaultDNSOperandNamespaceName(), &existingNamespace); err != nil {
 		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get dns namespace %q: %v", ns.Name, err)
+			return fmt.Errorf("failed to get dns namespace %q: %w", desiredNamespace.Name, err)
 		}
-		if err := r.client.Create(context.TODO(), ns); err != nil {
-			return fmt.Errorf("failed to create dns namespace %s: %v", ns.Name, err)
+		if err := r.client.Create(context.TODO(), desiredNamespace); err != nil {
+			return fmt.Errorf("failed to create dns namespace %s: %w", desiredNamespace.Name, err)
 		}
-		logrus.Infof("created dns namespace: %s", ns.Name)
+		logrus.Infof("created dns namespace: %s", desiredNamespace.Name)
+	} else {
+		// Make sure existing DNS namespace has all the labels from the Namespace manifest
+		changed, updatedNamespace := namespaceLabelsChanged(&existingNamespace, desiredNamespace)
+		if changed {
+			if err := r.client.Update(context.TODO(), updatedNamespace); err != nil {
+				return fmt.Errorf("failed to update dns namespace %s: %w", updatedNamespace.Name, err)
+			}
+			logrus.Infof("updated dns namespace: %s", updatedNamespace.Name)
+		}
 	}
 
 	if _, _, err := r.ensureDNSClusterRole(); err != nil {
@@ -336,6 +363,26 @@ func (r *reconciler) ensureDNSNamespace() error {
 	}
 
 	return nil
+}
+
+// namespaceLabelsChanged generates a new namespace with the desired labels, and reports true if the existing namespace
+// labels need to be updated to match the desired state.
+func namespaceLabelsChanged(existingNamespace, desiredNamespace *corev1.Namespace) (bool, *corev1.Namespace) {
+	changed := false
+	updatedNamespace := existingNamespace.DeepCopy()
+	if updatedNamespace.Labels == nil {
+		updatedNamespace.Labels = map[string]string{}
+	}
+
+	for k := range managedDNSNamespaceLabels {
+		existingVal, have := existingNamespace.Labels[k]
+		desiredVal, want := desiredNamespace.Labels[k]
+		if want && (!have || existingVal != desiredVal) {
+			updatedNamespace.Labels[k] = desiredNamespace.Labels[k]
+			changed = true
+		}
+	}
+	return changed, updatedNamespace
 }
 
 // ensureMetricsIntegration ensures that dns prometheus metrics are integrated with openshift-monitoring for the given DNS.
--- a/pkg/operator/controller/controller_test.go
+++ b/pkg/operator/controller/controller_test.go
@@ -1,0 +1,56 @@
+package controller
+
+import (
+	"github.com/openshift/cluster-dns-operator/pkg/manifests"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestDNSNamespaceLabelsChanged(t *testing.T) {
+	originalNamespace := manifests.DNSNamespace()
+
+	testCases := []struct {
+		description    string
+		mutate         func(namespace *corev1.Namespace)
+		changeExpected bool
+	}{
+		{
+			description:    "if nothing changes",
+			mutate:         func(_ *corev1.Namespace) {},
+			changeExpected: false,
+		},
+		{
+			description: "if arbitrary label added",
+			mutate: func(namespace *corev1.Namespace) {
+				if namespace.Labels == nil {
+					namespace.Labels = map[string]string{}
+				}
+				namespace.Labels["foo"] = "bar"
+			},
+			changeExpected: false,
+		},
+		{
+			description: "if required label changes",
+			mutate: func(namespace *corev1.Namespace) {
+				if namespace.Labels == nil {
+					namespace.Labels = map[string]string{}
+				}
+				namespace.Labels[namespacePodSecurityEnforceLabel] = ""
+			},
+			changeExpected: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		mutated := originalNamespace.DeepCopy()
+		tc.mutate(mutated)
+		if changed, updated := namespaceLabelsChanged(originalNamespace, mutated); changed != tc.changeExpected {
+			t.Errorf("%s, expect namespaceLabelsChanged to be %t, got %t", tc.description, tc.changeExpected, changed)
+		} else if changed {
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/controller.go
```diff
@@ -43,8 +44,17 @@ const (
...
+
+	NamespaceRunLevelLabel           = "openshift.io/run-level"             // "0"
+	NamespaceClusterMonitoringLabel  = "openshift.io/cluster-monitoring"    // "true"
+	NamespacePodSecurityEnforceLabel = "pod-security.kubernetes.io/enforce" // privileged
+	NamespacePodSecurityAuditLabel   = "pod-security.kubernetes.io/audit"   // privileged
+	NamespacePodSecurityWarnLabel    = "pod-security.kubernetes.io/warn"    // privileged
 )
 
+var DNSNamespaceLabels = sets.NewString(NamespaceClusterMonitoringLabel, NamespaceRunLevelLabel,
+	NamespacePodSecurityAuditLabel, NamespacePodSecurityEnforceLabel, NamespacePodSecurityWarnLabel)
```
Miciah, line 56: No need to make this variable public.  I suggest reformatting it as follows and adding godoc:
```suggestion
// managedDNSNamespaceLabels is a set of label keys for labels that the
// operator manages for the operand namespace.
var managedDNSNamespaceLabels = sets.NewString(
	NamespaceClusterMonitoringLabel,
	NamespaceRunLevelLabel,
	NamespacePodSecurityAuditLabel,
	NamespacePodSecurityEnforceLabel,
	NamespacePodSecurityWarnLabel,
)
```

pkg/operator/controller/controller.go
```diff
@@ -283,15 +293,36 @@ func (r *reconciler) ensureDNSDeleted(dns *operatorv1.DNS) error {
 // ensureDNSNamespace ensures all the necessary scaffolding exists for
 // dns generally, including a namespace and all RBAC setup.
 func (r *reconciler) ensureDNSNamespace() error {
-	ns := manifests.DNSNamespace()
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
+	existingNamespace := corev1.Namespace{}
+	desiredNamespace := manifests.DNSNamespace()
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: desiredNamespace.Name}, &existingNamespace); err != nil {
```
Miciah, line 298: Might be worth adding a function to `names.go` that would return the `types.NamespacedName` for the namespace.

pkg/operator/controller/controller.go
```diff
@@ -283,15 +293,36 @@ func (r *reconciler) ensureDNSDeleted(dns *operatorv1.DNS) error {
...
 		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get dns namespace %q: %v", ns.Name, err)
+			return fmt.Errorf("failed to get dns namespace %q: %v", desiredNamespace.Name, err)
```
Miciah, line 300: We should use `%w` in new code or code that we update:  
```suggestion
			return fmt.Errorf("failed to get dns namespace %q: %w", desiredNamespace.Name, err)
```

Miciah (commented):
It would be good to add some unit tests for the update logic.  That would be easier to do if you refactored `ensureDNSNamespace` to put the update logic in a separate function, which as a bonus would make `ensureDNSNamespace` more resemble our other `ensureFoo` methods.

pkg/operator/controller/controller.go
candita: Do you want me to update it everywhere, or just in what I added/changed?

candita:
verify test failure:

Installing yq into '_output/tools/bin/yq-2.4.0'
mkdir -p '_output/tools/bin/'
curl -s -f -L https://github.com/mikefarah/yq/releases/download/2.4.0/yq_linux_amd64 -o '_output/tools/bin/yq-2.4.0'
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/openshift/operator/../yq.mk:18: ensure-yq] Error 6

pkg/operator/controller/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	for _, tc := range testCases {
+		mutated := originalNamespace.DeepCopy()
+		tc.mutate(mutated)
+		if changed := ensureDNSNamespaceLabels(originalNamespace, *mutated); changed != tc.changeExpected {
+			t.Errorf("%s, expect namespaceLabelsChanged to be %t, got %t", tc.description, tc.changeExpected, changed)
+		} else if changed {
+			if changedAgain := ensureDNSNamespaceLabels(originalNamespace, *mutated); changedAgain {
+				t.Errorf("%s, namespaceLabelsChanged does not behave as a fixed point function", tc.description)
```
Miciah, line 52: The test failure message has "namespaceLabelsChanged" instead of "ensureDNSNamespaceLabels".  The name "namespaceLabelsChanged" actually makes sense to me—it's a similar name to a lot of similar functions in the operator.

pkg/operator/controller/controller.go
```diff
@@ -43,6 +44,22 @@ const (
 	DNSControllerFinalizer = "dns.operator.openshift.io/dns-controller"
 
 	controllerName = "dns_controller"
+
+	NamespaceRunLevelLabel           = "openshift.io/run-level"             // "0"
+	NamespaceClusterMonitoringLabel  = "openshift.io/cluster-monitoring"    // "true"
+	NamespacePodSecurityEnforceLabel = "pod-security.kubernetes.io/enforce" // privileged
+	NamespacePodSecurityAuditLabel   = "pod-security.kubernetes.io/audit"   // privileged
+	NamespacePodSecurityWarnLabel    = "pod-security.kubernetes.io/warn"    // privileged
```
Miciah, line 52: Sorry I didn't think of this before, but why are these all public?

pkg/operator/controller/controller.go
Miciah: Just in the code you're adding or changing.

pkg/operator/controller/controller_test.go
candita: Ok, I'll change it.

pkg/operator/controller/controller.go
candita: It's not necessary.  I was thinking about an e2e test, but didn't do one for this.

candita: Fixed

pkg/operator/controller/controller_test.go
```diff
@@ -0,0 +1,56 @@
+package controller
+
+import (
+	"github.com/openshift/cluster-dns-operator/pkg/manifests"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestEnsureDNSNamespaceLabels(t *testing.T) {
```
Miciah, line 10:
```suggestion
func Test_namespaceLabelsChanged(t *testing.T) {
```

pkg/operator/controller/controller.go
```diff
@@ -283,15 +300,25 @@ func (r *reconciler) ensureDNSDeleted(dns *operatorv1.DNS) error {
...
 		}
-		if err := r.client.Create(context.TODO(), ns); err != nil {
-			return fmt.Errorf("failed to create dns namespace %s: %v", ns.Name, err)
+		if err := r.client.Create(context.TODO(), desiredNamespace); err != nil {
+			return fmt.Errorf("failed to create dns namespace %s: %w", desiredNamespace.Name, err)
+		}
+		logrus.Infof("created dns namespace: %s", desiredNamespace.Name)
+	} else {
+		// Make sure existing DNS namespace has all the labels from the Namespace manifest
+		changed := namespaceLabelsChanged(&existingNamespace, *desiredNamespace)
```
gcs278, line 315: So looks like this `namespaceLabelsChanged` function is mutating `existingNamespace`, which I feel is a bit unexpected, as it reads as a boolean-style function. If we look at the pattern of other object `____Changed` functions, they don't mutate the existing object: [canaryNamespaceChanged](https://github.com/openshift/cluster-ingress-operator/blob/542124811a01039852d8b0846da91d33665037f8/pkg/operator/controller/canary/namespace.go#L82) or [dnsRecordChanged](https://github.com/openshift/cluster-ingress-operator/blob/ce263ce12526a398ae36188a1bb59c40e847787d/pkg/operator/controller/ingress/dns.go#L187).

The pattern I see in our code is `ensureObject` -> `updateObject` -> `objectChanged`. A very similar use case may be [ensureCanaryNamespace](https://github.com/openshift/cluster-ingress-operator/blob/542124811a01039852d8b0846da91d33665037f8/pkg/operator/controller/canary/namespace.go#L18) or [ensureRouterNamespace](https://github.com/openshift/cluster-ingress-operator/blob/ce263ce12526a398ae36188a1bb59c40e847787d/pkg/operator/controller/ingress/namespace.go#L20) I'd consider adapting the patterns there to be consistent.

pkg/operator/controller/controller_test.go
candita: Changed to TestDNSNamespaceLabelsChanged

pkg/operator/controller/controller.go
candita: I'm going to decline, hoping this is not a sticking point.  cluster-dns-operator is inconsistent with cluster-ingress-operator in **many** ways, especially in the `Reconcile`, and I don't want to start refactoring as a part of this bug fix.  The `namespaceLabelsChanged` code  was modeled after `daemonsetConfigChanged` in controller_dns_daemonset.go.

gcs278: Actually I thought I was in cluster-ingress-operator, so I agree with your consistency point. cluster-dns-operator is more inconsistent. Regarding the whole ensure -> Update -> changed functional pattern, that's not a problem.

But to your point, `daemonsetConfigChanged` doesn't mutate the `current` value. Consistency aside, I think `namespaceLabelChanged` is violating expectation here. Would you consider matching the pattern of `daemonsetConfigChanged` like, it should be a simple update:
```suggestion
		changed, updated := namespaceLabelsChanged(*existingNamespace, *desiredNamespace)
```

candita:
e2e-aws-operator failure:
 RUN   TestDNSLogging
I0926 21:17:09.781903    8988 request.go:601] Waited for 1.048679755s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-op-v9cjx84q-599c0.origin-ci-int-aws.dev.rhcloud.com:6443/apis/route.openshift.io/v1?timeout=32s
    operator_test.go:917: failed to get NXDOMAIN entry for test.svc.cluster.local. which does not exist
--- FAIL: TestDNSLogging (795.46s)
/test e2e-aws-operator

candita:
Discussed aws-serial etcd/ccpmso failure in https://coreos.slack.com/archives/C027U68LP/p1664384599199049
Should be succeeding now.

/test e2e-aws-serial

candita:
Hmm, another TestDNSLoggingFailure?
=== RUN   TestDNSLogging
I0927 15:40:01.314798    7992 request.go:601] Waited for 1.199051377s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-op-8xvgs1jg-599c0.origin-ci-int-aws.dev.rhcloud.com:6443/apis/operators.coreos.com/v1?timeout=32s
    operator_test.go:897: failed to set Debug logLevel for operator /dns: failed to find "class denial error"
--- FAIL: TestDNSLogging (271.61s)

/test e2e-aws-operator

candita:
@Miciah @gcs278 PTAL, I cleaned up based on changes, need the approval and lgtm labels.

pkg/operator/controller/controller.go
```diff
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
```
Miciah, line 6: This import is in a strange place.

Miciah: This can be addressed in a follow-up.

gcs278:
thanks
/lgtm

candita:
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2022-09-29T22:14:06Z"}

/test e2e-aws-serial

candita:
[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [IdP] test RequestHeaders IdP [apigroup:config.openshift.io][apigroup:user.openshift.io][apigroup:apps.openshift.io] [Suite:openshift/conformance/serial]

/test e2e-aws-serial

candita:
: [sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel] expand_less | 2s
-- | --
{  fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Oct  4 03:45:12.794: Unexpected alerts fired or pending after the test run:  alert OVNKubernetesControllerDisconnectedSouthboundDatabase fired for 30 seconds with labels: {container="kube-rbac-proxy-ovn-metrics", endpoint="ovn-metrics", ...

/test e2e-aws-serial

melvinjoseph86:
@candita don't we want the fix in 4.10 branch?

candita:
> @candita don't we want the fix in 4.10 branch?

@melvinjoseph86 I don't think so.  4.12 and 4.11 should be ok for this one.  There is also #340

PR #347, opened 2022-10-07
OCPBUGS-2112: [release-4.11] Backport Address e2e failures due to pod security

candita:
Backport of #336 and #339 into release-4.11

* Add the security context to buildPod() so it'll be admitted
* Move client pods to privileged namespaces

The client pods were landing in the default namespace so in lieu of relabeling that namespace, we're moving them to namespaces that allow privileged pods. The client pod for TestDNSForwarding will launch in openshift-dns and the client pod for TestDNSOverTLSForwarding will launch in a privileged namespace created as part of the e2e test.

* Simplify the test to avoid race conditions
* Move the dns logging test client to openshift-dns


brandisher:
Infra failures.

/retest

brandisher (commented):
This looks like a clean backport and I don't see any issues.

/lgtm

brandisher:
The latest failure of `e2e-aws-operator` looks like its due to:
```
INFO[2022-10-10T16:40:48Z] GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run ".*" ./test/e2e
# github.com/openshift/cluster-dns-operator/test/e2e [github.com/openshift/cluster-dns-operator/test/e2e.test]
test/e2e/utils.go:302:30: undefined: pointer
test/e2e/utils.go:306:18: undefined: pointer
```
In 4.12 we use `k8s.io/utils/pointer` to inline creation of pointer values but that wasn't included in `utils.go` in 4.11. In this case, instead of pulling in that dependency in a backport and updating `go.mod`, maybe it'd be easier to generalize with something like:
```
trueVar := true
falseVar := false
```
And use those in the `buildContainer()` instead.

candita:
> The latest failure of `e2e-aws-operator` looks like its due to:
> 
> ```
> INFO[2022-10-10T16:40:48Z] GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run ".*" ./test/e2e
> # github.com/openshift/cluster-dns-operator/test/e2e [github.com/openshift/cluster-dns-operator/test/e2e.test]
> test/e2e/utils.go:302:30: undefined: pointer
> test/e2e/utils.go:306:18: undefined: pointer
> ```
> 
> In 4.12 we use `k8s.io/utils/pointer` to inline creation of pointer values but that wasn't included in [`utils.go` in 4.11](https://github.com/openshift/cluster-dns-operator/blob/89980935d93894dd26bc155234c2ff851f093e38/test/e2e/utils.go). In this case, instead of pulling in that dependency in a backport and updating `go.mod`, maybe it'd be easier to generalize with something like:
> 
> ```
> falseVar := false
> ```
> 
> And use that in the `buildContainer()` instead. I think this would work:
> 
> ```
> diff --git a/test/e2e/utils.go b/test/e2e/utils.go
> index d4497caf..0b767df2 100644
> --- a/test/e2e/utils.go
> +++ b/test/e2e/utils.go
> @@ -307,16 +307,17 @@ func buildPod(name, ns, image string, cmd []string) *corev1.Pod {
>  // Container's name, image as the Container's image and cmd as
>  // Container's command.
>  func buildContainer(name, image string, cmd []string) corev1.Container {
> +	falseVar := false
>  	return corev1.Container{
>  		Name:    name,
>  		Image:   image,
>  		Command: cmd,
>  		SecurityContext: &corev1.SecurityContext{
> -			AllowPrivilegeEscalation: pointer.Bool(false),
> +			AllowPrivilegeEscalation: &falseVar,
>  			Capabilities: &corev1.Capabilities{
>  				Drop: []corev1.Capability{"ALL"},
>  			},
> -			RunAsNonRoot: pointer.Bool(false),
> +			RunAsNonRoot: &falseVar,
>  			SeccompProfile: &corev1.SeccompProfile{
>  				Type: corev1.SeccompProfileTypeRuntimeDefault,
>  			},
> ```

Did this happen because the PR was not squashed?  Because maybe I picked the wrong commit out of the four?

brandisher:
> @brandisher Did this happen because the PR was not squashed? Because maybe I picked the wrong commit out of the four?

The PR was squashed to one commit (https://github.com/openshift/cluster-dns-operator/commit/864a9fad83885c008115f75f70dc665f8309cbd4) but the inclusion of `k8s.io/utils/pointer` happened prior to #339

candita:
I think we also need to backport https://github.com/openshift/cluster-dns-operator/pull/336

brandisher (commented):
/lgtm

brandisher:
Some minor failures and etcd issues; looks unrelated.

/test e2e-aws-single-node

PR #320, opened 2022-05-17
Bug 2061244: desiredDNSDaemonSet: Allow autoscaler to evict

Miciah:
Label DNS pods to enable the cluster autoscaler to evict them when scaling down a node.

Before this change, node scale-in could cause DNS disruption for tainted nodes.  If a node is tainted such that it has no DNS pod, then DNS queries originating on that node are forwarded to DNS pods on other nodes.  If one of these nodes was removed by scale-in or otherwise, the DNS pod on that node would be abruptly removed when the node shut down, and it would take some time for the node lifecycle controller to mark the pod as not ready, then for the endpoints controller and endpointslice controller to remove the pod's endpoint, and then for the service proxy to stop forwarding queries to the pod.  This delay could cause DNS lookup failures when queries were sent to a pod that had been on a removed node.  After this change, the cluster autoscaler should evict the DNS pod before removing the node, which should prevent disruption to the DNS service.

* `pkg/operator/controller/controller_dns_daemonset.go` (`desiredDNSDaemonSet`): Specify the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` pod label.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`): Verify that `desiredDNSDaemonSet` sets the expected pod labels.

PR #350, opened 2022-11-29
[release-4.10] OCPBUGS-2602: Allow autoscaler to evict DNS pods

Miciah:
#### `desiredDNSDaemonSet`: Allow autoscaler to evict

Annotate DNS pods to enable the cluster autoscaler to evict them when scaling down a node.

Before this change, node scale-in could cause DNS disruption for tainted nodes.  If a node is tainted such that it has no DNS pod, then DNS queries originating on that node are forwarded to DNS pods on other nodes.  If one of these nodes was removed by scale-in or otherwise, the DNS pod on that node would be abruptly removed when the node shut down, and it would take some time for the node lifecycle controller to mark the pod as not ready, then for the endpoints controller and endpointslice controller to remove the pod's endpoint, and then for the service proxy to stop forwarding queries to the pod.  This delay could cause DNS lookup failures when queries were sent to a pod that had been on a removed node.  After this change, the cluster autoscaler should evict the DNS pod before removing the node, which should prevent disruption to the DNS service.

* `pkg/operator/controller/controller_dns_daemonset.go` (`enableDaemonSetEvictionAnnotationKey`): New const for the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` pod annotation key.
(`managedDNSDaemonSetAnnotations`): New variable for annotation keys of annotations that the operator manages on DNS daemonsets.  For now, this only includes `enableDaemonSetEvictionAnnotationKey`.
(`desiredDNSDaemonSet`): Specify the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` pod annotation.
(`daemonsetConfigChanged`): Check whether any managed annotations changed, and update annotations as appropriate.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`): Verify that `desiredDNSDaemonSet` sets the expected pod annotations and labels.
(`TestDaemonsetConfigChanged`): Verify that `daemonsetConfigChanged` returns true if the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` pod annotation is changed and false if only some other annotation changed.

#### Keep dns-default annotations intact

When adding the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` annotation to DNS pods, do not overwrite other annotations.  In particular, preserve the `target.workload.openshift.io/management` annotation.

* `pkg/operator/controller/controller_dns_daemonset.go` (`targetWorkloadManagementAnnotationKey`): New const.
(`managedDNSDaemonSetAnnotations`): Add `targetWorkloadManagementAnnotationKey`.
(`desiredDNSDaemonSet`): Preserve annotations that are specified in the manifest.  
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`): Expect the `target.workload.openshift.io/management` annotation to be set.
(`TestDaemonsetConfigChanged`): Verify that the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` and `target.workload.openshift.io/management` annotations are managed.  
* `test/e2e/operator_test.go` (`TestOperatorRecreatesItsManagedAnnotations`): New test.  Verify that the operator sets the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` and `target.workload.openshift.io/management` annotations and restores them if they are removed.

--

This is a manual cherry-pick of #320 and #340.

### DIFF
--- a/pkg/operator/controller/controller_dns_daemonset.go
+++ b/pkg/operator/controller/controller_dns_daemonset.go
@@ -18,8 +18,22 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// enableDaemonSetEvictionAnnotationKey is the annotation key for the
+	// annotation that enables eviction of a daemonset by the cluster
+	// autoscaler.
+	enableDaemonSetEvictionAnnotationKey = "cluster-autoscaler.kubernetes.io/enable-ds-eviction"
+)
+
+var (
+	// managedDNSDaemonSetAnnotations is a set of annotation keys for
+	// annotations that the operator manages for the DNS daemonset.
+	managedDNSDaemonSetAnnotations = sets.NewString(enableDaemonSetEvictionAnnotationKey)
 )
 
 // ensureDNSDaemonSet ensures the dns daemonset exists for a given dns.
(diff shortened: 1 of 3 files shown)


gcs278:
@Miciah What's the reason we combined https://github.com/openshift/cluster-dns-operator/pull/320 and https://github.com/openshift/cluster-dns-operator/pull/340 together in a single backport? Is there a reason we can't or shouldn't individually backport each of them?

gcs278:
Otherwise, this is a just backport. I confirmed the deltas are as expected from the two commits and everything looks good.
/lgtm

Miciah:
> @Miciah What's the reason we combined #320 and #340 together in a single backport? Is there a reason we can't or shouldn't individually backport each of them?

Because #320 causes the bug that #340 fixes.  Backporting them individually would be more work and run the risk that we accidentally ship the bug that #320 causes.

candita:
> Nov 29 23:02:12.968 E ns/e2e-volumelimits-2938 pod/volume-limits-xwwsn node/ip-10-0-170-165.us-east-2.compute.internal container/volume-tester container exited with code 137 (Error): 

/retest-required

candita:
Backport approved by @Deepthidharwar 10/24/22.
/backport-risk-assessed

PR #358, opened 2023-03-23
OCPBUGS-5943: Set DNS DaemonSet's maxSurge value to 10%

Miciah:
Enable surge for DNS pods so that a rolling update of DNS pods creates new pods *before* deleting old pods.

This change reduces the risk of disruption to DNS during upgrades.  Without this change, a rolling update of DNS pods deletes a pod on a node first and then create a new pod on that node, meaning that a node that had a local DNS pod before an upgrade would temporarily not have one at some point during the upgrade.

Always having a local DNS pod reduces the risk of disruption, especially when using a CNI provider that supports topology-aware hints or when using openshift-sdn and ovn-kubernetes, which have logic similar to topology-aware hints to prefer any available local DNS pod rather than forwarding queries to DNS pods on other nodes.

This PR also sets `minReadySeconds` so that the old pod only gets deleted once the new pod has had time to start up.  Note that this may increase the time it takes for DNS to become available at install time by 9 seconds.

This PR additionally changes sets `maxUnavailable` to zero because `maxUnavailable` may not be non-zero when `maxSurge` is non-zero.

* `assets/dns/daemonset.yaml`: Set `minReadySeconds` to 9, `maxSurge` to 10%, and `maxUnavailable` to 0.
* `pkg/operator/controller/controller_dns_daemonset.go` (`daemonsetConfigChanged`): Check `minReadySeconds` and set it in the updated daemonset.
* `pkg/operator/controller/controller_dns_daemonset_test.go` (`TestDesiredDNSDaemonset`): Verify that `minReadySeconds` and the update strategy are set correctly.
(`TestDaemonsetConfigChanged`): Add test cases for `minReadySeconds` and `maxSurge`.


Miciah:
https://github.com/openshift/cluster-dns-operator/compare/f684ecd5ec4d4f74a592e2ced5cb9c0b9310edf9..8f97b16d057f2d58c7bea3830b437ee9e4ee22aa changes `maxUnavailable` to null because otherwise the API rejects the update strategy from as reported in [the DNS operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-single-node/1638953824020336640/artifacts/e2e-aws-ovn-single-node/gather-extra/artifacts/pods/openshift-dns-operator_dns-operator-7c88455c7f-q6wkn_dns-operator.log) from [the last e2e-aws-ovn-single-node run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-single-node/1638953824020336640):

    time="2023-03-23T18:03:13Z" level=error msg="failed to reconcile request /default: failed to ensure dns default: failed to ensure daemonset for dns default: failed to create dns daemonset openshift-dns/dns-default: DaemonSet.apps \"dns-default\" is invalid: spec.updateStrategy.rollingUpdate.maxSurge: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:\"10%\"}: may not be set when maxUnavailable is non-zero"

Miciah:
https://github.com/openshift/cluster-dns-operator/compare/8f97b16d057f2d58c7bea3830b437ee9e4ee22aa..b0ca398b84de1287fe02d22bf9311f0053a02d1c changes `maxUnavailable` to 0 because it defaults to 1 when unspecified.

assets/dns/daemonset.yaml
```diff
@@ -85,8 +87,11 @@ spec:
...
-      # Note: The daemon controller rounds the percentage up
-      # (unlike the deployment controller, which rounds down).
-      maxUnavailable: 10%
+      # Set maxSurge to a positive value so that each node that has a pod
+      # continues to have a local ready pod during a rolling update.  This is
+      # important for topology-aware hints as well as for similar logic in
+      # openshift-sdn and ovn-kubernetes that prefers to use a local ready DNS
+      # pod whenever one exists.
+      maxSurge: 10%
+      # maxUnavailable must be zero when maxSurge is nonzero.
```
gcs278, line 96: *nit* Is "must be zero" true? It is more accurate to say "should be zero"?

I don't see that in the docs, but I agree that it needs to be zero in order to fully use maxSurge, that makes sense. It sounds like you could potentially have both set per the docs, but I'd imagine that'd be an messy update.

pkg/operator/controller/controller_dns_daemonset_test.go
```diff
@@ -469,14 +481,23 @@ func TestDaemonsetConfigChanged(t *testing.T) {
...
-					},
-				}
+				daemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable = pointerTo(intstr.FromString("10%"))
+			},
+			expect: true,
+		},
+		{
+			description: "if the update strategy's max surge parameter changes",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				daemonset.Spec.UpdateStrategy.RollingUpdate.MaxSurge = pointerTo(intstr.FromString("10%"))
```
gcs278, line 493: I'm confused, this doesn't change the default value?

pkg/operator/controller/controller_dns_daemonset_test.go
```diff
@@ -469,14 +481,23 @@ func TestDaemonsetConfigChanged(t *testing.T) {
...
+			},
+			expect: true,
+		},
+		{
+			description: "if spec.minReadySeconds changes",
+			mutate: func(daemonset *appsv1.DaemonSet) {
+				daemonset.Spec.MinReadySeconds = 9
```
gcs278, line 500: I'm confused, this doesn't change the default value?

gcs278 (commented):
generally looks good to me, and seem likes a good improvement overall for DNS stability.

assets/dns/daemonset.yaml
```diff
@@ -33,7 +35,7 @@ spec:
             port: 8181
             scheme: HTTP
           initialDelaySeconds: 10
-          periodSeconds: 3
+          periodSeconds: 3 # Update minReadySeconds if you change this value!
```
candita, line 38:
```suggestion
          periodSeconds: 3 # Update the daemonset spec.minReadySeconds if you change this value!
```

assets/dns/daemonset.yaml
```diff
@@ -2,6 +2,8 @@ kind: DaemonSet
 apiVersion: apps/v1
 # name, namespace and labels are set at runtime
 spec:
+  # minReadySeconds should be 3x the readiness probe's polling interval.
+  minReadySeconds: 9
```
candita, line 6: The api [docs](https://docs.openshift.com/container-platform/4.12/rest_api/workloads_apis/daemonset-apps-v1.html#spec) say this is The minimum number of seconds for which a newly created DaemonSet pod should be ready without any of its container crashing, for it to be considered available. When it's 0 (default) the pod is immediately ready.  So, does this enforce a delay of 9s?

assets/dns/daemonset.yaml
```diff
@@ -2,6 +2,8 @@ kind: DaemonSet
 apiVersion: apps/v1
 # name, namespace and labels are set at runtime
 spec:
+  # minReadySeconds should be 3x the readiness probe's polling interval.
```
candita, line 5:
```suggestion
  # minReadySeconds should be 3x the readiness probe's polling interval (periodSec)
```

assets/dns/daemonset.yaml
candita: I agree with @gcs278 .  The docs say maxSurge can not be 0 if MaxUnavailable is 0 and vice versa. Can't they both be non-zero?

Also, 10% seems more like a minimum that we would want, if maxSurge means the maximum number of nodes with an existing available DaemonSet pod that can have an updated DaemonSet pod during during an update.  If we'd want to have a pod available **on every node** before we delete the previous one, wouldn't this need to be 100% ?  I may misunderstand.

pkg/operator/controller/controller_dns_daemonset_test.go
candita: L495 `expect: true` means it expects a change

pkg/operator/controller/controller_dns_daemonset_test.go
candita: I'm confused because it is expected to change, but that is not a change.

candita, line 500:
```suggestion
				daemonset.Spec.MinReadySeconds = pointerTo(intstr.FromInt(9))
```

assets/dns/daemonset.yaml
Miciah: Yes, `maxUnavailable` must be zero or else the API rejects the operator's attempt to create the daemonset: `DaemonSet.apps "dns-default" is invalid: spec.updateStrategy.rollingUpdate.maxSurge: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:"10%"}: may not be set when maxUnavailable is non-zero"`.  That error comes from here: <https://github.com/openshift/kubernetes/blob/54b552033300ab1f94ddf8a9d16032e6619556da/pkg/apis/apps/validation/validation.go#L400-L405>.  The docs are inaccurate.  Someday I might get around to submitting a PR to fix them.

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: The old value doesn't specify `maxSurge`: https://github.com/openshift/cluster-dns-operator/blob/b0ca398b84de1287fe02d22bf9311f0053a02d1c/pkg/operator/controller/controller_dns_daemonset_test.go#L623-L628

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: The old value doesn't specify `minReadySeconds`: https://github.com/openshift/cluster-dns-operator/blob/b0ca398b84de1287fe02d22bf9311f0053a02d1c/pkg/operator/controller/controller_dns_daemonset_test.go#L566-L621

assets/dns/daemonset.yaml
Miciah: Yes, I believe so.

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: `minReadySeconds` takes an integer value: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#daemonsetspec-v1-apps https://github.com/kubernetes/api/blob/ec59454f24f514bb472628bc0a5c8c9125218dbc/apps/v1/types.go#L391-L395

assets/dns/daemonset.yaml
Miciah: Updated in https://github.com/openshift/cluster-dns-operator/compare/b0ca398b84de1287fe02d22bf9311f0053a02d1c..043594d697605804dae0952b77c741270c0042bc.

assets/dns/daemonset.yaml
Miciah: Updated in https://github.com/openshift/cluster-dns-operator/compare/b0ca398b84de1287fe02d22bf9311f0053a02d1c..043594d697605804dae0952b77c741270c0042bc.

assets/dns/daemonset.yaml
Miciah: Added a note to the commit message in https://github.com/openshift/cluster-dns-operator/compare/b0ca398b84de1287fe02d22bf9311f0053a02d1c..043594d697605804dae0952b77c741270c0042bc.

pkg/operator/controller/controller_dns_daemonset_test.go
candita: Sorry, I confused it with MaxUnavailable.

pkg/operator/controller/controller_dns_daemonset_test.go
candita: So, the assets/dns/daemonset.yaml specifies `maxSurge`, but that isn't in a new `DaemonSetUpdateStrategy`?

pkg/operator/controller/controller_dns_daemonset_test.go
candita: So, this is sort of a test of an upgrade to this feature?  Because a new installation would add the "assets" from assets/dns/daemonset.yaml, right?

assets/dns/daemonset.yaml
candita: Does 9 seconds mean that it stays in Pending for 9s until it becomes Ready?  That seems troublesome as a default.

pkg/operator/controller/controller_dns_daemonset_test.go
gcs278: > The old value doesn't specify minReadySeconds

Ack. This isn't the first time this has tripped me up.

gcs278:
/approve
I will let @candita lgtm after she is done reviewing and questions are answered.

assets/dns/daemonset.yaml
Miciah: For reference, https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#upgrade-and-reconfiguration recommends `maxUnavailable: 10%` for daemonsets for components that have any impact on user workloads and `maxUnavailable: 33%` for daemonsets for components that don't have any such impact.  I am carrying that convention over to `maxSurge` (although maybe the risk of impact from higher percentage values is lower with `maxSurge`—the old pods should only get deleted when the new pods report ready, and if the new pods report ready, then the rollout will proceed anyway and eventually roll out 100% new pods).

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: `maxSurge` isn't in the daemonset that is used in the test.  The daemonset that `TestDaemonsetConfigChanged` uses differs from the daemonset that `desiredDNSDaemonSet` returns.

pkg/operator/controller/controller_dns_daemonset_test.go
Miciah: > So, this is sort of a test of an upgrade to this feature? Because a new installation would add the "assets" from assets/dns/daemonset.yaml, right?

Yeah, that's one way to look at it.  The point of the test case is to verify that `TestDaemonsetConfigChanged` behaves correctly when the `minReadySeconds` field changes.

assets/dns/daemonset.yaml
Miciah: The value of `minReadySeconds` is the difference between when a pod is ready and when it is considered "available" (see <https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec>).  The pod should report "Running" once the container is started, and it should report 2/2 containers ready, same as before.  As far as I can tell, `minReadySeconds` just affects status reporting on the daemonset (viz. `status.numberAvailable` and `status.numberUnavailable`) and behavior of rolling updates.

Miciah: Based on my previous comment, I believe that the note that I added in the commit message in https://github.com/openshift/cluster-dns-operator/compare/b0ca398b84de1287fe02d22bf9311f0053a02d1c..043594d697605804dae0952b77c741270c0042bc is inaccurate, so I have removed it in https://github.com/openshift/cluster-dns-operator/compare/043594d697605804dae0952b77c741270c0042bc..5c4f0ad3430ac1195e990c20398e62bc1681796e.

Miciah:
[e2e-aws-ovn](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn/1641905255799066624) and [e2e-aws-ovn-single-node failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-single-node/1641905255891341312) because `[sig-arch] Managed cluster should only include cluster daemonsets that have maxUnavailable update of 10 or 33 percent` failed, which https://github.com/openshift/origin/pull/27819 will address.

[e2e-aws-ovn-upgrade failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-upgrade/1641905255920701440) because a control-plane node did not come up.  
/test e2e-aws-ovn-upgrade

Miciah:
https://github.com/openshift/origin/pull/27819 has merged.
/hold cancel

Miciah:
[e2e-aws-ovn failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn/1643685310078390272) because the control-plane nodes didn't all come up.  
/test e2e-aws-ovn

Miciah:
[e2e-aws-ovn failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn/1644039548138688512) because one of the control-plane nodes failed to drain: `Drain operation currently blocked by: [{Name:EtcdQuorumOperator Owner:clusteroperator/etcd}]` (from [`machines.json`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn/1644039548138688512/artifacts/e2e-aws-ovn/gather-extra/artifacts/machines.json)).  This looks a lot like https://issues.redhat.com/browse/OCPBUGS-11004, but that bug was recently verified.  
/test e2e-aws-ovn

Miciah:
[e2e-aws-ovn-operator  failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/358/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-operator/1645315681710772224) because `TestDNSLogging` failed.
/test e2e-aws-ovn-operator

Miciah:
I think we want both #358 (this PR) and #364 in 4.13.  
/cherry-pick release-4.13

PR #365, opened 2023-04-27
OCPBUGS-12859: deflake TestDNSLogging

frobware:
Previously, the test started a pod to issue 'dig' queries from and then fetched the list of coredns pods. For each coredns pod, a 'dig' request was made, and logs for the "dns" container were retrieved to assert that logging was both enabled and working by matching the expected dig query failure (because the host queried does not exist) in the logging output from coredns.

However, queries from any pod will use/target the coredns pod on the same node, leading to potential failures when the first checked coredns pod isn't on the same node as the test pod.

The test has been reworked to remove the test pod and execute the following command for each coredns pod found:

    oc exec <coredns-pod> -- dig @localhost -p 5353 <hostname>

Existing test assertions remain unchanged; the "dns" container logs are retrieved, and the test passes if a logged entry for `<hostname>` is found.

I have introduced two new functions `lookForStringInPodExecOneShot` and `lookForSubStringsInPodLogOneShot`. These are largely identical to the existing functions `lookForStringInPodExec` and `lookForSubStringsInPodLog` respectively. The real difference is that they don't retry (via `wait.PollImmediate`) on failure, leaving the decision up to the caller of these complementary `...OneShot` functions. 

Future work: the existing functions that insist on always retrying could be reworked to use the `...OneShot` functions but I deliberately chose not to do that in this PR to a) reduce the review footprint, and b) to not perturb the existing tests.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -190,7 +190,7 @@ func TestDNSLogging(t *testing.T) {
 
 	// Ensure that DNS is stable before starting the test, otherwise we'd need to tweak individual test durations.
 	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
-		t.Errorf("expected default DNS pods to be available: %v", err)
+		t.Fatalf("expected default DNS pods to be available: %v", err)
 	}
 
 	// Get the CoreDNS image used by the test upstream resolver.
@@ -232,13 +232,13 @@ func TestDNSLogging(t *testing.T) {
 
 	// Verify that default DNS pods are all available before inspecting them.
 	if err := waitForDNSConditions(t, cl, 5*time.Minute, dnsName, defaultAvailableDNSConditions...); err != nil {
-		t.Errorf("expected default DNS pods to be available: %v", err)
+		t.Fatalf("expected default DNS pods to be available: %v", err)
 	}
 
 	// Verify that the Corefile of DNS DaemonSet pods have been updated.
 	dnsDaemonSet := &appsv1.DaemonSet{}
 	if err := cl.Get(context.TODO(), operatorcontroller.DNSDaemonSetName(defaultDNS), dnsDaemonSet); err != nil {
-		_ = fmt.Errorf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
+		t.Fatalf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
 	}
 	selector, err := metav1.LabelSelectorAsSelector(dnsDaemonSet.Spec.Selector)
 	if err != nil {
@@ -332,7 +332,7 @@ func TestDNSLogging(t *testing.T) {
 
 	dnsOperatorDeployment := &appsv1.Deployment{}
 	if err := cl.Get(context.TODO(), operatorcontroller.DefaultDNSOperatorDeploymentName(), dnsOperatorDeployment); err != nil {
-		_ = fmt.Errorf("failed to get deployment %s/%s: %v", dnsOperatorDeployment.Namespace, dnsOperatorDeployment.Name, err)
+		t.Fatalf("failed to get deployment %s/%s: %v", dnsOperatorDeployment.Namespace, dnsOperatorDeployment.Name, err)
 	}
 	operatorSelector, err := metav1.LabelSelectorAsSelector(dnsOperatorDeployment.Spec.Selector)
 	if err != nil {
@@ -637,7 +637,7 @@ func TestDNSForwarding(t *testing.T) {
 	// Verify that the Corefile of DNS DaemonSet pods have been updated.
 	dnsDaemonSet := &appsv1.DaemonSet{}
 	if err := cl.Get(context.TODO(), operatorcontroller.DNSDaemonSetName(defaultDNS), dnsDaemonSet); err != nil {
-		_ = fmt.Errorf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
+		t.Fatalf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
 	}
 	selector, err := metav1.LabelSelectorAsSelector(dnsDaemonSet.Spec.Selector)
 	if err != nil {
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -250,79 +250,51 @@ func TestDNSLogging(t *testing.T) {
 	}
 	catCmd := []string{"cat", "/etc/coredns/Corefile"}
 
-	// Get the openshift-cli image.
-	var (
-		cliImage      string
-		cliImageFound bool
-	)
-	for _, ver := range co.Status.Versions {
-		if ver.Name == statuscontroller.OpenshiftCLIVersionName {
-			if len(ver.Version) == 0 {
-				break
-			}
-			cliImage = ver.Version
-			cliImageFound = true
-			break
-		}
+	if len(coreDNSPods.Items) == 0 {
+		t.Fatalf("found zero coredns pods!")
 	}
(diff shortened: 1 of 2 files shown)


test/e2e/operator_test.go
```diff
@@ -238,7 +238,7 @@ func TestDNSLogging(t *testing.T) {
 	// Verify that the Corefile of DNS DaemonSet pods have been updated.
 	dnsDaemonSet := &appsv1.DaemonSet{}
 	if err := cl.Get(context.TODO(), operatorcontroller.DNSDaemonSetName(defaultDNS), dnsDaemonSet); err != nil {
-		_ = fmt.Errorf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
+		t.Errorf("failed to get daemonset %s/%s: %v", dnsDaemonSet.Namespace, dnsDaemonSet.Name, err)
```
gcs278, line 241: fatal?

frobware: Done in https://github.com/openshift/cluster-dns-operator/pull/365/commits/440ae76b555bd3a5dc774158a10e8b8e431903d7.

test/e2e/operator_test.go
```diff
@@ -250,79 +250,40 @@ func TestDNSLogging(t *testing.T) {
 	}
 	catCmd := []string{"cat", "/etc/coredns/Corefile"}
```
gcs278, line 251: *nit* I think naming this variable would help understand the logic.

```suggestion
	catCorefileCmd := []string{"cat", "/etc/coredns/Corefile"}
```

test/e2e/operator_test.go
gcs278: Still shows `t.Errorf` for me, am I missing something?

test/e2e/operator_test.go
```diff
@@ -250,79 +250,40 @@ func TestDNSLogging(t *testing.T) {
...
-		}
+			host := fmt.Sprintf("test-%s-%v.svc.cluster.local", corednspod.Name, time.Now().UnixMicro())
+			digCmd := []string{"dig", "@127.0.0.1", "-p", "5353", host}
+			found, err = lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", digCmd, "NXDOMAIN")
+			if err != nil {
+				t.Logf("%s/%s: dig error: %v, retrying...", corednspod.Namespace, corednspod.Name, err)
+				return false, err
+			}
+			if !found {
+				t.Logf("%s%s: failed to match 'NXDOMAIN', retrying...", corednspod.Namespace, corednspod.Name)
```
gcs278, line 273:
```suggestion
				t.Logf("%s/%s: failed to match 'NXDOMAIN', retrying...", corednspod.Namespace, corednspod.Name)
```

test/e2e/utils.go
```diff
@@ -37,20 +38,49 @@ import (
...
-		return err
+	result, err := runCmd(cmdPath, args)
+	if err != nil {
+		return false, err
 	}
-	return nil
+	if !strings.Contains(result, expectedString) {
+		return false, nil
+	}
+	return true, nil
```
gcs278, line 83:
```suggestion
	return strings.Contains(result, expectedString), nil
```

gcs278 (commented):
mainly nit picks

test/e2e/operator_test.go
```diff
@@ -250,79 +250,40 @@ func TestDNSLogging(t *testing.T) {
...
-		// Dig the example dns forwarding host.
-		digCmd := []string{"dig", "test.svc.cluster.local"}
-		if err := lookForStringInPodExec(testClientForDNSLogging.Namespace, testClientForDNSLogging.Name, testClientForDNSLogging.Name, digCmd, "NXDOMAIN", 2*time.Minute); err != nil {
-			t.Fatalf("failed to dig %v", err)
-		}
-
-		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 5*time.Minute); err != nil {
-			t.Fatalf("failed to set Debug logLevel for operator %s: %v", opName, err)
-		}
+			host := fmt.Sprintf("test-%s-%v.svc.cluster.local", corednspod.Name, time.Now().UnixMicro())
```
gcs278, line 265: Related to my above comment, maybe it's best to drop some more block comments about what we are doing:
```suggestion
                       // Run a dig query that locally queries this DNS pod (localhost) to trigger a log entry in the CoreDNS pod.
			host := fmt.Sprintf("test-%s-%v.svc.cluster.local", corednspod.Name, time.Now().UnixMicro())
```

test/e2e/operator_test.go
```diff
@@ -250,79 +250,40 @@ func TestDNSLogging(t *testing.T) {
...
+				t.Logf("%s%s: failed to match 'NXDOMAIN', retrying...", corednspod.Namespace, corednspod.Name)
+				return false, nil
+			}
 
-		if found == 0 {
-			if err := lookForSubStringsInPodLog(corednspod.Namespace, corednspod.Name, "dns", 2*time.Minute, "A IN test.svc.cluster.local.", "NXDOMAIN"); err != nil {
-				found = 0
-			} else {
-				found = 1
+			found, err = lookForSubStringsInPodLog(corednspod.Namespace, corednspod.Name, "dns", fmt.Sprintf("A IN %s.", host), "NXDOMAIN")
```
gcs278, line 277:
```suggestion
                       // Look for the dig query we just executed in the logs to confirm logging is turned on.
			found, err = lookForSubStringsInPodLog(corednspod.Namespace, corednspod.Name, "dns", fmt.Sprintf("A IN %s.", host), "NXDOMAIN")
```

test/e2e/utils.go
```diff
@@ -103,6 +135,7 @@ func lookForString(cmd string, args []string, expectedString string, timeout tim
 	}
 	return nil
 }
+
 func lookForSubStrings(cmd string, args []string, timeout time.Duration, expectedStrings []string) (bool, error) {
```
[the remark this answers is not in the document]
frobware: Gone (in a future push)

test/e2e/operator_test.go
```diff
@@ -248,81 +248,45 @@ func TestDNSLogging(t *testing.T) {
...
-		// Dig the example dns forwarding host.
-		digCmd := []string{"dig", "test.svc.cluster.local"}
-		if err := lookForStringInPodExec(testClientForDNSLogging.Namespace, testClientForDNSLogging.Name, testClientForDNSLogging.Name, digCmd, "NXDOMAIN", 2*time.Minute); err != nil {
-			t.Fatalf("failed to dig %v", err)
-		}
-
-		if err := lookForStringInPodExec(corednspod.Namespace, corednspod.Name, "dns", catCmd, "class denial error", 5*time.Minute); err != nil {
-			t.Fatalf("failed to set Debug logLevel for operator %s: %v", opName, err)
-		}
+                       // Run a dig query that locally queries this DNS pod (localhost) to trigger a log entry in the CoreDNS pod.
```
gcs278, line 266: should be indented not spaces (go fmt?)

frobware: I reformatted after taking your suggestions.

frobware:
> mainly nit picks

- Many, many thanks for the quick review
- Took 90% of your suggestions, left `cmdCmd` as is to reduce the review footprint
- Going to let is soak in CI to verify that change

gcs278:
Thanks for the responses. Have you seen: https://github.com/openshift/cluster-dns-operator/pull/365/files#r1179575696. Did something not get updated, I still see `t.Errorf`. Shouldn't that be `t.Fatalf`?

gcs278:
Yea I can understand you not wanting to perturb the existing tests. Though, I do feel the functions are named a bit awkwardly, like `lookForStringInPodExec` repeats, but `lookForStringInPodExecOneShot` fires once. By default, I'd think `lookForStringInPodExec` should look once, but I won't hang up the PR over for you to refactor everything. I think someone needs get in here and clean up these functions someday.

Maybe we could eventually use some standard function like openshift/origin does: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/pod/resource.go#L360

/lgtm

PR #375, opened 2023-07-26
OCPBUGS-18034: Only bump lastTransitionTime on 'status' changes

wking:
From [the API docs][1]:

    lastTransitionTime is the time of the last update to the current status property.

And [the library-go `SetStatusCondition` implementation][2] includes:

```go
  if existingCondition.Status != newCondition.Status {
    existingCondition.Status = newCondition.Status
    existingCondition.LastTransitionTime = metav1.NewTime(time.Now())
  }

  existingCondition.Reason = newCondition.Reason
  existingCondition.Message = newCondition.Message
```

The motivation for that behavior is that it's often more useful to know "how long has this resource been `Degraded=True`?" and similar than it is to know how long it has had exactly the same `status`/`reason`/`message`. Messages in particular can be fairly mutable (`# of # pods available`, etc.), and changes there do not necessarily represent fundamental shifts in the underlying issue.

[1]: https://github.com/openshift/api/blob/81f778f3b3ec31c1dd344e795620a8fbaf2d9a51/config/v1/types_cluster_operator.go#L129
[2]: https://github.com/openshift/library-go/blob/c515269de16e5e239bd6e93e1f9821a976bb460b/pkg/config/clusteroperator/v1helpers/status.go#L29C1-L35C50

### DIFF
--- a/pkg/operator/controller/dns_status.go
+++ b/pkg/operator/controller/dns_status.go
@@ -265,8 +265,7 @@ func computeDNSUpgradeableCondition(oldCondition *operatorv1.OperatorCondition, 
 // setDNSLastTransitionTime sets LastTransitionTime for the given condition.
 // If the condition has changed, it will assign a new timestamp otherwise keeps the old timestamp.
 func setDNSLastTransitionTime(condition, oldCondition *operatorv1.OperatorCondition) operatorv1.OperatorCondition {
-	if oldCondition != nil && condition.Status == oldCondition.Status &&
-		condition.Reason == oldCondition.Reason && condition.Message == oldCondition.Message {
+	if oldCondition != nil && condition.Status == oldCondition.Status {
 		condition.LastTransitionTime = oldCondition.LastTransitionTime
 	} else {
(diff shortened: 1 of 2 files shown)


wking:
[Single-node CI failure][1]:

```
: [sig-arch] Cluster topology single node tests Verify that OpenShift components deploy one replica in SingleReplica topology mode [Suite:openshift/conformance/parallel]
--
fail [github.com/openshift/origin/test/extended/single_node/topology.go:97]: machine-os-builder in openshift-machine-config-operator namespace expected to have 1 replica but got 0
Expected
    <int>: 0
to equal
    <int>: 1
Ginkgo exit error 1: exit with code 1
```

seems orthogonal to this change.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-dns-operator/375/pull-ci-openshift-cluster-dns-operator-master-e2e-aws-ovn-single-node/1684294268409090048

candita:
/retitle OCPBUGS-18034: Only bump lastTransitionTime on 'status' changes

PR #376, opened 2023-07-26
pkg/operator/controller/dns_status: Allow 10% unavailable before going Degraded=True

wking:
Layering a new one-line pivot on top of #375, to see how they play together.

5c4f0ad3430a (#358) pivoted the dns-default DaemonSet from 10% `maxUnavailable` to 10% `maxSurge`.  That left this "is this DaemonSet sad?" checker defaulting to "if there are any more than one unready DNS pods, we're sad".  But in many clusters, especially large ones with the DaemonSet trying to schedule pods on many compute nodes, there are often more than 1 unready pod, and going `Degraded=True` is distracting noise, and not something that has a quality-of-service impact.  This commit adjusts the default to the old 10% `maxUnavailable`, so we have a less twitchy value we can use for surging DaemonSets.  We can't set that on the DaemonSet itself, because as described in 5c4f0ad3430a, API validation prevents us from setting `maxSurge` and `maxUnavailable` simultaneously.


wking:
#358 was reverted in #379, which makes this change obsolete.

PR #377, opened 2023-07-26
pkg/operator/controller/status: 5 minutes of inertia before propagating Degraded=True

wking:
Layering on top of both #375 ~and #376~.

The default DNS resource can go `Degraded=True` when it has insufficient available pods.  But it's possible to have insufficient available pods momentarily and subsequently recover.  For example, while the DaemonSet is rolling out an update, we expect `maxSurge` unavailable pods for the duration of the rollout, which could involve up to 10 (for 10% `maxSurge`) rounds of pods being spun up on the cluster's nodes:

1. DaemonSet bumped.
2. Pod a2 launched on node a, pod b2 launched on node b, etc.
3. Pod a2 goes ready, and the old pod a1 is deleted.
4. Pod c2 launched on node c.
5. Pod b2 goes ready, and the  old pod b1 is deleted.
...
n. Eventually all the new pods are ready.

During that rollout, there are always some unready pods.  But we're still making quick progress and things are happy. 
  However, we could also have:

1. DaemonSet bumped.
2. Pod a2 launched on node a, pod b2 launched on node b, etc.
3. Pod b2 goes ready, and the  old pod b1 is deleted.
4. Pod c2 launched on node c.
...
n. Pod a2 still stuck.

That will have a similar number of unready pods as the happy case, but a2 being stuck for so long is a bad sign, and might deserve admin intervention.  Ideally the DaemonSet controller would be watching individual pods and reporting conditions to let us know if it was concerned about progress or recovery from external disruption.  But [DaemonSet `status` has no conditions today][1].  We could look over the DaemonSet controller's shoulder and watch the pods directly, but that would be a lot of work.  So instead I'm adding a few minutes of inertia here, assuming that if the unready pods quickly resolve, it's unlikely to impact quality-of-service or require admin intervention.  And if the unready pods (or other issue) does not quickly resolve, it is likely to deserve admin intervention (now with some hopefully acceptable additional latency before summoning the admin).

[1]: https://github.com/kubernetes/kubernetes/blob/98358b8ce11b0c1878ae7aa1482668cb7a0b0e23/staging/src/k8s.io/api/apps/v1/types.go#L722

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -417,7 +418,7 @@ func computeOperatorDegradedCondition(haveDNS bool, dns *operatorv1.DNS) configv
 
 	var degraded bool
 	for _, cond := range dns.Status.Conditions {
-		if cond.Type == operatorv1.OperatorStatusTypeDegraded && cond.Status == operatorv1.ConditionTrue {
+		if cond.Type == operatorv1.OperatorStatusTypeDegraded && cond.Status == operatorv1.ConditionTrue && time.Now().Sub(cond.LastTransitionTime.Time) > 5*time.Minute {
 			degraded = true
 		}
 	}
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -625,13 +625,13 @@ func mergeConditions(conditions []configv1.ClusterOperatorStatusCondition, updat
 		for j, cond := range conditions {
 			if cond.Type == update.Type {
 				add = false
-				if conditionChanged(cond, update) {
-					conditions[j].Status = update.Status
-					conditions[j].Reason = update.Reason
-					conditions[j].Message = update.Message
+				if conditions[j].Status != update.Status {
 					conditions[j].LastTransitionTime = now
-					break
 				}
+				conditions[j].Status = update.Status
+				conditions[j].Reason = update.Reason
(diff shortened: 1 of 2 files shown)


wking:
I've pushed 35c37f2d -> c55f5d27, dropping #376's `maxUnavailable` backstopping now that #379 has returned the DaemonSet to using `maxUnavailable` explicitly.

candita:
Unrelated to change:

> {  fail [github.com/openshift/origin/test/extended/oauth/requestheaders.go:218]: full response header: HTTP/1.1 403 Forbidden
> ...
> {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"forbidden: User \"system:anonymous\" cannot get path \"/metrics\"","reason":"Forbidden","details":{},"code":403}
> ...
> Expected
>     <string>: 403 Forbidden
> to contain substring
>     <string>: 401 Unauthorized
> 

/test e2e-aws-ovn-serial

pkg/operator/controller/status/controller.go
```diff
@@ -417,7 +418,7 @@ func computeOperatorDegradedCondition(haveDNS bool, dns *operatorv1.DNS) configv
 
 	var degraded bool
 	for _, cond := range dns.Status.Conditions {
-		if cond.Type == operatorv1.OperatorStatusTypeDegraded && cond.Status == operatorv1.ConditionTrue {
+		if cond.Type == operatorv1.OperatorStatusTypeDegraded && cond.Status == operatorv1.ConditionTrue && time.Now().Sub(cond.LastTransitionTime.Time) > 5*time.Minute {
```
candita, line 421: We set the degraded condtition based on transition time in https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/controller/dns_status.go#L130, but I don't see a need to check `lastTransitionTime` again in this code.

candita:
@wking I've approved #375 separately.

PR #384, opened 2023-10-03
OCPBUGS-20024: Revert "Revert "Set DNS DaemonSet's maxSurge value to 10%""

Miciah:
The defect in the daemon controller, [OCPBUGS-19452](https://issues.redhat.com/browse/OCPBUGS-19452) that prevented us from using `maxSurge` has been fixed by https://github.com/openshift/kubernetes/pull/1716.

This reverts #379.

PR #386, opened 2023-10-12
OCPBUGS-20024: Ignore max unavailable for status

candita:
Followup to PR #384.
- pkg/operator/controller/controller_dns_node_resolver_daemonset.go - small update to a comment
- pkg/operator/controller/dns_status.go - hardcode maxUnavailable to 10% of desiredNumberScheduled and remove condition "invalid maxUnavailable value"
- pkg/operator/controller/dns_status_test.go - remove maxUnavailable format testing

### DIFF
--- a/pkg/operator/controller/dns_status.go
+++ b/pkg/operator/controller/dns_status.go
@@ -100,11 +100,9 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 		want := dnsDaemonset.Status.DesiredNumberScheduled
 		have := dnsDaemonset.Status.NumberAvailable
 		numberUnavailable := want - have
-		maxUnavailableIntStr := intstr.FromInt(1)
-		if dnsDaemonset.Spec.UpdateStrategy.RollingUpdate != nil && dnsDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable != nil {
-			maxUnavailableIntStr = *dnsDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable
-		}
+		maxUnavailableIntStr := intstr.FromString("10%")
 		maxUnavailable, intstrErr := intstr.GetScaledValueFromIntOrPercent(&maxUnavailableIntStr, int(want), true)
+
 		switch {
 		case want == 0:
 			status = operatorv1.ConditionTrue
@@ -115,6 +113,7 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 			degradedReasons = append(degradedReasons, "NoDNSPodsAvailable")
 			messages = append(messages, "No DNS pods are available.")
 		case intstrErr != nil:
+			// This should not happen, but is included just to safeguard against future changes.
 			degradedReasons = append(degradedReasons, "InvalidDNSMaxUnavailable")
 			messages = append(messages, fmt.Sprintf("The DNS daemonset has an invalid MaxUnavailable value: %v", intstrErr))
 		case int(numberUnavailable) > maxUnavailable:
--- a/pkg/operator/controller/dns_status_test.go
+++ b/pkg/operator/controller/dns_status_test.go
@@ -14,6 +14,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+var (
+	maxUnavailable = intstr.FromInt(1)
+)
+
 func TestDNSStatusConditions(t *testing.T) {
 	type testIn struct {
 		haveClusterIP                   bool
@@ -75,7 +79,6 @@ func TestDNSStatusConditions(t *testing.T) {
 		if tc.inputs.haveClusterIP {
 			clusterIP = "1.2.3.4"
 		}
-		maxUnavailable := intstr.FromInt(1)
 		var dnsDaemonset, nodeResolverDaemonset *appsv1.DaemonSet
 		if tc.inputs.haveDNS {
 			dnsDaemonset = &appsv1.DaemonSet{
@@ -201,7 +204,7 @@ func TestDNSStatusConditions(t *testing.T) {
 // TestComputeDNSDegradedCondition verifies the computeDNSDegradedCondition has
 // the expected behavior.
 func TestComputeDNSDegradedCondition(t *testing.T) {
-	makeDaemonSet := func(desired, available int, maxUnavailable intstr.IntOrString) *appsv1.DaemonSet {
+	makeDaemonSet := func(desired, available int) *appsv1.DaemonSet {
 		return &appsv1.DaemonSet{
 			Spec: appsv1.DaemonSetSpec{
 				UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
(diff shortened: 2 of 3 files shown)


candita:
> {"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2023-10-12T19:23:38Z"}

/test e2e-aws-ovn

candita:
> could not initialize namespace: failed to wait for authentication cache to warm up after 15s: projects.project.openshift.io "ci-op-tbmipxqy" is forbidden: User "system:serviceaccount:ci:ci-operator" cannot get resource "projects" in API group "project.openshift.io" in the namespace "ci-op-tbmipxqy"

/test e2e-aws-ovn

candita:
> After 4 hours:
> Received signal.                              signal=interrupt
> INFO[2023-10-12T19:23:38Z] error: Process interrupted with signal interrupt, cancelling execution... 
> INFO[2023-10-12T19:23:38Z] cleanup: Deleting release pod release-latest 

/test e2e-aws-ovn

candita:
@Miciah PTAL when you get a chance.

pkg/operator/controller/dns_status_test.go
```diff
@@ -224,122 +229,80 @@ func TestComputeDNSDegradedCondition(t *testing.T) {
...
-			nrDaemonset:  makeDaemonSet(6, 6, intstr.FromString("33%")),
-			expected:     operatorv1.ConditionUnknown,
-		},
-		{
-			name:         "node-resolver invalid MaxUnavailable (string with letters) is ok",
-			clusterIP:    "172.30.0.10",
-			dnsDaemonset: makeDaemonSet(6, 6, intstr.FromString("10%")),
-			nrDaemonset:  makeDaemonSet(6, 6, intstr.IntOrString{Type: intstr.String, StrVal: "TEST"}),
-			expected:     operatorv1.ConditionFalse,
-		},
```
Miciah, line 267: It's interesting that we still had these test cases for invalid `maxUnavailable` values for the node-resolver daemonset.  We kept these test cases when we changed the status reporting to ignore the node-resolver daemon set in #273.  I would guess the reasoning was that we wanted the test cases to verify that the status reporting code really was not basing status on the daemon set's`maxUnavailable` parameter.  Does the same reasoning still apply?  If so, it would make sense to keep (some of) these test cases.

pkg/operator/controller/dns_status.go
```diff
@@ -100,11 +100,9 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
 		want := dnsDaemonset.Status.DesiredNumberScheduled
 		have := dnsDaemonset.Status.NumberAvailable
 		numberUnavailable := want - have
-		maxUnavailableIntStr := intstr.FromInt(1)
-		if dnsDaemonset.Spec.UpdateStrategy.RollingUpdate != nil && dnsDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable != nil {
-			maxUnavailableIntStr = *dnsDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable
-		}
-		maxUnavailable, intstrErr := intstr.GetScaledValueFromIntOrPercent(&maxUnavailableIntStr, int(want), true)
+		maxUnavailableIntStr := intstr.FromString("10%")
+		maxUnavailable, _ := intstr.GetScaledValueFromIntOrPercent(&maxUnavailableIntStr, int(want), true)
```
Miciah, line 104: I'm wary of ignoring error values.  Even if an analysis of the code today determines that the error value will always be nil, it's possible that this code or the definition of `GetScaledValueFromIntOrPercent` could change in the future.

Miciah (commented):
The fundamental change is good.  I have some questions about tests and error handling.  

Also, regarding the title, `maxUnavailable` was never modifiable by the cluster-admin.  Maybe it would be clearer to say, "Ignore max unavailable for status" or, "Report degraded if >10% unavailable".

pkg/operator/controller/dns_status_test.go
candita: It seems to be an oversight that I didn't clean these units tests then.  We entirely removed the checking of node resolver daemonset and its `maxUnavailable` parameter. It doesn't make sense anymore to check anything having to do with node resolver when computing the degraded condition: https://github.com/openshift/cluster-dns-operator/pull/273/files#diff-32495132facf7e0819a407af732514958b198d9e40236656bc43b093345d1539L111

pkg/operator/controller/dns_status.go
candita: ok

pkg/operator/controller/dns_status_test.go
candita: In fact, I should remove the `nrDaemonset` field from the test cases.  Does that make more sense?

Miciah: > It seems to be an oversight that I didn't clean these units tests then. We entirely removed the checking of node resolver daemonset and its `maxUnavailable` parameter. It doesn't make sense anymore to check anything having to do with node resolver when computing the degraded condition

Are you sure it was an oversight?  My speculation was that it was intentional:  If parameter X *did* affect result Y, and you are changing the logic so that X *doesn't* affect Y, then it _does_ make sense for tests to verify that X doesn't affect Y, right?  More generally, I'd rather err on the side of keeping possibly redundant test cases over risking removing test cases that could possibly detect faults.  However, I'll defer to your judgment if you decide these test cases really should be removed and the test simplified.

candita: I am quite sure it was an oversight.  I removed the parameters `haveNodeResolverDaemonset` and `nodeResolverDaemonset` from the signature of `computeDNSDegradedCondition`, because we no longer use them in the function.  We still use them in `computeDNSProgressingCondition`, but not `computeDNSDegradedCondition`.

pkg/operator/controller/dns_status.go
```diff
@@ -100,23 +100,22 @@ func computeDNSDegradedCondition(oldCondition *operatorv1.OperatorCondition, clu
...
+
 		switch {
 		case want == 0:
 			status = operatorv1.ConditionTrue
 			degradedReasons = append(degradedReasons, "NoDNSPodsDesired")
 			messages = append(messages, "No DNS pods are desired; this could mean all nodes are tainted or unschedulable.")
+		case intstrErr != nil:
+			// This should not happen, but is included just to safeguard against future changes.
+			degradedReasons = append(degradedReasons, "InvalidDNSMaxUnavailable")
+			messages = append(messages, fmt.Sprintf("The DNS daemonset has an invalid MaxUnavailable value: %v", intstrErr))
```
Miciah, line 114: Was there a reason to move this case up?

pkg/operator/controller/dns_status_test.go
```diff
@@ -14,6 +14,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+var (
+	maxUnavailable = intstr.FromInt(1)
```
Miciah, line 18: Is there a reason to declare `maxUnavailable` here instead of inside `makeDaemonSet`?  We usually tend towards making tests and helpers self-contained.

pkg/operator/controller/dns_status_test.go
```diff
@@ -201,7 +205,8 @@ func TestDNSStatusConditions(t *testing.T) {
 // TestComputeDNSDegradedCondition verifies the computeDNSDegradedCondition has
 // the expected behavior.
 func TestComputeDNSDegradedCondition(t *testing.T) {
-	makeDaemonSet := func(desired, available int, maxUnavailable intstr.IntOrString) *appsv1.DaemonSet {
+
```
Miciah, line 208: Extra blank line?

Miciah (commented):
A couple minor things.

pkg/operator/controller/dns_status_test.go
candita: It is used in multiple functions and I removed it as a parameter to makeDaemonSet, so I declared it globally.  It's used in TestDNSStatusConditions, TestComputeDNSDegradedCondition, TestComputeDNSProgressingCondition, and TestSkippingStatusUpdates. 

For consistency, removed an extra declaration of it in TestDNSStatusConditions.

pkg/operator/controller/dns_status_test.go
candita: Fixed.  Forgot to push.

pkg/operator/controller/dns_status.go
candita: Fixed, forgot to push.

pkg/operator/controller/dns_status_test.go
Miciah: All right, I can live with that.  Test coverage is still good overall.

pkg/operator/controller/dns_status_test.go
Miciah: All right, I won't insist on changing it.

Miciah:
Thanks!
/approve
/lgtm

PR #393, opened 2023-11-08
CFE-852: Enable ocp_dnsnameresolver CoreDNS plugin

arkadeepsen:
This PR makes the necessary changes to the Corefile of CoreDNS and adds the necessary RBAC permissions to CoreDNS to enable ocp_dnsnameresolver CoreDNS plugin. This is as per the enhancement proposal openshift/enhancements#1335.


rfredette (commented):
One nit, otherwise it all looks good to me.

pkg/operator/operator.go
```diff
@@ -36,6 +42,57 @@ type Operator struct {
...
+		logrus.Info("FeatureGates initialized", "knownFeatures", featureGates.KnownFeatures())
+	case <-time.After(1 * time.Minute):
+		logrus.Error(nil, "timed out waiting for FeatureGate detection")
+		return nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
arkadeepsen: This is how it is advised to fetch the feature gates: https://docs.google.com/document/d/1zOL38_KDKwAvsx-LMHfyDdX4-jq3HQU1YWBjfuHaM0Q/edit#heading=h.o5ki6e7q871r

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -68,6 +68,11 @@ var corefileTemplate = template.Must(template.New("Corefile").Funcs(template.Fun
     cache {{ $.PositiveTTL }} {
         denial 9984 {{ $.NegativeTTL }}
     }
+    {{- if eq true $.OCPDNSNameResolver}}
+    ocp_dnsnameresolver {
+        namespaces {{range $.DNSNameResolverNamepsaces}}{{.}}{{end}}
```
alebedev87, line 73: Will we have a space separator? Can you please add a unit test with multiple namespaces?

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -1141,6 +1141,78 @@ func Test_coreDNSCache(t *testing.T) {
...
+						{
+							Name:  "buzz",
+							Zones: []string{"buzz.com", "example.buzz.com"},
+							ForwardPlugin: operatorv1.ForwardPlugin{
+								Upstreams: []string{"4.4.4.4"},
+							},
+						},
+					},
+				},
+			},
```
alebedev87, line 1192: Do we really need that many servers to check the plugin is added?

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -196,27 +206,31 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string, caBundleRevi
...
+		UpstreamResolvers:         upstreamResolvers,
+		PolicyStr:                 coreDNSPolicy,
+		LogLevel:                  coreDNSLogLevel(dns),
+		CABundleRevisionMap:       caBundleRevisionMap,
+		CABundleFileName:          caBundleFileName,
+		LameDuckDuration:          lameDuckDuration,
+		PositiveTTL:               pTTL,
+		NegativeTTL:               nTTL,
+		OCPDNSNameResolver:        dnsNameResolverEnabled,
+		DNSNameResolverNamepsaces: []string{DefaultDNSNameResolverNamespace},
```
alebedev87, line 233: Don't we want the namespaces as a part of the operator config along with the bool about enabled or not?

pkg/operator/controller/controller_dns_configmap.go
```diff
@@ -196,27 +206,31 @@ func desiredDNSConfigMap(dns *operatorv1.DNS, clusterDomain string, caBundleRevi
...
+		UpstreamResolvers         operatorv1.UpstreamResolvers
+		PolicyStr                 func(policy operatorv1.ForwardingPolicy) string
+		LogLevel                  string
+		CABundleRevisionMap       map[string]string
+		CABundleFileName          string
+		LameDuckDuration          time.Duration
+		PositiveTTL               uint32
+		NegativeTTL               uint32
+		OCPDNSNameResolver        bool
+		DNSNameResolverNamepsaces []string
```
alebedev87, line 220:
```suggestion
		DNSNameResolverNamespaces []string
```

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -1141,6 +1141,78 @@ func Test_coreDNSCache(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if cm, err := desiredDNSConfigMap(tc.dns, clusterDomain, cmMap, true); err != nil {
+				if !errors.Is(err, tc.expectedError) {
+					t.Errorf("Unexpected error : %v", err)
+				}
+			} else if tc.expectedError != nil {
+				t.Errorf("Error %v was expected", tc.expectedError)
+			} else if diff := cmp.Diff(cm.Data["Corefile"], tc.expectedCoreFile); diff != "" {
+				t.Errorf("unexpected Corefile;\n%s", diff)
```
alebedev87, line 1210:
```suggestion
				t.Errorf("Unexpected Corefile;\n%s", diff)
```
As all the other errors start with an upper case letter.

pkg/operator/controller/names.go
```diff
@@ -140,3 +146,11 @@ func NodeResolverDaemonSetPodSelector() *metav1.LabelSelector {
...
 	}
 }
+
+// FeatureGateClusterConfigName returns the namespaced name of the
+// featuregates.config.openshift.io resource of the cluster.
+func FeatureGateClusterConfigName() types.NamespacedName {
+	return types.NamespacedName{
+		Name: DefaultFeatureGate,
+	}
+}
```
alebedev87, line 156: I cannot see where this function is used.

pkg/operator/controller/controller_dns_configmap_test.go
arkadeepsen: I have added 4 servers to show that the ocp_dnsnameresolver plugin gets added to each server config.

pkg/operator/controller/controller_dns_configmap_test.go
arkadeepsen: Actually in the other tests the error message "Unexpected error :" is in upper case and ""unexpected Corefile;" is in lower case. I had just copy pasted the same. LMK if I should change the case here.

pkg/operator/controller/names.go
arkadeepsen: This is not used in this PR. I'll remove the function.

pkg/operator/operator.go
```diff
@@ -36,6 +42,57 @@ type Operator struct {
 
 // New creates (but does not start) a new operator from configuration.
 func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, error) {
+	ctx := context.Background()
+	ctx, cancelFn := context.WithCancel(ctx)
+	go func() {
+		defer cancelFn()
+		<-config.Stop
```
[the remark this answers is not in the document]
arkadeepsen: I have followed the same structure that is followed in cluster ingress operator: https://github.com/openshift/cluster-ingress-operator/blob/bf974a96704fb0e98dad586ab81b86cb65123145/pkg/operator/operator.go#L80-L130

PLMK if I should go ahead and make the suggested changes, or stick to the similar format that is used in cluster ingress operator.

pkg/operator/operator.go
```diff
@@ -70,7 +127,10 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		KubeRBACProxyImage:     config.KubeRBACProxyImage,
 		OperatorReleaseVersion: config.OperatorReleaseVersion,
 	}
-	if _, err := operatorcontroller.New(operatorManager, cfg); err != nil {
+	if _, err := operatorcontroller.New(operatorManager, operatorcontroller.Config{
+		Config:                 cfg,
+		DNSNameResolverEnabled: dnsNameResolverEnabled,
```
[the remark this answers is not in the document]
arkadeepsen: I am open to both. As `operatorconfig.Config` is being used in status controller as well, I thought adding it to specifically the `operatorcontroller.Config` it reduce the duplication of the information. PLMK your thoughts.

pkg/operator/controller/controller_dns_configmap.go
arkadeepsen: I'll add the namespace as part of the config.

pkg/operator/controller/controller_dns_configmap_test.go
alebedev87: Wouldn't 2 servers be sufficient then?

pkg/operator/operator.go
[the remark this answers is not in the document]
alebedev87: Yeah, the two operators differ in many aspects, making them diverge further doesn't seem to be practical. However the usage of `Stop` channel is still questionable: who closes it? if no routine does it then don't we care about the correct termination of the accesor and informer? if we don't why not passing a simple `context.TODO()`?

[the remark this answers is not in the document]
arkadeepsen: When I used the code from cluster ingress operator I also was wondering about the same thing. We can pass the `Stop` channel to `Start()` function of the operator and close the channel whenever the operator exits. WDYT?

pkg/operator/controller/controller_dns_configmap_test.go
arkadeepsen: Ok. I'll reduce the number of servers to 2.

pkg/operator/controller/controller_dns_configmap_test.go
arkadeepsen: Ok. I'll change it to upper case.

pkg/operator/operator.go
```diff
@@ -36,6 +42,57 @@ type Operator struct {
...
+
+	configClient, err := configclient.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	desiredVersion := config.OperatorReleaseVersion
+	missingVersion := "0.0.1-snapshot"
+
+	// By default, this will exit(0) the process if the featuregates ever change to a different set of values.
```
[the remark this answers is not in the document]
arkadeepsen: In future we might add other feature gates. Probably it's better not to mention the specific feature gate (similar to cluster ingress operator).

pkg/operator/controller/controller_dns_configmap.go
arkadeepsen: Yes. I have added the space separator and the unit test to check it.

arkadeepsen:
/hold cancel 
https://github.com/openshift/coredns-ocp-dnsnameresolver/pull/1 is merged

pkg/operator/operator.go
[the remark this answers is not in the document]
alebedev87: As discussed, the context created by the controller runtime's signal handler is used.

pkg/operator/operator.go
```diff
@@ -36,6 +42,57 @@ type Operator struct {
 
 // New creates (but does not start) a new operator from configuration.
 func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, error) {
+	ctx := context.Background()
+	ctx, cancelFn := context.WithCancel(ctx)
+	go func() {
+		defer cancelFn()
+		<-config.Stop
+	}()
```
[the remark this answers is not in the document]
alebedev87: As discussed, the context is now passed as an argument.

pkg/operator/operator.go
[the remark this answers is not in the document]
alebedev87: As discussed, let's keep this part similar to the cluster-ingress-operator hence the controller config can be kept.

pkg/operator/controller/controller_dns_configmap_test.go
```diff
@@ -1141,6 +1141,76 @@ func Test_coreDNSCache(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if cm, err := desiredDNSConfigMap(tc.dns, clusterDomain, cmMap, true, tc.namespaces); err != nil {
+				if !errors.Is(err, tc.expectedError) {
+					t.Errorf("Unexpected error : %v", err)
+				}
+			} else if tc.expectedError != nil {
+				t.Errorf("Error %v was expected", tc.expectedError)
+			} else if diff := cmp.Diff(cm.Data["Corefile"], tc.expectedCoreFile); diff != "" {
```
alebedev87, line 1207: Would be nice to dry-run the corefile to check the syntax. But this can be an item for the future.

alebedev87:
Thanks for the updates!

/lgtm
/approve

PR #405, opened 2024-02-07
OCPBUGS-20532: Fix weak cipher suites in cluster DNS operator

candita:
OCPBUGS-20532 Outlines an issue with weak cipher suites in cluster DNS, particularly those using CBC, 3DES, or RSA key exchange withou ECDHE.  This commit removes three weak cipher suites and adds seven secure cipher suites used for kube-rbac-proxy configuration.  According to the kube-rbac-proxy  the tls-cipher-suites flag represents a                   "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used"  We don't want to use the [default Go cipher suites ](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/crypto/tls/cipher_suites.go;l=53) because they contain insecure/weak ciphers such as the ones cited in this bug report.

The old list of ciphers (sorted):
TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 - removed (CBC)
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 - ok
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 - removed (CBC)
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 - ok
TLS_RSA_WITH_AES_128_CBC_SHA256 - removed (CBC, RSA without ECDHE)

The new list of ciphers was checked against the tool at https://ciphersuite.info and found to be secure or recommended.  The new list is:
TLS_AES_128_GCM_SHA256,
TLS_AES_256_GCM_SHA384,
TLS_CHACHA20_POLY1305_SHA256,
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, (pre-existing)
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, (pre-existing)
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256


candita:
"openshift cluster install failed with cluster creation"

/retest

candita:
This solution isn't acceptable.
/close

PR #409, opened 2024-05-06
WIP - Updating ose-cluster-dns-operator-container image to be consistent wi…

lihongan:
Updating ose-cluster-dns-operator-container image to be consistent with ART for 4.16

test: try to replace https://github.com/openshift/cluster-dns-operator/pull/401 ( job ci/prow/verify always failed)


candita:
I believe this is no longer needed.
/close

PR #412, opened 2024-05-17
[WIP] OCPBUGS-33914: Bump openshift/api for features in cluster-dns-operator

candita:
Pick up changes to Gateway API feature gate so that it isn't tested in tech preview CI tests.

 go get github.com/openshift/api@1dea0237181fc64493521c5e1b86ca4209019c75
 go mod vendor
 go mod tidy


gcs278:
@candita I don't see any references to Gateway API in cluster-dns-operator, nor use of the `FeatureGateGatewayAPI` feature gate. All of Gateway API DevPreview implementation was in cluster-ingress-operator, even the DNS logic. Am I misunderstanding something?

candita:
> @candita I don't see any references to Gateway API in cluster-dns-operator, nor use of the `FeatureGateGatewayAPI` feature gate. All of Gateway API DevPreview implementation was in cluster-ingress-operator, even the DNS logic. Am I misunderstanding something?

@grant as I mentioned in the bug description, when the e2e-aws-ovn-techpreview test runs, you can see the GatewayAPI feature enabled in the logs.  Even though cluster-dns-operator doesn't run any specific tests for GatewayAPI at the moment, it would be best to correct this just in case we need to work with cluster-dns-operator in the future.

candita:
Not needed.
/close

PR #437, opened 2025-06-19
Implements https://issues.redhat.com/browse/OCPSTRAT-2006

rgordill:
This commit adds a new ipv6Filter parameter to the DNS Custom Resource Definition (CRD), defaulting to false. When enabled (true), it returns a NOERROR response with an empty answer section, effectively suppressing IPv6 query propagation to upstream DNS servers when desired.

Use with caution, as it may introduce unintended side effects.


khalid-lemghari:
Hi team,

If a workload is IPv6-only or expects IPv6 first, resolution will fail, so the proposition is bad practice. I recommend to use `rewrite` instead of `template`:

```sh
rewrite stop type AAAA A
```

PR #355, opened 2023-01-31
OCPBUGS-4359: update-node-resolver.sh: Check for errors from >>

Miciah:
Check for errors when appending entries to `/etc/hosts.tmp` in the `update-node-resolver.sh` script.

Follow-up to #352.

* `assets/node-resolver/update-node-resolver.sh`: Check for failures when appending entries to `/etc/hosts.tmp`.
* `pkg/manifests/bindata.go`: Regenerate.  

---

FYI @bverschueren.
/assign @frobware

PR #439, opened 2025-07-02
OCPBUGS-59781: Read only root filesystem

jacobsee:
Switch operator & managed daemonset to use readOnlyRootFilesystem: true

### DIFF
--- a/manifests/0000_70_dns-operator_02-deployment.yaml
+++ b/manifests/0000_70_dns-operator_02-deployment.yaml
@@ -54,6 +54,10 @@ spec:
           capabilities:
             drop:
             - ALL
+          readOnlyRootFilesystem: true
+        volumeMounts:
+        - mountPath: /tmp
+          name: tmp-dir
       - name: kube-rbac-proxy
         image: quay.io/openshift/origin-kube-rbac-proxy:latest
         args:
@@ -75,15 +79,20 @@ spec:
           capabilities:
             drop:
             - ALL
+          readOnlyRootFilesystem: true
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
         - mountPath: /etc/tls/private
           name: metrics-tls
           readOnly: true
+        - mountPath: /tmp
+          name: tmp-dir
       volumes:
       - name: metrics-tls
         secret:
           secretName: metrics-tls
+      - name: tmp-dir
+        emptyDir: {}
       terminationGracePeriodSeconds: 2
       tolerations:
       - key: "node-role.kubernetes.io/master"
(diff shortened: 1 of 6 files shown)


jacobsee:
Cleaned up commits now that required tests have passed.

frobware:
@dusk125 I'n not working in this area so unassigning myself.

/unassign

manifests/0000_70_dns-operator_02-deployment.yaml
```diff
@@ -54,6 +54,10 @@ spec:
           capabilities:
             drop:
             - ALL
+          readOnlyRootFilesystem: true
+        volumeMounts:
+        - mountPath: /tmp
+          name: tmp-dir
```
Miciah, line 60: Does the dns-operator container really need a writable `/tmp`, or is this change the result of applying a generic change to add a writable `/tmp` in every container?

manifests/0000_70_dns-operator_02-deployment.yaml
```diff
@@ -75,15 +79,20 @@ spec:
...
             drop:
             - ALL
+          readOnlyRootFilesystem: true
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
         - mountPath: /etc/tls/private
           name: metrics-tls
           readOnly: true
+        - mountPath: /tmp
+          name: tmp-dir
```
Miciah, line 89: Does the kube-rbac-proxy container really need a writable `/tmp`?  Is it intentional that the dns-operator container and the kube-rbac-proxy container share a `/tmp` volume?

Miciah (commented):
Thanks for the PR!  

Usually a PR is linked to either an OCPBUGS ticket or a story belonging to some epic in Jira, rather than linking directly to an RFE ticket.  Is there a reason not to have an epic and story in this case?

Please also reference the Jira ticket in the commit message.  

---

It is interesting to note that these changes probably would have resolved [OCPBUGS-4359](https://issues.redhat.com/browse/OCPBUGS-4359).  In that issue, we were missing some error checks and writing an empty `/etc/hosts.tmp` file when the host volume with `/etc/hosts` was full.  To solve that issue, #352 and #355 added checks to fail immediately when writing the `/etc/hosts.tmp` file failed.  Using `/tmp/hosts.tmp` with an emptydir volume would avoid the full-volume situation entirely.  However, it is still good to have the error checks that we added before as these checks could catch other errors besides having a full volume.

manifests/0000_70_dns-operator_02-deployment.yaml
dusk125: It's the latter, we've been adding the tmp mounts even when tests pass since somethings silently need to write to /tmp for various reasons.

manifests/0000_70_dns-operator_02-deployment.yaml
dusk125: Similar to above, more for a "just in case" rather than a specific need. And it's not needed that they share the /tmp dir, more just to keep from adding more emptyDirs than necessary. If this is of concern, we can remove it.

Miciah:
Thanks!

/approve

PR #441, opened 2025-08-01
RFE-4145: Support custom list of services to be added to /etc/hosts by Node resolver

t-cas:
Today, node-resolver daemonset in openshift-dns-operator namespace is referencing a SERVICES environment variable allowing to update /etc/hosts of all nodes of the cluster with entries for custom services of the platform, so it can be targeted directly from nodes. This allow the nodes to have access to internal openshift registry directly using DNS entry image-registry.openshift-image-registry.svc
Only issue is that this value for _SERVICES_ env variable is harcoded in cluster-dns-operator code, so it does not allow to reference other services available also at node level.

This PR update DNS operator.openshift.io custom resource and associated controller to support a list of openshift services so they will be added automatically in /etc/hosts and available on all cluster nodes.

This is tracked by [RFE-4145](https://issues.redhat.com/browse/RFE-4145)

### DIFF
--- a/pkg/operator/controller/controller_dns_node_resolver_daemonset.go
+++ b/pkg/operator/controller/controller_dns_node_resolver_daemonset.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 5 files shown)


t-cas:
@candita thanks for your feedback, api change done in this PR https://github.com/openshift/api/pull/2435 as requested

The change is rather small, and it is here already implemented and tested, how can we move forward with RFE ?

pkg/operator/controller/controller_dns_node_resolver_daemonset.go
```diff
@@ -38,6 +37,27 @@ var (
...
 
+// buildServicesList combines the default service with additional services from the DNS spec.
+// The default service is always included first, followed by any additional services.
+func buildServicesList(dns *operatorv1.DNS) string {
+	// Start with the default service
+	services := []string{defaultService}
+
+	// Add any additional services from the spec
+	if len(dns.Spec.NodeServices) > 0 {
+		for _, service := range dns.Spec.NodeServices {
```
sdodson, line 48: @candita Will this ensure that the list, assuming items aren't changing, are in a stable order or do we need to sort them to prevent churn?

candita:
@t-cas @sdodson just an update - decision needs to wait until our team lead is back from PTO next week.  However, just to set expectations, we do have higher priority items for review.

PR #443, opened 2025-08-08
NE-1476 - Added network policies for DNS

knobunc:
Added the framework for network policies for DNS for the operator and the dns pods.

The operator has a deny all network policy that for the openshift-dns-operator namespace and an allow policy for egress to the apiserver and dns ports at any IP.

The operator installs a deny all network policy for the openshift-dns namespace.

Then for each dns that it manages it installs an allow policy for ingress for dns traffic and metrics.

It has to allow ingress from the dns pods to any IP because we allow configuration to set the upstream server and port, so any valid IP and port needs to be allowed.

It also needs access to the api server, but that is covered by the wildcard allow policy.

https://issues.redhat.com/browse/NE-1476


pkg/manifests/assets/dns/networkpolicy-allow.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - protocol: TCP
+      port: 9154
+    - protocol: UDP
+      port: 5353
+    - protocol: TCP
+      port: 5353
+    - protocol: TCP
+      port: 8080
+    - protocol: TCP
+      port: 8181
```
rikatz, line 22: IIRC the health check ports don't need to be open, as kubelet should be able to reach the Pods with or without the network policy (unless ovn-kubernetes does differently). 

It is worth testing tho

candita:
> === RUN   TestDNSForwarding
>     operator_test.go:694: failed to dig 172.30.224.202: failed to find "1.2.3.4"
> 

/retest required

pkg/manifests/assets/dns/networkpolicy-allow.yaml
knobunc: I was getting test errors (which surprised me) until these were opened.  I wasn't sure if something else was hitting them to determine when it was live.

pkg/operator/controller/controller_dns_networkpolicy.go
```diff
@@ -0,0 +1,121 @@
...
+// ensureDNSNetworkPolicy ensures that a NetworkPolicy exists for the given DNS.
+func (r *reconciler) ensureDNSNetworkPolicy(dns *operatorv1.DNS) (bool, *networkingv1.NetworkPolicy, error) {
+	haveNP, current, err := r.currentDNSNetworkPolicy(dns)
+	if err != nil {
+		return false, nil, err
+	}
+
+	desired := desiredDNSNetworkPolicy(dns)
+
+	switch {
```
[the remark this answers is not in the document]
knobunc: I did it that way to match the pattern in other files (e.g. controller_dns_configmap.go

pkg/manifests/assets/dns/networkpolicy-allow.yaml
rikatz: hum, weird.

I have just applied the network policy manually, so the port 8080 was blocked from a pod on another namespace but I could still curl it from kubelet:

```
 oc debug node/ip-10-0-86-223.us-east-2.compute.internal -- curl -kv http://10.131.0.5:8080/health
....
OK
```

then trying from my pod:
```
kubectl exec -it nginx-5869d7778c-pl7fq -- curl -m 3 http://10.131.0.5:8080/health
curl: (28) Connection timed out after 3000 milliseconds
```

pkg/manifests/assets/dns/networkpolicy-allow.yaml
```diff
@@ -0,0 +1,43 @@
...
+  # Allow the dns operator namespaces to hit the healthcheck ports
+  - ports:
+    - protocol: TCP
+      port: 8080
+    - protocol: TCP
+      port: 8181
+    from:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns-operatorXXX
```
rikatz, line 35: Is this something from a previous test that was left here?

manifests/0000_70_dns-operator_01-network-policy.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - protocol: TCP
+      port: 6443
+    - protocol: TCP
+      port: 53
+    - protocol: UDP
+      port: 53
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 9393
```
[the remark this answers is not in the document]
rikatz: this can be marked as solved

PR #445, opened 2025-08-27
OCPBUGS-59781: Read only root filesystem

jacobsee:
Missed the kube-rbac-proxy container in #439


jacobsee:
Hi @Miciah - this proxy container was missed as a part of #439, any chance you could give this one a look to wrap that up? Thanks.

PR #474, opened 2026-05-05
OCPBUGS-55179: Set internalTrafficPolicy to Local on dns-default service

Ultimate-etamitlU:
## Summary

DNS queries from pods on primary user-defined networks (UDNs) scatter randomly across dns-default pods on all nodes instead of being handled by the local node's dns-default pod.

UDN pods reach the dns-default service (172.30.0.10) through a path that crosses network boundaries: UDN pod → UDN cluster router → management port → default network → OVN load balancer → dns-default pod. The OVN load balancer on the default network treats dns-default as a standard ClusterIP service and distributes traffic across all backend pods cluster-wide.

Setting `internalTrafficPolicy: Local` restricts the EndpointSlice to contain only the node-local dns-default backend. Since dns-default runs as a DaemonSet with a pod on every node, this is safe and guarantees that DNS queries are always handled by the local pod.

The existing `trafficDistribution: PreferSameNode` (added in PR #457) provides a soft hint for same-node preference but does not guarantee locality for cross-network UDN traffic. `internalTrafficPolicy: Local` provides the hard constraint needed.

## Why DNS operator, not OVN?

The DNS query scatter from primary UDN pods is not caused by an OVN routing bug. In OCP 4.14+ (interconnect mode), each node has its own zone and its own UDN cluster router with a single route — no ECMP happens at the UDN router level. The scatter occurs downstream: once UDN traffic crosses into the default network via the management port, the OVN load balancer distributes dns-default traffic across all backend pods because the service has `internalTrafficPolicy: Cluster` (the default). The fix belongs in the DNS operator, which owns the dns-default service spec.

## Verification

- `make test` — passed
- `make verify` — passed
- On-cluster validation on OCP 4.20 (UPI, 5 nodes) and OCP 4.21 (IPI, 5 nodes):
  - Baseline (ITP=Cluster): DNS queries from UDN pods confirmed scattering across multiple nodes
  - Fix (ITP=Local): OVN LB on each node confirmed to have only the local dns-default pod as backend
  - DNS resolution from UDN pods works correctly with ITP=Local

Jira: https://issues.redhat.com/browse/OCPBUGS-55179

## Summary by CodeRabbit

* **Chores**
  * DNS service updated to prefer routing pod-to-pod traffic on the local node (internal traffic policy set to Local), reducing cross-node network hops.
* **Tests**
  * Added unit coverage to verify the DNS service internal traffic policy is set to Local.

### DIFF
--- a/pkg/manifests/assets/dns/service.yaml
+++ b/pkg/manifests/assets/dns/service.yaml
@@ -4,6 +4,7 @@ apiVersion: v1
 spec:
   # clusterIP will be automatically managed.
   # selector is set at runtime
+  internalTrafficPolicy: Local
   trafficDistribution: PreferSameNode
   ports:
   - name: dns
--- a/pkg/operator/controller/controller_dns_service_test.go
+++ b/pkg/operator/controller/controller_dns_service_test.go
@@ -204,8 +204,7 @@ func TestDesiredDNSServiceInternalTrafficPolicy(t *testing.T) {
 	svc := desiredDNSService(dns, "172.30.0.10", false, daemonsetRef)
 
 	assert.NotNil(t, svc.Spec.InternalTrafficPolicy)
-	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy,
-		"dns-default service must use InternalTrafficPolicy Local to ensure node-local DNS resolution")
+	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy)
 }
 
 func Test_shouldEnableTopologyAwareHints(t *testing.T) {
--- a/pkg/operator/controller/controller_dns_service_test.go
+++ b/pkg/operator/controller/controller_dns_service_test.go
@@ -196,6 +196,20 @@ func TestDNSServiceChanged(t *testing.T) {
 	}
 }
 
+func TestDesiredDNSServiceInternalTrafficPolicy(t *testing.T) {
+	dns := &operatorv1.DNS{
+		ObjectMeta: metav1.ObjectMeta{Name: "default"},
+	}
+	daemonsetRef := metav1.OwnerReference{}
+	svc := desiredDNSService(dns, "172.30.0.10", false, daemonsetRef)
+
+	if !assert.NotNil(t, svc.Spec.InternalTrafficPolicy, "expected InternalTrafficPolicy to be set") {
+		return
+	}
+	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy,
+		"dns-default service must use InternalTrafficPolicy Local to ensure node-local DNS resolution")
+}
+
 func Test_shouldEnableTopologyAwareHints(t *testing.T) {
 	emptyLabels := map[string]string{}
 	someCPU := map[corev1.ResourceName]resource.Quantity{
--- a/pkg/operator/controller/controller_dns_service_test.go
+++ b/pkg/operator/controller/controller_dns_service_test.go
@@ -203,9 +203,7 @@ func TestDesiredDNSServiceInternalTrafficPolicy(t *testing.T) {
 	daemonsetRef := metav1.OwnerReference{}
 	svc := desiredDNSService(dns, "172.30.0.10", false, daemonsetRef)
 
-	if !assert.NotNil(t, svc.Spec.InternalTrafficPolicy, "expected InternalTrafficPolicy to be set") {
-		return
-	}
+	assert.NotNil(t, svc.Spec.InternalTrafficPolicy)
 	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy,
 		"dns-default service must use InternalTrafficPolicy Local to ensure node-local DNS resolution")
 }


Miciah:
> Setting `internalTrafficPolicy: Local` restricts the EndpointSlice to contain only the node-local dns-default backend. Since dns-default runs as a DaemonSet with a pod on every node, this is safe and guarantees that DNS queries are always handled by the local pod.

This is a false assumption.  There are situations in which the dns-default pods do not necessarily run on every node, for example if a node is tainted, if `dnses.operator.openshift.io/cluster` specifies `spec.nodePlacement` with a custom selector, or if a node is shutting down during a cluster upgrade.  We really need the behavior of prefer-local *with fallback*.  That's why we set `trafficDistribution: PreferSameNode` and why ovn-kubernetes has special prefer-local-with-fallback behavior for the dns-default pods: https://github.com/openshift/ovn-kubernetes/blob/952886fd8af2ca3ecf1717a2cb69311a32f25c06/go-controller/pkg/ovn/controller/services/lb_config.go#L258-L264

pkg/manifests/assets/dns/service.yaml
```diff
@@ -4,6 +4,7 @@ apiVersion: v1
 spec:
   # clusterIP will be automatically managed.
   # selector is set at runtime
+  internalTrafficPolicy: Local
```
Miciah, line 7: You would also need to update `daemonsetConfigChanged` in `pkg/operator/controller/controller_dns_daemonset.go` to ensure existing an daemon set got updated.

pkg/operator/controller/controller_dns_service_test.go
```diff
@@ -196,6 +196,20 @@ func TestDNSServiceChanged(t *testing.T) {
...
+func TestDesiredDNSServiceInternalTrafficPolicy(t *testing.T) {
+	dns := &operatorv1.DNS{
+		ObjectMeta: metav1.ObjectMeta{Name: "default"},
+	}
+	daemonsetRef := metav1.OwnerReference{}
+	svc := desiredDNSService(dns, "172.30.0.10", false, daemonsetRef)
+
+	if !assert.NotNil(t, svc.Spec.InternalTrafficPolicy, "expected InternalTrafficPolicy to be set") {
+		return
+	}
```
Miciah: You generally don't need `if` with `assert` (using `if` kind of defeats the purpose of using testify to improve concision):
```suggestion
	assert.NotNil(t, svc.Spec.InternalTrafficPolicy)
```

pkg/operator/controller/controller_dns_service_test.go
```diff
@@ -196,6 +196,20 @@ func TestDNSServiceChanged(t *testing.T) {
...
+	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy,
+		"dns-default service must use InternalTrafficPolicy Local to ensure node-local DNS resolution")
```
Miciah: Is the message helpful, or can we simplify this assertion?  
```suggestion
	assert.Equal(t, corev1.ServiceInternalTrafficPolicyLocal, *svc.Spec.InternalTrafficPolicy)
```

Miciah: Is there a reason to define a new test rather than putting the assertions in `TestDesiredDNSDaemonset`?

Ultimate-etamitlU:
> > Setting `internalTrafficPolicy: Local` restricts the EndpointSlice to contain only the node-local dns-default backend. Since dns-default runs as a DaemonSet with a pod on every node, this is safe and guarantees that DNS queries are always handled by the local pod.
> 
> This is a false assumption. There are situations in which the dns-default pods do not necessarily run on every node, for example if a node is tainted, if `dnses.operator.openshift.io/cluster` specifies `spec.nodePlacement` with a custom selector, or if a node is shutting down during a cluster upgrade. We really need the behavior of prefer-local _with fallback_. That's why we set `trafficDistribution: PreferSameNode` and why ovn-kubernetes has special prefer-local-with-fallback behavior for the dns-default pods: https://github.com/openshift/ovn-kubernetes/blob/952886fd8af2ca3ecf1717a2cb69311a32f25c06/go-controller/pkg/ovn/controller/services/lb_config.go#L258-L264

| Thank you for suggestions, I agree - with-fallback these tainting/upgrade or even node replacement would turn to issues.  Initially I thought of OVN-K based check/fix but couldn't really conclude to that. Now even as I see "trafficDistribution: PreferSameNode" is much needed I feel I shouldn't  looking at DNS operator, rather would like to hear your suggestion.  Rest, this is around a UDN traffic issue - may be I can try learning something more based on your suggestions. 

+ I read that 457 PR and considered OVN-K would add support for preferLocal, the ask is time being.  So may be this isn't the moment I shall be looking for the change here :/

pkg/operator/controller/controller_dns_service_test.go
Ultimate-etamitlU: Didn't se any existing service, likely missed on that ,  so rather adding thought of creating new one. Can try this, though as you have shared the suggestions above I feel it would need more hands than mine.

pkg/manifests/assets/dns/service.yaml
Ultimate-etamitlU: Bit confused here, I assume this shall help me keeping iTP to local.

Ultimate-etamitlU:
Closing based on reviewer feedback from @Miciah.

The assumption that dns-default runs on every node is incorrect — nodes can be tainted, `spec.nodePlacement` can restrict the DaemonSet, or a node may be draining during upgrade. Setting `internalTrafficPolicy: Local` without fallback would cause a full DNS blackhole in those scenarios, which is worse than the original scatter problem.

The correct fix is **prefer-local with fallback** at the OVN-K layer, which already has hooks for dns-default preferLocal behavior in `lb_config.go`. This fix belongs in ovn-kubernetes, not the DNS operator.

Reopening tracking under the ovn-kubernetes component. `trafficDistribution: PreferSameNode` (PR #457) remains as the soft hint in the interim.

PR #477, opened 2026-05-28
OCPBUGS-86009: Fix dns operator reporting Progressing=True on scale up

bentito:
The `cluster-dns-operator` was incorrectly reporting `Progressing=True` during normal cluster scale-up events. This was caused by `computeDNSProgressingCondition` checking `UpdatedNumberScheduled` against `DesiredNumberScheduled`. When a new node is added, `DesiredNumberScheduled` increases immediately, but `UpdatedNumberScheduled` lags until the pod is scheduled/running, causing the operator to appear to be rolling out.

This PR introduces an `isDaemonSetRollingOut` helper, modeled after similar logic in the `cluster-ingress-operator` (see [PR #1299](https://github.com/openshift/cluster-ingress-operator/pull/1299)). It checks if the DaemonSet `Generation` matches `ObservedGeneration`, and checks if `UpdatedNumberScheduled` is less than `CurrentNumberScheduled`. If not, the operator is simply scaling and should not report `Progressing=True`.

Tested locally by ensuring the test suite correctly accommodates the new structure and verifies progression states.

Bug: [OCPBUGS-86009](https://redhat.atlassian.net/browse/OCPBUGS-86009)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS component status reporting during DaemonSet updates. “Have…want…” progressing messages are now shown only when the DNS and node-resolver DaemonSets are actively rolling out, and are suppressed when counts don’t match but a rollout isn’t in progress.

### DIFF
--- a/pkg/operator/controller/dns_status.go
+++ b/pkg/operator/controller/dns_status.go
@@ -208,7 +208,9 @@ func computeDNSProgressingCondition(oldCondition *operatorv1.OperatorCondition, 
 		have := dnsDaemonset.Status.UpdatedNumberScheduled // num of nodes running the updated pod.
 		// It's progressing when have < want.  If have >= want, that's okay.
 		if have < want {
-			messages = append(messages, fmt.Sprintf("Have %d up-to-date DNS pods, want %d.", have, want))
+			if isDaemonSetRollingOut(dnsDaemonset) {
+				messages = append(messages, fmt.Sprintf("Have %d up-to-date DNS pods, want %d.", have, want))
+			}
 		}
 
 		haveSelector := dnsDaemonset.Spec.Template.Spec.NodeSelector
@@ -231,7 +233,9 @@ func computeDNSProgressingCondition(oldCondition *operatorv1.OperatorCondition, 
 
 		// It's progressing when have < want.  If have >= want, that's okay.
 		if have < want {
-			messages = append(messages, fmt.Sprintf("Have %d available node-resolver pods, want %d.", have, want))
+			if isDaemonSetRollingOut(nodeResolverDaemonset) {
+				messages = append(messages, fmt.Sprintf("Have %d available node-resolver pods, want %d.", have, want))
+			}
 		}
 	}
 	if len(messages) != 0 {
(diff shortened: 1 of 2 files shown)


bentito:
This PR is taking a similar approach as https://github.com/openshift/cluster-ingress-operator/pull/1299 did for same problems on CIO

davidesalerno (commented):
The changes lgmt evenif I think we still need a dedicated regression test for the exact scenario this PR is fixing: scale-up without rollout.

Right now these fixture updates set `CurrentNumberScheduled` equal to `DesiredNumberScheduled`, so the tests don’t exercise the case where:

- `DesiredNumberScheduled > CurrentNumberScheduled`
- `UpdatedNumberScheduled == CurrentNumberScheduled`
- `Generation == ObservedGeneration`

In that state, the DaemonSet is scaling up but not rolling out, and `computeDNSProgressingCondition` should remain P`rogressing=False`.

Could we add a table-driven case that covers that combination explicitly? That would make the regression protection much stronger for this bug.

bentito:
> The changes lgmt evenif I think we still need a dedicated regression test for the exact scenario this PR is fixing: scale-up without rollout.
> 
> Right now these fixture updates set `CurrentNumberScheduled` equal to `DesiredNumberScheduled`, so the tests don’t exercise the case where:
> 
>     * `DesiredNumberScheduled > CurrentNumberScheduled`
> 
>     * `UpdatedNumberScheduled == CurrentNumberScheduled`
> 
>     * `Generation == ObservedGeneration`
> 
> 
> In that state, the DaemonSet is scaling up but not rolling out, and `computeDNSProgressingCondition` should remain P`rogressing=False`.
> 
> Could we add a table-driven case that covers that combination explicitly? That would make the regression protection much stronger for this bug.

Thanks @davidesalerno  sorry about delay. I've added unit tests to better lock this in.

davidesalerno:
/payload-aggregate periodic-ci-openshift-release-master-nightly-5.1-e2e-aws-ovn-serial 3

davidesalerno:
/payload-aggregate periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-serial 3

melvinjoseph86:
Tested using clusterbot.
```

Verified Progressing stays False during node scale-up and  also
Verified Progressing=True still fires during a real rollout

➜  Downloads oc get co/dns -w
NAME   VERSION                                               AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
dns    5.0.0-0-2026-07-29-074554-test-ci-ln-770qy6t-latest   True        False         False      82m     
```

Hence marking this as verified 
/verified by @mjoseph

PR #482, opened 2026-07-09
NE-2743: Support centralized TLS security profiles for CDO metrics

Thealisyed:
## Summary

Removes kube-rbac-proxy from the **operator** deployment and replaces it with controller-runtime native TLS serving via `filters.WithAuthenticationAndAuthorization`. This is **Step 1** of the two-step kube-rbac-proxy removal agreed by the team.

- Vendor `openshift/api` to pick up `TLSProfileSpec.Groups` (`TLSGroupPreferences`) and bump `k8s.io` dependencies to v0.35.1
- Remove kube-rbac-proxy sidecar from the operator deployment (port 9393), serve metrics directly with TLS + auth
- Read cluster-wide APIServer TLS security profile at startup for cipher suites, min TLS version, and curve preferences
- Add `TLSConfigFromProfile` / `TLSGroupToCurveID` helpers with warning logs for unsupported groups and ciphers
- Consolidate DNSNameResolver CRD from 3 feature-gated variants to 1 ungated CRD

### What's deliberately kept

`KubeRBACProxyImage`, `kubeRBACProxyArgs()`, and the `KUBE_RBAC_PROXY_IMAGE` env var remain because the **CoreDNS DaemonSet** still uses kube-rbac-proxy in this release.

### Follow-up

**[NE-2814](https://redhat.atlassian.net/browse/NE-2814)** — Step 2: Remove kube-rbac-proxy from the CoreDNS operand using exporter-toolkit mTLS (target: 5.1). Blocked on upstream [prometheus/exporter-toolkit#417](https://github.com/prometheus/exporter-toolkit/pull/417) and [openshift/coredns#195](https://github.com/openshift/coredns/pull/195).

### Jira

- **Story**: [NE-2743](https://redhat.atlassian.net/browse/NE-2743)
- **Epic**: [NE-2334](https://redhat.atlassian.net/browse/NE-2334)

## Summary by CodeRabbit

* **Security**
  * Updated the DNS operator proxy to use modern TLS cipher suites.
  * Enforced a minimum TLS version of 1.2.

* **Maintenance**
  * Updated Go toolchain version.
  * Updated Kubernetes/OpenShift dependencies to newer patch releases.
  * Updated DNSNameResolver CRD handling: removed the TechPreviewNoUpgrade API variant manifest and broadened the CRD feature-set annotation to include DevPreviewNoUpgrade and TechPreviewNoUpgrade.

* **Documentation**
  * Expanded CRD documentation for toleration operators to include numeric comparisons (`Lt`, `Gt`), feature-gated.


Thealisyed:
@coderabbitai can you review my PR please?

Thealisyed:
where is your feedback? @coderabbitai

Thealisyed:
1. `tls.go` doesn't exist: The function lives in `controller_dns_daemonset.go` and already falls back to `Intermediate` when `Custom` is nil.
  2. `MinTLSVersion` fallback: Not needed because `tlsProfileSpecForProfile` already guarantees a valid spec with `MinTLSVersion` populated before `kubeRBACProxyArgs` ever runs.
  3. CVE-2026-29181: valid CVE but it's an indirect dep as we don't use baggage propagation and out of scope for this PR.
  4. APIServer watch predicate: This is standard OCP operator practice as reconciliation is a no-op when the `DaemonSet` hasn't changed, so filtering adds complexity for no real gain.

Does this make sense to you @coderabbitai (in response to your latest review commet)

candita:
@Thealisyed please do your initial reviews from the command line instead of within github.  Also, this PR is incomplete, so mark it WIP or draft.

pkg/operator/controller/tls.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid TLS version %q: %w", spec.MinTLSVersion, err)
+		}
+		cfg.MinVersion = v
+	}
+
+	if len(spec.Groups) > 0 {
+		var curves []tls.CurveID
+		for _, g := range spec.Groups {
+			if id, ok := TLSGroupToCurveID(g); ok {
```
candita, line 64: The apiserver is the authority here, and we may need to accept curves that we haven't planned to see (such as in a custom profile), so this isn't going to work.

bentito:
Hi @Thealisyed so more of a top level comment here:
The approach here of plumbing the TLS profiles through to the kube-rbac-proxy sidecar is a lot of work for a component that is actually being phased out.

Upstream Kubernetes (sig-auth) has deprecated the gcr.io/kubebuilder/kube-rbac-proxy image due to security concerns, and it's no longer being published. Across OpenShift, there is a major ongoing migration (tracked under [OCPSTRAT-1835](https://redhat.atlassian.net/browse/OCPSTRAT-1835)) to completely remove kube-rbac-proxy sidecars from all components.

Instead of investing effort into dynamically configuring the deprecated proxy, we should lift out kube-rbac-proxy entirely. Since cluster-dns-operator is a controller-runtime based operator, we can replace the sidecar with native supporting code from controller-runtime.

Starting in v0.16.0 (and we are on v0.23.1), controller-runtime provides `filters.WithAuthenticationAndAuthorization`, which natively embeds authn/authz into the operator's metrics server using TokenReview and SubjectAccessReview.

It will be a bigger change but I think we should pivot this PR (or open a new one) to implement this migration for the operator.

### Implementation Steps
  
To adopt this pattern for the operator's metrics:

- Update main.go: Configure the metrics server to use the filter.
```go
import "sigs.k8s.io/controller-runtime/pkg/metrics/filters"
// ...
metricsOpts := metricsserver.Options{
  BindAddress:   ":8443", // Or your preferred port
  SecureServing: true,
  FilterProvider: filters.WithAuthenticationAndAuthorization,
}
```
- Remove the Sidecar: Drop the kube-rbac-proxy container from the dns-operator Deployment manifest (manifests/0000_70_dns-operator_02-deployment.yaml).
- Update RBAC: Ensure the metrics endpoint has the proper RBAC so the OpenShift Prometheus service account can scrape it via TokenReview/SubjectAccessReview.

### Examples
Several operators have already completed this migration. Here are a couple of excellent references you can use as a blueprint:
- kubernetes-nmstate: Upstream PR #1426 (https://github.com/nmstate/kubernetes-nmstate/pull/1426) / Downstream PR #718 (https://github.com/openshift/kubernetes-nmstate/pull/718)
- Operator Framework (operator-controller): PR #216 (https://github.com/openshift/operator-framework-operator-controller/pull/216)

hack/verify-generated-crd.sh
candita, line 1: Please provide an update on why the generator script was changed.

pkg/operator/config/config.go
```diff
@@ -15,7 +17,21 @@ type Config struct {
 	// OpenshiftCLIImage is the openshift client image to manage.
 	OpenshiftCLIImage string
 
-	// KubeRBACProxyImage is the kube-rbac-proxy image to to use
-	// to secure the metrics endpoint.
+	// KubeRBACProxyImage is the kube-rbac-proxy image used by the
+	// CoreDNS DaemonSet operand to secure its metrics endpoint.
 	KubeRBACProxyImage string
```
candita, line 22: The design goal is to replace the kube-rbac-proxy.  Why is this still here?

pkg/operator/controller/controller_dns_daemonset.go
```diff
@@ -196,28 +196,11 @@ func kubeRBACProxyArgs(tlsSecurityProfile *configv1.TLSSecurityProfile) []string
 		"--tls-private-key-file=/etc/tls/private/tls.key",
 		fmt.Sprintf("--tls-cipher-suites=%s", strings.Join(ianaCiphers, ",")),
 		fmt.Sprintf("--tls-min-version=%s", profileSpec.MinTLSVersion),
+		// TODO(NE-2743): Pass --tls-curve-preferences once kube-rbac-proxy
+		// supports it (kube-rbac-proxy#414), using profileSpec.Groups.
```
candita, line 200: Why is this still here?

hack/verify-generated-crd.sh
Thealisyed: The openshift/api v0.35.1 vendor bump consolidated the `DNSNameResolver` CRD, it went from three separate per-feature-gate manifests (TPNoUpgrade, CustomNoUpgrade, DevPreviewNoUpgrade) down to a single ungated CRD. So the script change just follows suit, we verify one file instead of three now.

pkg/operator/config/config.go
Thealisyed: This PR removes `kube-rbac-proxy` from the operator's own deployment (port 9393) but `KubeRBACProxyImage` is still referenced because the CoreDNS `DaemonSet` (port 9154) still uses it. 

Removing it from the operand is a separate step that's blocked on the CoreDNS rebase to v1.14.4 (PR #195), which brings the exporter-toolkit TLS support we need.

pkg/operator/controller/controller_dns_daemonset.go
Thealisyed: Same story as the image reference, `kubeRBACProxyArgs()` and that `TODO` are both for the CoreDNS DaemonSet's `kube-rbac-proxy`, which we're intentionally keeping in this PR. 
The whole function goes away in the follow-up.

pkg/operator/controller/tls.go
Thealisyed: Unfortunately Go doesn't give us a way to convert a curve name string to a `tls.CurveID`. There's an [open proposal](https://github.com/golang/go/issues/77712) for `tls.SupportedCurves()` but it hasn't landed so IIUC every OpenShift operator doing this mapping hits the same problem.

That said, the common case already works well here: when the profile doesn't specify groups, `CurvePreferences` stays nil and Go 1.24+ automatically includes PQC curves. The map only comes into play for Custom profiles that explicitly list groups. I'll add the two missing entries and put in a log warning for any unknown group so we're not silently dropping anything.

bentito:
@Thealisyed I think you can remove WIP label and remove Draft status now, no?

pkg/operator/controller/tls.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"crypto/tls"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/sirupsen/logrus"
+)
+
+var tlsGroupToCurveID = map[configv1.TLSGroup]tls.CurveID{
```
davidesalerno, line 13: I'll put a TODO here so that we can replace this struct and the function below with the library-go equivalent (CurveIDsForTLSGroups) when this https://github.com/openshift/library-go/pull/2347/ will be merged

I'll align the names and the function implementation to that so that it will be easier the replacement

pkg/operator/controller/tls.go
```diff
@@ -0,0 +1,105 @@
...
+				return &profile.Custom.TLSProfileSpec
+			}
+			logrus.Warningf("TLSProfileSpecForSecurityProfile: custom TLS profile has nil spec, using Intermediate default")
+		}
+	}
+	return configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
+}
+
+// copyTLSSpec creates a defensive copy of the given TLSProfileSpec.
+func copyTLSSpec(in *configv1.TLSProfileSpec) *configv1.TLSProfileSpec {
```
davidesalerno: Is this function used somewhere?

pkg/operator/controller/tls.go
```diff
@@ -0,0 +1,105 @@
...
+
+	cfg := &tls.Config{}
+
+	if len(spec.Ciphers) > 0 {
+		ianaNames := crypto.OpenSSLToIANACipherSuites(spec.Ciphers)
+		var suites []uint16
+		for _, name := range ianaNames {
+			id, err := crypto.CipherSuite(name)
+			if err != nil {
+				continue
```
davidesalerno, line 46: I'll add a warning log here

cmd/dns-operator/main.go
```diff
@@ -49,18 +56,51 @@ func main() {
...
+	// Build TLS options for the operator metrics server from the
+	// cluster-wide TLS security profile when secure serving is enabled.
+	var metricsTLSOpts []func(*tls.Config)
+	if *metricsCertDir != "" {
+		cfgClient, err := configclient.NewForConfig(kubeConfig)
+		if err != nil {
+			logrus.Fatalf("failed to create config client: %v", err)
+		}
+		apiServer, err := cfgClient.ConfigV1().APIServers().Get(context.TODO(), "cluster", metav1.GetOptions{})
+		var tlsSecurityProfile *configv1.TLSSecurityProfile
```
davidesalerno, line 75: I think that we have to document the fact that we are reading the TLSSecurityProfile when the operator starts but we won't react to runtime TLS profile changes

davidesalerno (commented):
I shared some minor feedback and suggestion, let me know what you think.

pkg/operator/controller/tls.go
Thealisyed: Done!
Added the TODO referencing library-go#2347.

pkg/operator/controller/tls.go
Thealisyed: Good catch, it wasn't used anywhere

cmd/dns-operator/main.go
```diff
@@ -49,18 +56,53 @@ func main() {
...
 	}
 
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		logrus.Fatalf("failed to get kube config: %v", err)
+	}
+
+	// Build TLS options for the operator metrics server from the
+	// cluster-wide TLS security profile when secure serving is enabled.
+	// NOTE: the TLS profile is read once at startup; runtime changes to
```
bentito, line 66: While this static loading behavior is historically common in `controller-runtime` operators, it does introduce a subtle asymmetry in our design:
1. When the cluster TLS profile changes, the operator's reconciler dynamically updates the **CoreDNS DaemonSet** (operand), immediately rolling it out to match the new profile.
2. The **operator itself**, however, remains non-compliant on port `9393` until a manual restart or node reschedule occurs.

Across the OpenShift ecosystem, we have a standardized pattern to solve this exact dynamic update limitation. Instead of attempting to hot-reload TLS parameters (which can lead to inconsistent states on long-lived connections), operators use a **`SecurityProfileWatcher`** from the `openshift/controller-runtime-common`  library to trigger a graceful self-shutdown when the TLS profile changes. Because the operator's Deployment has `restartPolicy: Always`, Kubernetes instantly spawns a new pod, which re-reads the fresh TLS config on boot.

You can see an example of this in the [ptp-operator main.go](https://github.com/openshift/ptp-operator/blob/main/main.go).

This is definitely a separate, well-defined enhancement that doesn't need to block this PR. Let's merge this work as-is to unblock the `kube-rbac-proxy` removal, and we can track implementing the `SecurityProfileWatcher` in a follow-up Jira story. What do you think?

I think we should add a Jira to track this and do this as a follow-up PR.

bentito:
I had one comment (https://github.com/openshift/cluster-dns-operator/pull/482/changes#r3615162144) about TLS config changing that we can work on in a follow up story/PR. I think you've addressed all the comments and the PR looks solid!

/lgtm

melvinjoseph86:
checked with clusterbot.
```
➜  oc get clusterversion
NAME      VERSION                                               AVAILABLE   PROGRESSING   SINCE   STATUS
version   5.0.0-0-2026-07-21-074804-test-ci-ln-7wfc98t-latest   True        False         3h11m   Cluster version is 5.0.0-0-2026-07-21-074804-test-ci-ln-7wfc98t-latest

Verify kube-rbac-proxy sidecar is removed from operator pod,  show only "dns-operator", NOT "kube-rbac-proxy"
➜  oc get deployment dns-operator -n openshift-dns-operator -o json |  jq '.spec.template.spec.containers[].name'
"dns-operator"

Verify the operator is serving metrics directly with TLS:
➜  oc get deployment dns-operator -n openshift-dns-operator -o json | jq '.spec.template.spec.containers[] | select(.name=="dns-operator") | .args'
[
  "--metrics-bind-addr",
  ":9393",
  "--metrics-cert-dir",
  "/etc/tls/metrics"
]

➜   oc get deployment dns-operator -n openshift-dns-operator -o json |  jq '.spec.template.spec.containers[] | select(.name=="dns-operator") | .volumeMounts'
[
  {
    "mountPath": "/etc/tls/metrics",
    "name": "metrics-tls",
    "readOnly": true
  },
  {
    "mountPath": "/tmp",
    "name": "tmp-dir"
  }
]

Verify the new ClusterRole and ClusterRoleBinding exist and bind prometheus-k8s SA from openshift-monitoring
➜    oc get clusterrole openshift-dns-operator-metrics-reader -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
  creationTimestamp: "2026-07-21T08:11:47Z"
  name: openshift-dns-operator-metrics-reader
  ownerReferences:
  - apiVersion: config.openshift.io/v1
    controller: true
    kind: ClusterVersion
    name: version
    uid: 537e1482-321a-4d5a-bae9-84cf285b5221
  resourceVersion: "1823"
  uid: 9b350368-beba-42ac-b7b6-20b9bc443533
rules:
- nonResourceURLs:
  - /metrics
  verbs:
  - get

Verify CoreDNS DaemonSet Still Uses kube-rbac-proxy
➜  oc get ds dns-default -n openshift-dns -o json | \
    jq '.spec.template.spec.containers[].name'
"dns"
"kube-rbac-proxy"

 Verified metrics endpoint is reachable with auth and  Prometheus is still scraping
Also QE regression cases are passing
18 pass, 1 skip (45m18s)
```

Hence marking as verified 
/verified by @mjoseph

